PR #19, opened 2023-10-10
Datamodel Deployment

doctrino:
## Plan
- [x] Split the /data_model into /domain_model - incoming data and /solution_model - used by the inField.
- [x] Remove all read specific things form the data models, we should use the write version.
- [x] Change data model format from `.json` to `.yaml`.
- [x] Add `apm_model_version` to the `config.yaml`.
- [x] Add deployment script for data models.
  1.  Find all data models in the /domain_model and /solution_model directories. The type, data_model, view, or container, is expected (as it is today) in the file model.
  2. Check CDF if contianer, view, or data model exists.
  3. If exists, print diff.
  4. If changes, deploy
  5. If no changes, skip deployment.
  6. If no exists, deploy.

Output:
```
Found 3 containers in build\modules\cdf_apm_simple\domain_models.
Loaded from files: 
  container: 3
Found 1 spaces
Found 1 changed spaces.
  Would update 1 spaces.
Found 3 changed containers.
  Would update 3 containers.
Found 1 datamodels in build\modules\cdf_apm_simple\solution_models.
Found 3 views in build\modules\cdf_apm_simple\solution_models.
Loaded from files: 
  datamodel: 1
  view: 3
Found 1 spaces
Found 1 changed spaces.
  Would update 1 spaces.
Found 3 new views.
  Would create 3 views.
Found 1 changed datamodels.
  Would update 1 datamodels.
```

# Next PR
* Setup GitHub Actions


scripts/datamodel.py
gregertw, line 1: Fantastic!!

scripts/templates.py
gregertw, line 1: As we discussed, maybe we should split into a build_scripts and deploy_scripts dir so it will be easier to migrate logic into a CDF config service as we go?

scripts/templates.py
```diff
@@ -20,11 +22,12 @@ def read_yaml_files(yaml_dirs):
...
-                try:
-                    data.update(yaml.safe_load(f))
-                except:
-                    ...
+            try:
+                config_data = yaml.safe_load(yaml_file.read_text())
+            except yaml.YAMLError:
+                print(f"Error reading {yaml_file}")
+                continue
+            data.update(unpack_config(config_data))
```
gregertw, line 30: We should probably evaluate how we want to handle yaml going forward so we don't end up with several. load_yaml() loads a yaml file using @dataclass to enforce schema on transformations yaml files. It makes it easier to do client-side validation, but going forward, the less logic we have client-side, the better (if we can get a config CDF service running).

scripts/utils.py
```diff
@@ -226,7 +227,7 @@ def example(self, value: str):
         # Since we now have a new configuration, check the dataset and set the id
         self._data_set_id = self.verify_dataset()
 
-    def verify_client(self, capabilities: dict[list] = {}) -> CogniteClient:
+    def verify_client(self, capabilities: dict[str, list[str]] | None = None) -> CogniteClient:
```
gregertw, line 230: As discussed here, would be fantastic to get a generic verification into the SDK, but the current one here relying on the example config and preset data_set_id just makes problems for us. I'll leave it to your judgment on how much to do here vs work on the SDK side.

gregertw (approved):
I have made a few suggestions for changes. Please evaluate yourself and conclude what you want to do. Feel free to ping me for another review, but I approve, so you can make the call.

config.yaml
```diff
@@ -1,7 +1,11 @@
-
-apm_simple_raw_db: tutorial_apm
-apm_simple_datamodel: tutorial_apm_simple
-apm_simple_space: tutorial_apm_simple
-apm_simple_data_set: Valhall_System_23
-apm_simple_data_set_desc: Valhall_System_23
-cdf_cluster: ${CDF_CLUSTER}
\ No newline at end of file
+apm_simple_:
```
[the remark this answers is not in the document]
doctrino: Agree, will revert this change, and save the moving for a later PR

deploy.py
```diff
@@ -53,9 +52,13 @@ def run(build_dir: str) -> None:
         load_transformations_dump(
             ToolGlobals, file=None, drop=True, directory=f"{build_dir}/transformations"
         )
-    if Path(f"{build_dir}/data_models").is_dir():
+    if (models_dir := Path(f"{build_dir}/domain_models")).is_dir():
         load_datamodel_dump(
-            ToolGlobals, drop=True, directory=f"{build_dir}/data_models"
+            ToolGlobals, drop=True, directory=models_dir, dry_run=True
+        )
+    if (models_dir := Path(f"{build_dir}/solution_models")).is_dir():
```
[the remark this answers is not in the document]
doctrino: Agreed reversed.

modules/cdf_apm_simple/solution_models/WorkItem.view.yaml
```diff
@@ -0,0 +1,79 @@
...
+    containerPropertyIdentifier: itemName
+    name: itemName
+  linkedAssets:
+    direction: outwards
+    name: linkedAssets
+    source:
+      externalId: Asset
+      space: '{{apm_simple_space}}'
+      type: view
+      version: 8069f1498c7f9a
```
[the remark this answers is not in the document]
doctrino: That is a mistake, did not think of version in the one to many edges.

scripts/templates.py
doctrino: We need to consider to use `pydantic` for client side validation.

PR #22, opened 2023-10-13
Gtw/clean_up_tooling

gregertw:
* Factor away .example stuff from data-model-examples
* Clean up deploy.sh

### DIFF
--- a/deploy.py
+++ b/deploy.py
@@ -21,7 +21,7 @@ log = logging.getLogger(__name__)
 load_dotenv(".env")
 
 
-def run(build_dir: str) -> None:
+def run(build_dir: str, drop: bool = True, dry_run: bool = True) -> None:
     print(f"Deploying config files from {build_dir}...")
     # Configure a client and load credentials from environment
     build_path = Path(__file__).parent / build_dir
@@ -29,8 +29,6 @@ def run(build_dir: str) -> None:
         print(f"{build_dir} does not exists.")
         exit(1)
     ToolGlobals = CDFToolConfig(client_name="cdf-project-templates")
-    # TODO: #14 This is confusing heritage from data-model-examples. Refactor to use config.yaml and module structure.
-    ToolGlobals.example = "default"
     print("Using following configurations: ")
     print(ToolGlobals)
     # TODO: #6 This is a very limited support. Needs to be expanded to support configurable groups.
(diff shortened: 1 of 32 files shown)


deploy.py
```diff
@@ -43,18 +41,34 @@ def run(build_dir: str) -> None:
...
+    if (models_dir := Path(f"{build_dir}/domain_models")).is_dir():
+        load_datamodel_dump(
+            ToolGlobals, drop=drop, directory=models_dir, dry_run=dry_run
         )
-    if (models_dir := Path(f"{build_dir}/data_models")).is_dir():
+    if (models_dir := Path(f"{build_dir}/solution_models")).is_dir():
         load_datamodel_dump(
-            ToolGlobals, drop=True, directory=models_dir, dry_run=True
+            ToolGlobals, drop=drop, directory=models_dir, dry_run=dry_run
         )
```
doctrino, line 72: Reintroducing source, domain, and solution models at the top level of a build package?

modules/cdf_apm_simple/config.yaml
```diff
@@ -1,4 +1,13 @@
 # Only valid for this module, loads template variables from environment
+raw_db: tutorial_apm
+datamodel: tutorial_apm_simple
+space: tutorial_apm_simple
+data_set: Valhall_System_23
+data_set_desc: Valhall_System_23
+datamodel_version: "1"
+view_Asset_version: "1"
+view_WorkOrder_version: "1"
+view_WorkItem_version: "1"
```
doctrino, line 10: Great

scripts/datamodel.py
```diff
@@ -237,10 +253,18 @@ def load_datamodel_dump(
     for type_, resources in cognite_resources_by_type.items():
         print(f"  {type_}: {len(resources)}")
 
-    space_list = list({r.space for _, resources in cognite_resources_by_type.items() for r in resources})
+    space_list = list(
+        {
+            r.space
+            for _, resources in cognite_resources_by_type.items()
+            for r in resources
+        }
```
doctrino, line 261:
```suggestion
        {
            r.space
            for resources in cognite_resources_by_type.values()
            for r in resources
        }
```

doctrino (approved):
Seems like we have a fighting linter, as there are a few style changes. Will add some pre-commit checks when I get back to avoid this.

deploy.py
gregertw: You hadn't merged them into top level, so I just temporary fixed the loading.

gregertw: That wasn't very clearly explained: You HAD merged into top-level, but when doing a build, each of the sub-folders with data models is merged into one for each data model type. We need to implement support for ordering import of modules, but temporary, I just implemented step-wise loading of modules in the deploy.sh script.

scripts/datamodel.py
gregertw: Sorry, auto-merge was on, I will turn that off, it's not particularly good practice unless the change is tiny.

PR #23, opened 2023-10-13
Introducing self-contained and testable "workers" for each module resource

ronpal:
The idea is to have self-contained, testable logic for each of the resources within a module. Hence we can have 

```
<module>
   raw
      content
      worker.py
   time_series
      content
      worker.py
      ...
```

and

```
tests
    <module>
      raw
         test_worker.py
```

I've introduced pydantic for validation.

We will also need an orchestrator that ensures that the module workers are run in correct sequence according to config files in each folder.


gregertw:
Ref our discussion this morning: we need to introduce unit testing of the tooling at the root level to ensure that we don't introduce errors in e.g. yaml parsing etc. We can probably use the apm_simple module (or maybe an even smaller version?) as a unittest module. Then, we can look at introducing module level validation at a later point when we know more specifically what kind of issues we will meet.

PR #27, opened 2023-10-16
Gtw/support_global_module_config

gregertw:
* Add support for packages in global.yaml
* Add support for local.yaml to configure module loading and order
* Add support for integer (1. 2. ...) prefixing of config files in each module to support ordered loading
* Preserve order when building configuration files


ronpal (approved):
Looks good!

PR #28, opened 2023-10-16
Gtw/further_cleanups

gregertw:
* Update packages and pytjon SDK
* Support --dry-run
* Minor fixes to loading

### DIFF
--- a/scripts/datamodel.py
+++ b/scripts/datamodel.py
@@ -12,7 +12,6 @@ from dataclasses import dataclass
 import yaml
 from cognite.client.data_classes._base import CogniteResource
 from cognite.client.data_classes.data_modeling import (
-    View,
     ViewApply,
     SpaceApply,
     ContainerApply,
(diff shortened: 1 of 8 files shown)


scripts/datamodel.py
```diff
@@ -228,7 +227,7 @@ def load_datamodel_dump(
     if directory is None:
         raise ValueError("directory must be supplied.")
     model_files_by_type: dict[str, list[Path]] = defaultdict(list)
-    models_pattern = re.compile(r"^(\w+\.)?(container|view|datamodel)\.yaml$")
+    models_pattern = re.compile(r"^(.*\.)?(container|view|datamodel)\.yaml$")
```
ronpal, line 230: Personal preference: Not to big fan of regex, as it is hard to read and comprehend. Can this be written as plain Python?

ronpal (commented):
Added comment about regex

PR #29, opened 2023-10-16
Add support for dry-run and raw loading with db name

gregertw:
* --dry-run and --drop now supported as expected
* load RAW tables into raw_db config.yaml property configured per module

### DIFF
--- a/scripts/load.py
+++ b/scripts/load.py
@@ -64,8 +64,9 @@ def load_raw(
         f"Uploading {len(files)} .csv files to RAW database using {raw_db} if not set in filename..."
     )
     for f in files:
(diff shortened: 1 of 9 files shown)


scripts/load.py
```diff
@@ -94,30 +60,35 @@ def load_raw(
                     files.append(f)
     if len(files) == 0:
         return
-    print(f"Uploading {len(files)} .csv files to {raw_db} RAW database...")
+    print(
+        f"Uploading {len(files)} .csv files to RAW database using {raw_db} if not set in filename..."
+    )
     for f in files:
+        (_, db, _) = re.match(r"(\d+)\.(\w+)\.(\w+)\.csv", f).groups()
```
ronpal, line 67: What happens if the pattern doesn't match? Exception?

PR #30, opened 2023-10-17
GitHub action for deployments

ronpal:


scripts/templates.py
```diff
@@ -86,7 +87,11 @@ def read_yaml_files(yaml_dirs, name: str = "config.yaml"):
 
 
 def process_config_files(dirs, yaml_data, build_dir="./build"):
-    Path(build_dir).mkdir(exist_ok=True)
+    
+    path = Path(build_dir)
+    if path.exists():
+        shutil.rmtree(path)
```
DD5HT, line 93: if someone misconfigures their build dir, you nuke it? :D

scripts/worker.py
gregertw, line 1: I'm not sure I understand what worker.py is for? We shouldn't duplicate loading functionality. load_raw() in load.py should be used. Also, for any operations against the CDF project, please use the pattern already established by instantiating an CDFToolConfig (see utils.py) and then pass the object into the function, which will then have environ variables, and pre-instantiated .client available with all the correct settings.

tests/modules/cdf_apm_simple/raw/test_worker.py
gregertw, line 1: So, here we should run the tests of the building and deploying functions from load.py etc. The mock client can be injected into CDFToolGlobals (believe that needs to be implemented).

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,36 @@
...
+          python-version: 3.9
+      - uses: snok/install-poetry@v1
+        with:
+          version: 1.6.1
+          virtualenvs-create: false
+      - name: Install dependencies
+        run: poetry install
+      - name: "Build the package"
+        run: |
+          poetry run python build.py --package
```
gregertw: What is the --package meant to be? It is not implemented from what I can see?
Could we do --run-tests here, so that the gh-action runs through build and deploy stages using a mock CDF client and thus fails quickly? If build.sh and deploy.sh then just uses pytest to execute their respective tests, it will be easy to verify locally that all tests pass (and by using pytest, we can run individual tests from VSC).

scripts/templates.py
ronpal: The idea was to ensure deterministic behaviour. If we don't clean the directory, you may have unwanted leftover artifacts from previous build. I partly agree that a `--force` or `--clean` flag is slightly better.

scripts/worker.py
ronpal: It is just "spare parts" for now, will delete it.

.github/workflows/deploy.yml
ronpal: the --package was a dead end, will be removed

ronpal: Regarding testing: will pick this up in a separate PR #32

.github/workflows/trial.yml
gregertw, line 1: Are you going to merge this workflow to main?

build.py
```diff
@@ -12,11 +12,12 @@
 load_dotenv(".env")
 
 
-def run(build_dir: str) -> None:
+def run(build_dir: str, clean: bool = False) -> None:
     print(
-        f"Building config files from templates in ./modules and ./common into {build_dir}..."
+        f"Building config files from templates in into {build_dir}..."
```
gregertw, line 17:
```suggestion
        f"Building config files from templates into {build_dir}..."
```

gregertw (approved):
Looks great! Added a suggestion and a comment. Please merge at will.

PR #33, opened 2023-10-18
Add application data models for APM

gulkaktus:


gregertw (changes requested):
@gulkaktus this looks great! Yes, I suggest we keep this in sync with what is expected to work in production, so please add those missing as system models (and then we do an update when you roll out the system models). The testing of these templates will happen in greenfield, bluefield, and orangefield. 
I have been doing some refactoring of the tooling in parallel, and just finished a PR yesterday (https://github.com/cognitedata/cdf-project-templates/pull/31) where I have flattened out the split between types of models (to support proper ordering of how they are applied). Also, we have decided to keep cdf_apm_simple as a test module and rather build out the production templates in separate modules.
To make your PR compatible with these changes, please do the following:

1. Create a new module: modules/cdf_apm_base and a sub-dir data_models where you put ALL yaml files
2. Use the number prefixes to order how you want them loaded (but do all the containers first, then views)

gulkaktus:
@gregertw Added source data models, updated ordering and moved everything into a new module

gregertw:
Cleaned up and verified that everything is loading. Reviewed with Erik, who pointed out some errors. He will have a look Monday.

gregertw (approved):

gregertw:
Verified that the models load correctly, we should be good to go to merge this PR.

PR #34, opened 2023-10-19
Gtw/add_pre_commit

gregertw:
* Add pre-commits from python SDK
* Leave mypy temporary disabled
* Fix build.py --clean


ronpal (approved):
LGTM

PR #49, opened 2023-10-23
Support environments

gregertw:
build.py and deploy.py now support --env=<local|demo|dev|staging|prod>
Default is dev.

local.yaml supports configuration of environments, so you can deploy different modules dependent on environment.

config.yaml files now support prefixed variables with environment: <env>.my_var allowing you to set vars based on environment

CDF_ENVIRON and CDF_BUILD_TYPE are set as environment variables based on the settings in local.yaml.


ronpal (approved):
Looks good

PR #51, opened 2023-10-23
InField base groups

erikangelsen-cognite:
InField has 4 roles, these roles must have matching groups.
This PR defines the base roles with the correct cabapilities and scopes so that the groups can be created per configured location.

Only two groups hold different capabilities and scopes (viewer / normal), the additional checklist admin and tempalte admin has the same capabilities and the normal role, but is used to group users that has additional feature capabilities in InField.

- [ ] Add base groups for the 4 roles we have in InField
- [ ] Make an option to change the group / role names
- [ ] Link groups to locations, as each location must generate its 4 groups


modules/default_location/auth/infield_checklist_admin_role.group.yaml
```diff
@@ -0,0 +1,89 @@
...
+          'cdf_infield',
+          {{apm_datamodel_space}} # from modules/config.yaml
+        ]
+  - dataModelInstancesAcl:
+      actions:
+        - READ
+      scope:
+        spaceScope: [
+          'APM_Config',
+          'cdf_apm',
```
gulkaktus, line 74: Side note: Discussed with FDM team, we won't need `instance read` on the system data models if we remove the `globalNode`s (referenceTemplateItems etc.), which are optional.

modules/cdf_infield_location/config.yaml
```diff
@@ -0,0 +1,20 @@
+---
+location_name: default
+template_version: '1'
```
gregertw, line 3: module_version

modules/cdf_infield_location/config.yaml
```diff
@@ -0,0 +1,20 @@
...
+apm_datamodel_space: default_apm_datamodel_space
+location_app_data_dataset: default_location_app_data_dataset
+location_source_data_dataset: default_location_source_data_dataset
+location_source_data_space: default_location_source_data_space
+location_app_data_space: default_location_app_data_space
```
gregertw, line 8: Get rid of these ref discusison

scripts/templates.py
```diff
@@ -174,6 +175,9 @@ def process_config_files(
                     # assuming template variables are in the format {{key}}
                     content = content.replace(f"{{{{{k}}}}}", str(v))
 
+                for unmatched in re.findall(pattern=r"\{\{.*?\}\}", string=content):
+                    print(f"WARNING: Unmatched template variable {unmatched} in {dirpath}/{file}")
```
gregertw, line 179: Nice!!

gregertw (approved):
This is really great! Getting the scoping to data sets per location documented (and deployed controlled) is going to save so much time!

PR #56, opened 2023-10-24
Infield app space

ronpal:


gregertw (approved):
Ref discussion: we use a common module to prepare also for possible future common needs for infield. cc @erikangelsen-cognite

PR #58, opened 2023-10-26
Gtw/fix_raw_transformations

gregertw:
* Fix up loading of raw to not create tables with numbered prefix and handle deleted/deletion of database correctly
* Fix transformations to handle authentication and schedule
* Expose oauth_credentials from CDFToolConfig for transformations loading to use


ronpal (approved):
Looks good!

PR #59, opened 2023-10-26
Rename to apm_simple and fix up to use new auth and schedule functionality

gregertw:


modules/cdf_apm_simple/transformations/apm_simple-load-asset2children.yaml
```diff
@@ -1,5 +1,5 @@
-externalId: tutorial-load-asset2children
-name: tutorial-load-Asset2children
+externalId: apm_simple-load-asset2children
+name: apm_simple-load-Asset2children
```
ronpal, line 2: Pet peeve: mixing `underscore` and `dash` in a string often leads to confusion or mistakes. Can it be dashes all the way?

ronpal (approved):
Looks good

modules/cdf_apm_simple/transformations/apm_simple-load-asset2children.yaml
gregertw: Agreed. The challenge is that here we have two different "cultures" clashing. "apm_simple" is a prefix used in many different contexts where a dash will not work that well. The rest is "inherited" from the pattern used for transformations. The bigger question here is probably standardisation on externalId. We are now all over the place. "Asset", "pi:12334", "snake_case", and "this-as-a-transformation"....
I think we should probably do a single PR standardising based what we choose.

PR #66, opened 2023-10-27
Gtw/various_fixes

gregertw:
* add max_workers before applying data models
* text fixes
* Fixup yaml


global.yaml
ronpal, line 1: 🙌

ronpal (approved):
Nice!

PR #80, opened 2023-10-30
Pump Example DataSet

doctrino:
Creates an example module with a pump datasets. This is needed in the step to build a demo for migration.

A few things:
* I hide the `local.yaml` and instead made an `example.local.yaml` as I assume the `local.yaml` will be different for each user.
* Made a few helping lines where I ran into trouble in the deployment steps.
* Some fixes to make the `load.py` compatible with 3.9 as that is what the `pyproject.toml` is saying.


.gitignore
ronpal, line 1: local.yaml is supposed to be checked in when a template is used, but for the template repo we should have it as an example file (like we do with example `.env`s)

.gitignore
```diff
@@ -272,4 +272,7 @@ $RECYCLE.BIN/
 
 # Custom rules (everything added below won't be overriden by 'Generate .gitignore File' if you use 'Update' option)
 
-tmp/
\ No newline at end of file
+tmp/
+
+# Local configuration file (uncomment if needed)
+local.yaml
```
ronpal, line 278: local.yaml is supposed to be checked in when a template is used, but for the template repo we should have it as an example file (like we do with example .envs)

ronpal (approved):
See comment about local.yaml, otherwise: 👍

.gitignore
doctrino: Will remove this than and merge.

scripts/utils.py
```diff
@@ -111,8 +111,11 @@ def __init__(
             )
 
     def __str__(self):
+        environment = self._environ.copy()
+        if "IDP_CLIENT_SECRET" in environment:
+            environment["IDP_CLIENT_SECRET"] = "***"
```
gregertw, line 116: Great!
We should probably add COGNITE_CLIENT_SECRET and TRANSFORMATION_CLIENT_SECRET here as well-

modules/example_pump_asset_hierarchy/transformations/pump_asset_hierarchy-load-collections_pump.yaml
```diff
@@ -0,0 +1,21 @@
...
+  # Optional: If idP requires providing the scopes
+  cdfProjectName: {{cdfProjectName}}
+  scopes:
+    - {{scopes}}
+  # Optional: If idP requires providing the audience
+  audience: {{audience}}
+schedule:
+  # every hour
+  interval: '0 * * * *'
+  is_paused: true
```
gregertw, line 21: Just realised I have mixed camelCase and snake_case in the apm_simple transformations as well. How is this handled on the SDK side? Should we allow anything?

scripts/utils.py
doctrino: Yes definitely, but think the whole authentication needs a refactoring. Should not be necessary to setup URLs.

modules/example_pump_asset_hierarchy/raw/1.pump_assets.collections_pump.csv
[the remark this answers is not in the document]
doctrino: That is my mistake I should have checked more carefully before getting started. Advantage of this is that comes from the free which means it has to demonstrate how to fit an external source into CDF when there is no parent external id. In addition, it has much more metadata which will demonstrate the advantage of CDF.

modules/example_pump_asset_hierarchy/config.yaml
```diff
@@ -0,0 +1,13 @@
+# Only valid for this module, loads template variables from environment
+raw_db: pump_assets
+data_set: src:lift_pump_stations
+data_set_desc: This dataset <a href="https://data.bendoregon.gov/maps/collections-pump">lift_pump_stations</a> contains the lift pump stations.
+clientId: ${IDP_CLIENT_ID}
```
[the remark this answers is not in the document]
gregertw: Yeah, so this probably holds for clientId ++

modules/example_pump_asset_hierarchy/transformations/pump_asset_hierarchy-load-collections_pump.yaml
gregertw: I think maybe non-matching attributes are just silently ignored. If we use isPaused here, I guess it will work...

modules/example_pump_asset_hierarchy/raw/1.pump_assets.collections_pump.csv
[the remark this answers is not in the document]
gregertw: Yeah, I'm not opposed to another data set. We just need to ensure that the end result in FDM is compatible.

modules/example_pump_asset_hierarchy/transformations/pump_asset_hierarchy-load-collections_pump.yaml
doctrino: Yes, non-matching attributes are silently ignored (to avoid breaking the SDK when new ones are added to the API). 

`is_paused` is actually missing from the SDK, so it will not be set.

PR #81, opened 2023-10-30
Recursive delete support

ronpal:


clean.py
```diff
@@ -38,7 +56,24 @@ def run(build_dir: str) -> None:
     parser = argparse.ArgumentParser(epilog="Further functionality to be added")
     parser.add_argument(
         "build_dir",
+        default="./build",
+        nargs="?",
         help="Where to pick up the config files to deploy",
     )
+    parser.add_argument(
```
gregertw, line 63: You may want to add a default here. The current default is implicitly defined.

modules/cdf_infield_location/auth/infield_normal_role.group.yaml
gregertw, line 1: You should probably split out the cdf_infield_location changes into a separate PR. The changes are unrelated.

scripts/delete.py
```diff
@@ -19,7 +19,7 @@
 from cognite.client.data_classes.data_modeling import ViewId
 from cognite.client.data_classes.time_series import TimeSeries
 
-from .transformations_config import parse_transformation_configs
+# from scripts.transformations.transformations_config import parse_transformation_configs
```
gregertw, line 22: Just delete this. It's a leftover from the code that used the transformations-cli code (and delete.py was not refactored)

scripts/delete.py
```diff
@@ -109,19 +109,19 @@ def delete_timeseries(ToolGlobals: CDFToolConfig, dry_run=False, directory=None)
...
-    configs = parse_transformation_configs(directory)
-    transformations_ext_ids = [t.external_id for t in configs.values()]
-    try:
-        if not dry_run:
-            client.transformations.delete(external_id=transformations_ext_ids)
-        print(f"Deleted {len(transformations_ext_ids)} transformations.")
-    except Exception:
-        print("Failed to delete transformations. They may not exist.")
-        return
+# def delete_transformations(ToolGlobals: CDFToolConfig, dry_run=False, directory=None) -> None:
```
gregertw, line 112: Just delete

scripts/run.py
```diff
@@ -14,24 +14,23 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Optional
 
-from .transformations_config import parse_transformation_configs
-from .utils import CDFToolConfig
+# from .transformations_config import parse_transformation_configs
```
gregertw, line 18: And delete here. Do we want to run transformations at all? We should probably rely on schedules found in the config. Maybe delete the entire file?

gregertw (commented):
See suggestions.

gregertw (approved):
Great!

PR #85, opened 2023-10-30
Pump Example

doctrino:
Ready to review. 

The guide written in the official docs following this example https://pr-1865.docs.preview.cogniteapp.com/cdf/dm/dm_guides/dm_create_asset_hierarchy


modules/cdf_asset_hierarchy_migration/data_models/1.Asset.container.yaml
```diff
@@ -0,0 +1,77 @@
...
+    autoIncrement: false
+    defaultValue: null
+    description: The labels of the asset.
+    name: labels
+    nullable: true
+    type:
+      collation: ucs_basic
+      list: true
+      type: text
+# Include Labels and GeoLocation?
```
doctrino, line 64:
```suggestion
# Include GeoLocation?
```

modules/cdf_asset_hierarchy_migration/data_models/1.Asset.container.yaml
uzairwali, line 1: root is another property on Asset types. Direct relation to another asset node.

examples/example_pump_data_model/data_models/1.Pump.container.yaml
```diff
@@ -0,0 +1,48 @@
+externalId: Pump
+name: Pump
+space: '{{space}}'
```
gregertw, line 3: We should already now follow the proper pattern for handling spaces:
1. Global spaces are used as core containers and views to be extended (e.g. cdf_core)
2. Customer extended governed/global containers, views, and data models should live in a read-only "custom_data_model" space with read-only
3. Data instances should be ingested into a separate data space referring to the customer extended views

Even more importantly from a migration point of view, we need an 1.5 where Asset and related modeling entities should live in a non-system space (as they are not sanctioned as Cognite system entities), but where we don't want variance across customers. I.e. we should establish a "global_templates" space that only contains template-models that are managed by Cognite on top of the global, system models, but separate from the customer's own data models.

experimental/cdf_asset_hierarchy_replication/transformations/asset_hierarchy_replication-load-data_model.sql
```diff
@@ -0,0 +1,13 @@
+select
+  cast(`externalId` as STRING) as externalId,
+  node_reference('{{space}}', `parentExternalId`) as parent,
+  node_reference('{{space}}', '{{root_asset_external_id}}') as root,
+  cast(`name` as STRING) as name,
+  cast(`source` as STRING) as source,
+  cast(`description` as STRING) as description,
+  cast(`labels` as ARRAY < STRING >) as labels,
+  to_json(`metadata`) as metadata
```
gregertw, line 9: I think it would be wise to already now show the pattern for how to extract specific keys from the metadata and transform into properties in the view.

gregertw (commented):
I would suggest that you create a separate PR for the changes to the scripts and elevate the clientId, clientSecret ++ up to default.config.yaml. It will simplify this PR and align with the current status of the tooling.
I have also commented on some specifics wrt the data model entities. My gut feeling is that we should aim at moving cdf_asset_hierarchy_replication into a modules/cdf_ module and install it default as quickly as possible. And then we should in the example show how to extend with a customer-specific pump++.

common/cdf_auth_readwrite_all/config.yaml
gregertw, line 1: We should not commit config.yaml files in the modules, only default.config.yaml. The challenge is that if we include them in .gitignore, template copies require removing exclusions before being able to commit config.yaml files in customer repos.

examples/example_pump_data_model/README.md
gregertw, line 1: Should you reference the docs here? You already know the URL, right?

examples/example_pump_data_model/default.config.yaml
```diff
@@ -0,0 +1,18 @@
+# Only valid for this module, loads template variables from environment
+model_space: cdfTemplate
```
gregertw, line 2: This is a bit generic name? It would be good to have it more specific and maybe indicate with a comment how it should be overridden in config.yaml?

examples/example_pump_data_model/transformations/pump_model-populate-lift_station_pumps_edges.sql
```diff
@@ -0,0 +1,8 @@
+select
+  concat(cast(`parent`.externalId as STRING), ':', cast(`externalId` as STRING)) as externalId,
+  `parent` as startNode,
+  node_reference('{{instance_space}}', cast(`externalId` as STRING)) as endNode
+from
+  cdf_data_models("{{source_model_space}}", "{{source_model}}", "1", "Asset")
+where
+  startswith(title, 'Pump')
```
gregertw, line 8: Should this be a variable? Either Pump or maybe even the full where statement for how to pick out pumps?

examples/example_pump_data_model/default.config.yaml
```diff
@@ -0,0 +1,18 @@
...
+instance_space: cdfTemplateInstances
+source_model_space: ExtendedSourceDataModels
+source_model: ExtendedSourceData
+view_Pump_version: "1"
+view_LiftStation_version: "1"
+data_model_version: "1"
+data_model: "PumpLiftStations"
+clientId: ${IDP_CLIENT_ID}
```
gregertw, line 10: I have suggested this before and would like to do it again to avoid lots of duplication: let's move the CI/CD service principal credentials to the global default.config.yaml and make it explicit. E.g. cicd_client_id or similar.

examples/example_pump_data_model/transformations/pump_model-populate-lift_station_pumps_edges.yaml
```diff
@@ -0,0 +1,27 @@
+externalId: pump_model-populate-lift_station_pumps_edges
+name: pump_model-populate-lift_station_pumps_edges
+destination:
+  dataModel:
+    space: {{model_space}}
+    externalId: {{data_model}}
+    version: '{{view_LiftStation_version}}'
```
gregertw, line 7: We are mixing snake_case and CamelCase a bit too much. Should we standardise? I don't have strong opinions beyond the mixing of the two as it becomes superhard to remember. One possibility would be to standardise on CamelCase (i.e. lower first): viewLiftStationVersion alternatively PascalCase, ViewLiftStationVersion.

gregertw: Or just make it simple and do snake_case all over...

examples/example_pump_data_model/transformations/pump_model-populate-pump_container.yaml
```diff
@@ -0,0 +1,26 @@
...
+  clientSecret: {{clientSecret}}
+  tokenUri: {{tokenUri}}
+  # Optional: If idP requires providing the scopes
+  cdfProjectName: {{cdfProjectName}}
+  scopes: {{scopes}}
+  # Optional: If idP requires providing the audience
+  audience: {{audience}}
+schedule:
+  # every hour
+  interval: '0 * * * *'
```
gregertw, line 25: We are starting to get quite a few places with hardcoded schedules now. Maybe we should add schedules as defaul.config.yaml variables:
`scheduleHourly: '0 * * * *'
scheduleQuarterly: ...`

experimental/cdf_asset_source_model/default.config.yaml
```diff
@@ -0,0 +1,15 @@
+# Only valid for this module, loads template variables from environment
+model_space: ExtendedSourceDataModels
+instance_space: cdfTemplateInstances
+view_asset_version: '1'
+data_model_version: '1'
+root_asset_external_id: 'lift_pump_stations:root'
+clientId: ${IDP_CLIENT_ID}
```
gregertw, line 7: As already noted: Should these be moved to global default.config.yaml

experimental/cdf_asset_source_model/transformations/sync-asset_hierarchy_cdf_asset_source_model.sql
gregertw, line 1: This is great! I would assume we should apply the same approach to the infield transformation @ronpal

experimental/cdf_asset_source_model/transformations/sync-asset_hierarchy_cdf_asset_source_model.sql
```diff
@@ -0,0 +1,48 @@
+-- Root Asset
+-- The asset must be set up in hierarchical order as the container for the parent asset requires the
+-- parent asset to be created first.
+select
```
gregertw, line 4: Would be great with a simple description of what the logic behind the select is.

experimental/cdf_asset_source_model/transformations/sync-asset_hierarchy_cdf_asset_source_model.yaml
```diff
@@ -0,0 +1,26 @@
...
+    externalId: ExtendedSourceData
+    version: '{{data_model_version}}'
+    destinationType: Asset
+  instanceSpace: {{instance_space}}
+  type: instances
+
+# Specify credentials separately like this:
+# You can also use different credentials for the running transformations than the ones you use to deploy
+authentication:
+  clientId: {{clientId}}
```
gregertw, line 15: And here

local.yaml
gregertw, line 1: Don't commit this file.

tests/test_approval_modules_snapshots/cdf_auth_readwrite_all.yaml
gregertw, line 1: Should these changes be committed or is this by accident?

gregertw (commented):
I have added some comments to standardisation that I think we need to make a decision on now. If you conclude and make those changes, I would like to do a test on the modules before we merge.

cdf.py
gregertw, line 1: Should we rename cdf.py to cdf-tk.py and then install cdf-tk as a script in pyproject.toml?

common/cdf_auth_readwrite_all/config.yaml
doctrino: That was an accident, as it is not ignored is tricky.

examples/example_pump_data_model/transformations/pump_model-populate-lift_station_pumps_edges.sql
doctrino: this is a bit particular part of the data set. I would ideally use the external id, but there was a bug in transformations preventing me from doing that (reported to them)

examples/example_pump_data_model/default.config.yaml
doctrino: Let me do a specific config PR.

cognite_toolkit/cdf_tk/load.py
```diff
@@ -521,6 +521,10 @@ def load_datamodel(
             cognite_resources_by_type[type_].append(
                 resource_cls.load(load_yaml_inject_variables(file, ToolGlobals.environment_variables()))
             )
+    # Remove duplicates
```
gregertw, line 524: Should this result in a warning? How come we end up with duplicates? The same module mentioned twice in local.yaml?

cognite_toolkit/examples/example_pump_data_model/README.md
gregertw, line 1: It might be an idea to put both modules in experimental for now if we want to merge to make it simple to apply to a test project before Wednesday next week?

cognite_toolkit/examples/example_pump_data_model/data_models/2.Pump.container.yaml
```diff
@@ -0,0 +1,48 @@
...
+    type:
+      list: false
+      type: float64
+constraints:
+  requiredAsset:
+    constraintType: requires
+    require:
+      type: container
+      space: cdf_core
+      externalId: Asset
```
gregertw, line 48: I think this is potentially very confusing to people: requiring Asset container without implement Asset view. If you understand the details of how this work, it makes of course sense, but I think this is a thing that be very difficult to get your head around.

cognite_toolkit/examples/example_pump_asset_hierarchy/README.md
gregertw, line 1: This module is now essentially empty? Is this a leftover from a renaming? Ah, no. Just realised that this module was already merged in another PR, so these are just changes.

cognite_toolkit/default.packages.yaml
```diff
@@ -18,3 +18,8 @@ packages:
     - cdf_apm_base
     - cdf_infield_common
     - cdf_infield_location
+  example_pump:
+    - example_pump_asset_hierarchy
+    - cdf_asset_source_model
+    - example_pump_data_model
```
gregertw, line 24: There's something that is not logical about the naming of the modules. You would think that there is a dependency on the data model from the example_pump_asset_hierarchy. I think we need more info in the READMEs to also describe the relationships between the models. Maybe we should introduce a naming convention for modules with data? like example_pump_asset_hierarchy_data

PR #102, opened 2023-11-03
Ingesting app config

ronpal:


gregertw (approved):
Ref discussion/note for future: we want to turn this into a generic load instance loader, but pushing this to unblock infield.

PR #103, opened 2023-11-03
Interactive deploy

doctrino:
This is useful when developing locally and need to switch between which modules you deploy.


gregertw (approved):
Great addition!

PR #106, opened 2023-11-03
SDK v7

doctrino:


examples/cdf_apm_simple/default.config.yaml
```diff
@@ -10,10 +10,9 @@ view_WorkOrder_version: '1'
 view_WorkItem_version: '1'
 clientId: ${IDP_CLIENT_ID}
 clientSecret: ${IDP_CLIENT_SECRET}
-tokenUrl: ${IDP_TOKEN_URL}
+tokenUri: ${IDP_TOKEN_URL}
 # Optional: If idP requires providing the scopes
 cdfProjectName: ${CDF_PROJECT}
-scopes:
-  - ${IDP_SCOPES}
+scopes: ${IDP_SCOPES}
```
doctrino, line 16: Scopes already comes as a list from `IDP_SCOPES` which caused this to become a list of list.

examples/cdf_apm_simple/transformations/apm_simple-load-asset2children.yaml
```diff
@@ -27,4 +27,4 @@ authentication:
 schedule:
   # every hour
   interval: '0 * * * *'
-  is_paused: true
\ No newline at end of file
+  isPaused: true
```
doctrino, line 30: Eariler `isPaused` did not work when you deployed it, now it does.

scripts/load.py
```diff
@@ -300,60 +299,25 @@ def load_transformations(
...
-                        external_id=tmp.external_id,
-                        interval=config.get("schedule", {}).get("interval", ""),
-                    ),
-                    has_source_oidc_credentials=(len(source_oidc_credentials) > 0),
-                    has_destination_oidc_credentials=(len(destination_oidc_credentials) > 0),
-                    data_set_id=tmp.data_set_id,
-                )
-            )
+        raw = yaml.safe_load(f.read_text())
+        # The `authentication` key is custom for this template:
```
gregertw, line 310: Should we switch to something that is closer to the API here at the same time? The current yaml is sort of a mix of transformation manifest and API.

modules/cdf_infield_location/default.config.yaml
```diff
@@ -21,7 +21,7 @@ infield_location_viewer_users_source_id: 7bdcb20c-3e6a-400c-b9ef-cf835f64f05e
 # Transformation credentials
 clientId: ${IDP_CLIENT_ID}
 clientSecret: ${IDP_CLIENT_SECRET}
-tokenUrl: ${IDP_TOKEN_URL}
+tokenUri: ${IDP_TOKEN_URL}
```
gregertw, line 24: Why do you want to switch to tokenUri? It will always be a url.

gregertw (commented):
I'm not sure about the tokenUri switch and I sort of expected that the two load functions in utils.py now could be deprecated?

modules/cdf_infield_location/default.config.yaml
doctrino: It is actually not, transformations has `tokenUri`, see the current version of the SDK https://github.com/cognitedata/cognite-sdk-python/blob/master/cognite/client/data_classes/transformations/common.py#L300

As well as the spec:
image

scripts/load.py
doctrino: I think we should, my thinking here was just to get this PR in with as little changes as possible compared to the current flow.

modules/cdf_infield_location/default.config.yaml
gregertw: Hm, this is bad as we always refer to tokenUrl other places (which is more specific). Although URI is not wrong per se, it is imprecise as it always will be a url. Ok, I guess we should let the API spec trump consistency here?

gregertw (approved):

modules/cdf_infield_location/default.config.yaml
doctrino: I guess so, unfortunately

PR #123, opened 2023-11-06
QA InField Template

doctrino:


scripts/templates.py
```diff
@@ -209,7 +209,7 @@ def process_config_files(
                         file = file.split(".", 1)[1]
                     # If we are processing raw tables, we want to pick up the raw_db config.yaml
                     # variable to determine the database name.
-                    if dirpath.split("/")[-1] == "raw":
```
doctrino, line 212: This only works on Linux as Windows use `\\`.

PR #125, opened 2023-11-07
Gtw/consolidate_scripts

gregertw:
* Consolidate everything into the cdf.py command with support for cdf.py build|deploy|clean
* Clean up status texts to make it easier to follow what happens and leverage rich printing


pyproject.toml
```diff
@@ -15,6 +15,7 @@ pyyaml = "^6.0.1"
 dacite = "^1.8.1"
 regex = "^2023.6.3"
 chardet = "^5.1.0"
+typer = {version = "^0.9.0", extras = ["all"]}
```
doctrino, line 18: Worth to add an entry script so you can do `cdf [command]` instead of `python cdf.py [command]`?

gregertw: Not sure what an entry script is? I just do ./cdf.py (due to the #!/usr/bin/env python at line 1). I was comtemplating adding a poetry script entry (maybe that's what you mean?), but then you have to install first, right? I was thinking maybe you wanted to add that in a subsequent PR :)

doctrino (approved):
A no-brainer to me :)

PR #130, opened 2023-11-08
Gtw/add_auth_command

gregertw:
* Improve command docs and running without a command
* Add --cluster and --project as global options
* Add auth verify command


scripts/bootstrap.py
```diff
@@ -0,0 +1,355 @@
...
+            if type(a) is not type(cap):
+                continue
+            for action in cap.actions:
+                if action in a.actions:
+                    continue
+                all_ok = False
+                break
+            if all_ok:
+                break
+        if not all_ok:
```
haakonvt, line 266: Fyi, this will soon become very much simpler. This might also fail wrongly when more than one group has the same type of capability.

haakonvt: Actually, even with just one group you might have several different definitions of the same capability.

gregertw: Yes, the way this is built up is cumbersome... It's always the challenge of dataset and space scope vs all as well.

scripts/bootstrap.py
```diff
@@ -0,0 +1,355 @@
...
+    client_secret = ToolGlobals.environ(
+        "IDP_CLIENT_SECRET",
+        None,
+        fail=False,
+    )
+    print("[bold]Checking current service principal/application...[/]")
+    if cluster is None or len(cluster) == 0:
+        print("  [bold red]ERROR:[/]Environment variable CDF_CLUSTER must be set.")
+        ToolGlobals.failed = True
+        return
```
doctrino, line 66:
```suggestion
        exit(1)
```

scripts/bootstrap.py
```diff
@@ -0,0 +1,355 @@
...
+    if verbose:
+        print(f"  CDF_CLUSTER={cluster} is set correctly.")
+    if project is None or len(project) == 0:
+        print("  [bold red]ERROR:[/]Environment variable CDF_PROJECT must be set.")
+        ToolGlobals.failed = True
+        return
```
doctrino, line 72:
```suggestion
        exit(1)
```

scripts/bootstrap.py
```diff
@@ -0,0 +1,355 @@
...
+        print(f"  CDF_PROJECT={project} is set correctly.")
+    if token is not None and len(token) > 0:
+        print("  Env variable CDF_TOKEN is set, using it as Bearer token for authorization.")
+    else:
+        if verbose:
+            print("  CDF_TOKEN is not set, expecting IDP_CLIENT_ID and IDP_CLIENT_SECRET...")
+        if client_id is None or len(client_id) == 0:
+            print("  [bold red]ERROR:[/]Environment variable IDP_CLIENT_ID must be set.")
+            ToolGlobals.failed = True
+            return
```
doctrino, line 83:
```suggestion
            exit(1)
```

scripts/bootstrap.py
```diff
@@ -0,0 +1,355 @@
...
+        if client_secret is None or len(client_secret) == 0:
+            print("  [bold red]ERROR:[/]Environment variable IDP_CLIENT_SECRET must be set.")
+            ToolGlobals.failed = True
+            return
```
doctrino, line 87:
```suggestion
            exit(1)
```

scripts/bootstrap.py
```diff
@@ -0,0 +1,355 @@
...
+        # Using the token/inspect endpoint to check if the client has access to the project.
+        # The response also includes access rights, which can be used to check if the client has the
+        # correct access for what you want to do.
+        resp = ToolGlobals.client.iam.token.inspect()
+        if resp is None or len(resp.capabilities) == 0:
+            print(
+                "  [bold red]ERROR[/]: Valid authentication token, but it does not give any access rights. Check credentials (CDF_CLIENT_ID/CDF_CLIENT_SECRET or CDF_TOKEN)."
+            )
+            ToolGlobals.failed = True
+            return
```
doctrino, line 100:
```suggestion
            exit(1)
```

doctrino (approved):
Lots of return None, I guess should just be `exit(1)` but not a big issue. 

The only thing is that this does not seem to solve the issue of the sequence for execution, first give access, then create spaces and dataset ids, before creating access for those spaces and datasets.

scripts/bootstrap.py
gregertw: I use this pattern because the caller can then decide how to handle the failure situation. Maybe not in this case, but in other cases, the same error should be treated as exit(1) or fail silently. I'm thus a bit hesitant to terminate execution from a library function.

PR #142, opened 2023-11-13
Added required magic group for InField to infield common

tesande:


ronpal:
Correct me if I'm wrong, but this group enables app admin view in Fusion? The app config UI has been moved to within InField in 2.0, do you still see the need for the UI in Fusion? https://cognite-infield-staging.firebaseapp.com/application-configuration

ronpal (approved):

PR #143, opened 2023-11-13
Gtw/create_script_pkg

gregertw:
* Make cdf-tk an installable package
* Introduce cdf-tk init [--upgrade] as a way to create/upgrade a project


pyproject.toml
```diff
@@ -1,11 +1,12 @@
...
+name = "cognite-toolkit"
 version = "0.1.0"
-description = "Official Cognite Data Fusion project templates"
-authors = ["Greger Wedel <greger.wedel@cognite.com>"]
+description = "Official Cognite Data Fusion tool for project templates and configuration deployment"
+authors = ["Cognite AS <support@cognite.com>"]
 license = "Apache-2"
 readme = "README.md"
 packages = [{include = "cdf_tk", from = "."}]
+include = ["cdf.py", "docs/**/**","common/**/*", "modules/**/*", "examples/**/*", "local_modules/**/*", "default.*.yaml", "config.yaml", "local.yaml", "packages.yaml", "README.md"]
```
doctrino, line 9: Do we really include everything into the package? I think we need to put it all in one folder, as if we do not, it can easily get mixed up with other installations.

cdf.py
```diff
@@ -498,5 +502,114 @@ def auth_verify(
...
+    template_dir = resources.files("cdf_tk").parent
+    target_dir = Path.cwd() / f"{init_dir}"
+    if target_dir.exists():
+        if not upgrade:
+            print(f"Directory {target_dir} already exists.")
+            exit(1)
+        else:
+            print(f"[bold]Upgrading directory {target_dir}...[/b]")
+    if not dry_run and not upgrade:
+        os.mkdir(target_dir)
```
doctrino, line 563:
```suggestion
        target_dir.mkdir(exists_ok=True)
```

doctrino (commented):
Main feedback is that I think we should keep everything in one folder, such that when it is installed it will not easily mix with everything else that is installed in that environment.

doctrino (approved):

gregertw:
@cognitedata/opsup-team fyi, these changes are tooling only, so your review is not needed,

PR #155, opened 2023-11-16
Gtw/improve_gh_actions

gregertw:
Clean up GH actions and standardize
Introduce release branch and package push to pypi


.github/workflows/release.yaml
```diff
@@ -0,0 +1,91 @@
...
+          virtualenvs-create: false
+      - name: Install dependencies
+        run: poetry install
+      - name: Build package
+        run: poetry build
+      - name: Release to PyPI
+        env:
+          TWINE_USERNAME: __token__
+          TWINE_PASSWORD: ${{ secrets.PYPI_API_TOKEN }}
+        run: twine upload --repository testpypi --verbose dist/* || echo 'Version exists'
```
doctrino, line 91: Need twine as a dev dependency.

.github/workflows/build.yml
```diff
@@ -1,13 +1,10 @@
...
-
+name: Lint and test
 on:
   pull_request:
     branches: [main]
-
 env:
-  PYTHON_VERSION: '3.9'
-
+  PYTHON_VERSION: '3.11'
```
doctrino, line 7: In the `pyproject.toml` we have a lower bound for 3.9, we need to test for this as well then, so we do not introduce syntax which is not allowed in 3.9.

.github/workflows/demo.yml
```diff
@@ -1,8 +1,10 @@
-name: Deploy modules to dev
+name: Deploy modules to demo project
 on:
   push:
     branches:
       - main
+env:
+  PYTHON_VERSION: '3.11'
```
doctrino, line 7: Same argument here, if we allow `3.9` as the lower bound we should use it to ensure that we do not use Python syntax for 3.10 and above

.github/workflows/release.yaml
```diff
@@ -0,0 +1,91 @@
...
+      - name: Initialize project
+        run: cdf-tk init demo_project
+      - name: "Pre-processing for demo environment"
+        run: ./demo/preproc.sh
+      - name: "Build the templates"
+        run: cdf-tk build --build-dir=./build --env=demo ./demo_project
+  deploy:
+    runs-on: ubuntu-latest
+    environment: CD
+    needs: [lint, test, build]
```
doctrino, line 73: This has been a debate in the Python-SDK. Do we run everything of linting, test and build in main before release? It has already been run on the branch before merging. The problem in the SDK is that sometimes due to instability in the API (very often Transformations) the tests fail and the release does not do through. This leads to multiple versions never being released to pypi.

.github/workflows/release.yaml
gregertw: Right... :)

.github/workflows/build.yml
gregertw: Ok

.github/workflows/release.yaml
gregertw: Yes, see that point. But we are not doing anything live here. With a release branch, you can theoretically merge in from something else than main (unless we do extra enforcement), so this is an extra safeguard.

doctrino (approved):

PR #157, opened 2023-11-17
Generalized Load

doctrino:
This is a pure refactoring, meaning there is no change to functionality (with the exception of allowing `.parquet` format for dataframes)

Basic Idea: Have a base class `Loader` which function as an interface for all resources. It holds information such where the `api_methods` are, the name of the `folder` in the template, and the data class used. In addition, it has default implementation of `.create`, `.delete`, `retrieve` and `.load_file`. Then, each supported resource type inherits from it and overrides what is special about that type.

I have skipped everything that has to do with  data models as I think there are a few uncleared questions.
* I think we should split nodes from data models, views, and containers in the build directory.
* We probably need to introduce a schema class for views + containers (and data models). Likely, should be in the SDK.
* There should be naming convention for `.graphql` schemas to infer space and data model name. Instead of having it hard coded to `datamodel.graphql`.
* The change logic for data models needs to either be replicated for all resource types, or removed. 
* I think we should avoid implicit `spaces` but instead require that they are always explicitly stated.


cognite_toolkit/cdf.py
```diff
@@ -278,43 +255,20 @@ def deploy(
...
-            directory=f"{build_dir}/transformations",
-        )
-        if ToolGlobals.failed:
-            print("[bold red]ERROR: [/] Failure to load transformations as expected.")
-            exit(1)
-    if CDFDataTypes.data_models in include and (models_dir := Path(f"{build_dir}/data_models")).is_dir():
+
+    for folder_name, LoaderCls in LOADER_BY_FOLDER_NAME.items():
+        if folder_name in include and (directory := (Path(build_dir) / folder_name)).is_dir():
+            load_resources(
```
gregertw, line 261: Running with `./cognite_toolkit/cdf.py deploy --drop --drop-data --env=local ./build ` I get a failure on timeseries as data_set_id in ToolGlobals is None. Line 164 in load.py

gregertw: `  File "/Users/wedel/src/cdf-project-templates/cognite_toolkit/cdf_tk/load.py", line 164, in get_required_capability
    TimeSeriesAcl.Scope.DataSet([ToolGlobals.data_set_id]),
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<string>", line 4, in __init__
  File "/Users/wedel/src/cdf-project-templates/.venv/lib/python3.11/site-packages/cognite/client/data_classes/capabilities.py", line 330, in __post_init__
    object.__setattr__(self, "ids", [int(i) for i in self.ids])
                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/wedel/src/cdf-project-templates/.venv/lib/python3.11/site-packages/cognite/client/data_classes/capabilities.py", line 330, in <listcomp>
    object.__setattr__(self, "ids", [int(i) for i in self.ids])
                                     ^^^^^^
TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'`

cognite_toolkit/cdf_tk/load.py
```diff
@@ -60,249 +74,119 @@ def __next__(self):
...
+    def __init__(self, client: CogniteClient):
+        self.client = client
+        self.api_class = self._get_api_class(client, self.api_name)
+
+    @staticmethod
+    def _get_api_class(client, api_name: str):
+        parent = client
+        if (dot_count := Counter(api_name)["."]) == 1:
+            parent_name, api_name = api_name.split(".")
+            parent = getattr(client, parent_name)
```
gregertw, line 115: Should you catch the exception here?

cognite_toolkit/cdf_tk/load.py
```diff
@@ -60,249 +74,119 @@ def __next__(self):
...
-        print("[bold red]ERROR:[/] Failed to upload timeseries.")
-        print(e)
-        ToolGlobals.failed = True
-        return
-    print(f"  Created {len(timeseries)} timeseries from {len(files)} files.")
+            raise ValueError(f"Invalid api_name {api_name}.")
+        return getattr(parent, api_name)
+
+    @classmethod
+    def create_loader(cls, ToolGlobals: CDFToolConfig):
```
gregertw, line 123: Are there any reasons for why we shouldn't push verify_client() and verify_dataset() (and then verify_space) into CogniteClient? We have the standardised env variable loading that also auth verify relies on, but that's a create_from_environment() cls method, right? I was wondering if it's worth doing that now as part of this refactoring. Having CDFToolConfig as another abstraction layer complicates understanding the code.

cognite_toolkit/cdf_tk/load.py
```diff
@@ -60,249 +74,119 @@ def __next__(self):
...
+        raise NotImplementedError(f"get_required_capability must be implemented for {cls.__name__}.")
+
+    @classmethod
+    @abstractmethod
+    def get_id(cls, item: T_Resource) -> T_ID:
+        raise NotImplementedError
+
+    # Default implementations that can be overridden
+    def create(
+        self, items: T_Resource | Sequence[T_Resource], ToolGlobals: CDFToolConfig, drop: bool
```
gregertw, line 139: Have you ditched the drop functionality? We have --drop and --drop-data. I think we should introduce a diffing as default create behaviour, i.e. we should retrieve the yaml and compare and upsert where we can. However, having a --drop flag that will actively delete the resources before creating again, is a needed functionality. E.g. where you want to clean out transformation job history or be absolutely sure that you are starting from scratch. I think we should ditch --drop-data now and rely on cdf-tk clean only.

cognite_toolkit/cdf_tk/load.py
```diff
@@ -313,120 +197,260 @@ def load_transformations(
...
+    @classmethod
+    def get_required_capability(cls, ToolGlobals: CDFToolConfig) -> Capability:
+        return RawAcl([RawAcl.Action.Read, RawAcl.Action.Write], RawAcl.Scope.All())
+
+    @classmethod
+    def get_id(cls, item: T_Resource) -> T_ID:
+        raise NotImplementedError("Raw data does not have an id")
+
+    def load_file(self, filepath: Path, ToolGlobals: CDFToolConfig) -> pd.DataFrame:
+        for pattern in [r"(\d+)\.(\w+)\.(\w+)", r"(\d+)\.(\w+)"]:
```
gregertw, line 320: Should we introduce explicit definition of raw tables through yaml files instead of reimplementing the old implicit based on csv data files? Declaring desired raw dbs and tables also simplifies error handling.

cognite_toolkit/cdf_tk/load.py
```diff
@@ -313,120 +197,260 @@ def load_transformations(
...
+            print(f"[bold red]ERROR:[/] Filename {filepath.name} does not match expected format.")
+            ToolGlobals.failed = True
+            return pd.DataFrame()
+        if filepath.suffix == ".csv":
+            # The replacement is used to ensure that we read exactly the same file on Windows and Linux
+            file_content = filepath.read_bytes().replace(b"\r\n", b"\n").decode("utf-8")
+            df = pd.read_csv(io.StringIO(file_content), dtype=str)
+            df.fillna("", inplace=True)
+            return df
+        elif filepath.suffix == ".parquet":
```
gregertw, line 335: The additional support for parquet (other things?) should be added to the docs.

cognite_toolkit/cdf_tk/load.py
```diff
@@ -313,120 +197,260 @@ def load_transformations(
...
+        return f"{cls.id_prefix}_{item.name}"
+
+    def load_file(self, filepath: Path, ToolGlobals: CDFToolConfig) -> Path:
+        return filepath
+
+    def create(
+        self, items: Path | Sequence[Path], ToolGlobals: CDFToolConfig, drop: bool
+    ) -> T_Resource | T_ResourceList | None:
+        for item in items:
+            self.client.files.upload(
```
gregertw, line 386: One could argue that we should add support for explicitly declaring files as well, so that asset mapping can be done. That's more of a feature request though, but maybe worth capturing as a TODO or jira?

cognite_toolkit/cdf_tk/load.py
```diff
@@ -313,120 +197,260 @@ def load_transformations(
...
+        for item in batch:
+            # Set the context info for this CDF project
+            if hasattr(item, "data_set_id"):
+                item.data_set_id = ToolGlobals.data_set_id
+            if drop:
+                drop_items.append(loader.get_id(item))
+        try:
+            if drop:
+                if not dry_run:
+                    loader.delete(drop_items)
```
gregertw, line 434: Ah, ok. You added support for drop here.

gregertw (commented):
This is really great! I have commented throughout, and I was not able to run it due to failure on timeseries, so if you can fix, I'll do another pass :)

cognite_toolkit/cdf.py
doctrino: See it. An argument in favor of `mypy` which would have caught this one.

cognite_toolkit/cdf_tk/load.py
doctrino: Agreed, I have just not prioritized it yet.

cognite_toolkit/cdf_tk/load.py
doctrino: Lets use Jira, do not like todos in code (as I think they should instead just be fixed immediately), and this sounds like a feature.

cognite_toolkit/cdf_tk/load.py
doctrino: Where do we have the different formats documented?

cognite_toolkit/cdf_tk/load.py
doctrino: I agree on this, same argument as I have for `space`s. But lets do that in another PR to keep this as pure refactoring.

cognite_toolkit/cdf_tk/load.py
```diff
@@ -438,43 +516,66 @@ def load_resources(
 
     items: Iterable[Sequence[T_Resource]]
     if loader.load_files_individually:
-        items = ([loader.load_file(f, ToolGlobals)] for f in files)
+        items = [loader.load_file(f, ToolGlobals) for f in files]
```
doctrino, line 519: Changes this from a generator to a list, means we load everything and the `load_files_individually` does not load each file, upload it to CDF, before the next.

tests/test_approval_modules_snapshots/cdf_apm_simple.yaml
```diff
@@ -326,15 +326,6 @@ DataModel:
...
     version: '1'
-Database:
-- name: apm_simple
-- name: apm_simple
-- name: apm_simple
-- name: apm_simple
-- name: apm_simple
-- name: apm_simple
-- name: apm_simple
-- name: apm_simple
```
doctrino, line 337: Database is no longer explicitly created. Instead it is included when uploading the raw tables with the ensure_parent arg.

tests/test_approval_modules_snapshots/cdf_apm_simple.yaml
```diff
@@ -1615,6 +1606,17 @@ deleted:
...
+  - db_name: apm_simple
+    name:
+    - asset2children
+    - assets
+    - timeseries2assets
+    - workitem2assets
+    - workitems
+    - workorder2assets
+    - workorder2items
+    - workorders
```
doctrino, line 1619: Now drop is supported for Raw tables.

gregertw (approved):
All tests are working. I have thoroughly tested manually as well.
Question remains on whether data sets are created before groups?

PR #159, opened 2023-11-19
Extended test for load raw, node, graphql, files, and datapoints

doctrino:
Several of the loaders were not properly tested, which means refactoring them would be a high risk. 

This is necessary for #157 

In addition, fixed a bug for the time series approval tests.


cognite_toolkit/cdf_tk/load.py
```diff
@@ -103,8 +104,10 @@ def load_raw(
                 print(f"[bold red]ERROR:[/] Filename {f} does not match expected format.")
                 ToolGlobals.failed = True
                 return
-        with open(f"{directory}/{f}") as file:
-            dataframe = pd.read_csv(file, dtype=str)
+        with open(f"{directory}/{f}", mode="rb") as file:
+            # The replacement is used to ensure that we read exactly the same file on Windows and Linux
+            file_content = file.read().replace(b"\r\n", b"\n").decode("utf-8")
```
gregertw, line 109: Great!

tests/conftest.py
```diff
@@ -112,6 +135,61 @@ def create(*args, **kwargs) -> Any:
         state[resource_cls.__name__].extend(created)
         return write_list_cls(created)
 
+    def insert_dataframe(*args, **kwargs) -> None:
```
gregertw, line 138: Not obvious to me what this function does? Maybe a short doc string?

gregertw (approved):
Really great!

PR #164, opened 2023-11-21
Extended approval test with drop = True

doctrino:
**Bonus** Found a bug in the TimeSeries configuration.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -112,7 +112,9 @@ def cognite_client_approval() -> CogniteClient:
             if deleted_resources:
                 dumped["deleted"] = {}
                 for key in sorted(deleted_resources):
-                    dumped["deleted"][key] = sorted(deleted_resources[key])
+                    dumped["deleted"][key] = sorted(
+                        deleted_resources[key], key=lambda x: x.get("externalId") if isinstance(x, dict) else x
+                    )
 
             return dumped
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -110,7 +110,10 @@ def cognite_client_approval() -> CogniteClient:
                         key=lambda x: x.get("externalId", x.get("name")),
                     )
             if deleted_resources:
-                dumped["deleted"] = dict(sorted(deleted_resources.items()))
+                dumped["deleted"] = {}
+                for key in sorted(deleted_resources):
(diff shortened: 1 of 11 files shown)


tests/test_approval_modules.py
```diff
@@ -109,7 +109,7 @@ def fake_read_yaml_files(
             build_dir=str(local_tmp_path),
             build_env="test",
             interactive=False,
-            drop=False,
+            drop=True,
```
gregertw, line 112: Do we then always test with drop=True or do we also run the tests with drop=False? A bit hard to say without going outside the code review.

tests/conftest.py
```diff
@@ -64,34 +66,51 @@ def cognite_client_approval() -> CogniteClient:
...
+        client.time_series = create_mock_api(TimeSeriesAPI, TimeSeriesList, written_resources, deleted_resources)
+        client.raw.databases = create_mock_api(RawDatabasesAPI, DatabaseList, written_resources, deleted_resources)
+        client.transformations = create_mock_api(
+            TransformationsAPI, TransformationList, written_resources, deleted_resources
+        )
         client.transformations.schedules = create_mock_api(
-            TransformationSchedulesAPI, TransformationScheduleList, state
+            TransformationSchedulesAPI, TransformationScheduleList, written_resources, deleted_resources
+        )
+        client.data_modeling.containers = create_mock_api(
```
gregertw, line 82: How do we treat deletion of containers and spaces now with the refactoring if there are instances? Will they fail? I believe we originally had that they failed unless --delete-data flag was present, then instances were cleaned out. If we go with my suggestion to only delete data with cdf-tk clean, it makes sense to fail deletion of spaces and containers if there are instances. Our current approach is based on default write and assume nothing exists, thus we need --drop to delete before writing. I assume we will want to change the default behaviour to upsert where there are changes (by comparing yamls). Then --drop will continue to work as today: drop the config before upserting, thus upsert becomes a write.

gregertw (approved):
Added a couple of comments, but approving.

tests/test_approval_modules.py
doctrino: Drop is behind a if statement, so I do not think there is much of a value to test with drop=False for this type of test. It was a mistake by me earlier to not set it to true.

PR #173, opened 2023-11-24
Gtw/improve_apm_simple

gregertw:
* Split apm_simple into data models and oid raw data
* Fix building and loading of files
* Fix data_set loading


cognite_toolkit/cdf_tk/load.py
```diff
@@ -226,6 +226,15 @@ def get_id(self, item: DataSet) -> str:
     def delete(self, ids: Sequence[str]) -> None:
         raise NotImplementedError("CDF does not support deleting data sets.")
 
+    def create(
+        self, items: Sequence[T_Resource], ToolGlobals: CDFToolConfig, drop: bool, filepath: Path
+    ) -> T_ResourceList | None:
```
doctrino, line 231:
```suggestion
    ) -> T_Resource | T_ResourceList | None:
```
You only return a single in file, the `drop_load_resources` function must be updated to ensure it supports it.

cognite_toolkit/cdf_tk/load.py
```diff
@@ -226,6 +226,15 @@ def get_id(self, item: DataSet) -> str:
...
+        try:
+            return self.api_class.create(items)
+
+        except CogniteDuplicatedError:
+            return None
```
doctrino, line 236: This is risky. For example, if you try to create 100 transformations and one of them already exists. All of them will fail, and they will silently not be created.

doctrino (commented):
Few things

cognite_toolkit/cdf_tk/load.py
```diff
@@ -495,18 +504,25 @@ def get_required_capability(cls, ToolGlobals: CDFToolConfig) -> Capability:
...
+        return file
 
     def create(
         self, items: Sequence[FileMetadata], ToolGlobals: CDFToolConfig, drop: bool, filepath: Path
-    ) -> FileMetadataList:
+    ) -> FileMetadata:
         if len(items) != 1:
             raise ValueError("Files must be loaded one at a time.")
         meta = items[0]
         datafile = filepath.parent / meta.name
```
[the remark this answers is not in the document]
gregertw: No, not exactly :)  I assume the name declared in the yaml matches the filename on disk, hence meta.name as parsed from the file. The reason is to avoid that you get a filename of file <--> filename of yaml <--> name in yaml mismatch issue. The yaml can then actually have any name (or could potentially be a single file? maybe better?), so you need to declare the name of each file in the yaml and then the metadata.

cognite_toolkit/cdf_tk/load.py
gregertw: No, it's overloaded create for datasets only. But we should then probably enforce one by one creation?

cognite_toolkit/cdf_tk/load.py
gregertw: Right.

gregertw: Hm, this is datasets loader?

cognite_toolkit/cdf_tk/load.py
```diff
@@ -495,18 +504,25 @@ def get_required_capability(cls, ToolGlobals: CDFToolConfig) -> Capability:
...
-        self.client.files.delete(external_id=[item.external_id for item in items])
+    def delete(self, ids: Sequence[str]) -> None:
+        self.client.files.delete(external_id=ids)
+
+    def load_file(self, filepath: Path, ToolGlobals: CDFToolConfig) -> FileMetadata:
+        file = FileMetadata.load(load_yaml_inject_variables(filepath, ToolGlobals.environment_variables()))
+        if not Path(filepath.parent / file.name).exists():
+            raise FileNotFoundError(f"Could not find file {file.name} referenced in filepath {filepath.name}")
+        if file.data_set_id is not None:
+            file.data_set_id = ToolGlobals.verify_dataset(file.data_set_id)
```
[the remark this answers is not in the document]
gregertw: Which load method?

cognite_toolkit/cdf_tk/load.py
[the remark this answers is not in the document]
doctrino: Good point

cognite_toolkit/cdf_tk/load.py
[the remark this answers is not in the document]
doctrino: Sorry, `drop_load_resources`.

cognite_toolkit/cdf_tk/load.py
doctrino: You can catch the exception and create those that were not duplicates. I can push a commit if you want it.

doctrino (approved):

cognite_toolkit/cdf_tk/load.py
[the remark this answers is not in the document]
gregertw: I'm a bit confused as in you have the same pattern for GroupLoader and I don't see it in drop_load_resources. But if you are ok with how it's now, then great!

PR #180, opened 2023-11-28
Gtw/fix_file_409

gregertw:
## Description

- Require all spaces to be explicitly defined
- Add space yaml files for existing data models
- Fix use of integer value in version for data models for apm_base
- Fix container comparison to detect identical containers
- Add support for drop_data in load_datamodel to delete instances (to fix clean command not working)
- Clean up error on resource does not exist when deleting
- Fix group deletion on use of clean command to actually delete
- Add protection on group deletion and skip any groups that service principal belong to

## Checklist:
- [ ] Tests added/updated.
- [ ] Documentation updated.

### DIFF
--- a/cognite_toolkit/cdf_tk/load.py
+++ b/cognite_toolkit/cdf_tk/load.py
@@ -64,7 +64,7 @@ from cognite.client.data_classes.data_modeling import (
     ViewId,
 )
 from cognite.client.data_classes.iam import Group, GroupList
-from cognite.client.exceptions import CogniteAPIError, CogniteDuplicatedError
+from cognite.client.exceptions import CogniteAPIError, CogniteDuplicatedError, CogniteNotFoundError
 from rich import print
 
 from .delete import delete_instances
@@ -639,10 +639,13 @@ def drop_load_resources(
                 loader.delete(drop_items)
                 if verbose:
                     print(f"  Deleted {len(drop_items)} {loader.api_name}.")
+            except CogniteAPIError as e:
+                if e.code == 404:
+                    print(f"  [bold yellow]WARNING:[/] {len(drop_items)} {loader.api_name} do not exist.")
+            except CogniteNotFoundError:
+                print(f"  [bold yellow]WARNING:[/] {len(drop_items)} {loader.api_name} do not exist.")
             except Exception as e:
-                print(
-                    f"  [bold yellow]WARNING:[/] Failed to delete {len(drop_items)} {loader.api_name}. It/they may not exist. Error {e}"
-                )
+                print(f"  [bold yellow]WARNING:[/] Failed to delete {len(drop_items)} {loader.api_name}. Error {e}")
         else:
             print(f"  Would have deleted {len(drop_items)} {loader.api_name}.")
     if not load:
--- a/cognite_toolkit/cdf_tk/load.py
+++ b/cognite_toolkit/cdf_tk/load.py
@@ -197,8 +197,9 @@ class Loader(ABC, Generic[T_ID, T_Resource, T_ResourceList]):
     ) -> T_ResourceList:
         return self.api_class.create(items)
 
-    def delete(self, items: Sequence[T_Resource]) -> None:
-        return self.api_class.delete(items)
+    def delete(self, items: Sequence[T_Resource]) -> int:
+        self.api_class.delete(items)
+        return len(items)
 
     def retrieve(self, ids: Sequence[T_ID]) -> T_ResourceList:
         return self.api_class.retrieve(ids)
@@ -232,8 +233,9 @@ class TimeSeriesLoader(Loader[str, TimeSeries, TimeSeriesList]):
     def retrieve(self, ids: Sequence[str]) -> TimeSeriesList:
         return self.client.time_series.retrieve_multiple(external_ids=ids, ignore_unknown_ids=True)
 
-    def delete(self, ids: Sequence[str]) -> None:
+    def delete(self, ids: Sequence[str]) -> int:
         self.client.time_series.delete(external_id=ids, ignore_unknown_ids=True)
+        return len(ids)
 
 
 @final
@@ -254,7 +256,7 @@ class DataSetsLoader(Loader[str, DataSet, DataSetList]):
     def get_id(self, item: DataSet) -> str:
         return item.external_id
 
-    def delete(self, ids: Sequence[str]) -> None:
+    def delete(self, ids: Sequence[str]) -> int:
         raise NotImplementedError("CDF does not support deleting data sets.")
 
     def create(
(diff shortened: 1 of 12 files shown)


cognite_toolkit/modules/cdf_infield_common/auth/applications-configuration.yaml
```diff
@@ -2,7 +2,6 @@ name: 'applications-configuration'
 sourceId: '{{applicationsconfiguration_source_id}}'
 metadata:
   origin: 'cdf-project-templates'
-  cdf_template_version: '{{cdf_template_version}}'
```
doctrino, line 5: Any reason why you are removing this?

cognite_toolkit/cdf_tk/load.py
```diff
@@ -198,8 +199,9 @@ def create(
     ) -> T_ResourceList:
         return self.api_class.create(items)
 
-    def delete(self, items: Sequence[T_Resource]) -> None:
-        return self.api_class.delete(items)
+    def delete(self, items: Sequence[T_Resource]) -> int:
```
doctrino, line 202: Shouldn't delete methods take in IDs and not items?
```suggestion
    def delete(self, items: Sequence[T_ID]) -> int:
```

cognite_toolkit/cdf_tk/load.py
```diff
@@ -419,6 +423,34 @@ def retrieve(self, ids: Sequence[T_ID]) -> T_ResourceList:
         found = [g for g in remote if g.name in ids]
         return found
 
+    def delete(self, ids: Sequence[T_Resource]) -> int:
```
doctrino, line 426:
```suggestion
    def delete(self, ids: Sequence[int]) -> int:
```

cognite_toolkit/cdf_tk/load.py
```diff
@@ -419,6 +423,34 @@ def retrieve(self, ids: Sequence[T_ID]) -> T_ResourceList:
...
+            groups = self.client.iam.groups.list().data
+        except Exception as e:
+            print(
+                f"[bold red]ERROR:[/] Failed to retrieve the current service principal's groups. Aborting group deletion.\n{e}"
+            )
+            return
+        my_source_ids = []
+        for g in groups:
+            if g.source_id not in my_source_ids:
+                my_source_ids.append(g.source_id)
```
doctrino, line 438: O(1) instead of O(n) lookup
```suggestion
        my_source_ids = set()
        for g in groups:
            if g.source_id not in my_source_ids:
                my_source_ids.add(g.source_id)
```

doctrino (approved):
just a few minor comments

cognite_toolkit/cdf_tk/load.py
```diff
@@ -125,13 +125,15 @@ class Loader(ABC, Generic[T_ID, T_Resource, T_ResourceList]):
 
     support_drop = True
     support_upsert = False
+    mode: Literal["load", "all_only"] = "load"
```
doctrino, line 128: Why do you add `mode` to the interface? It is only used by the `AuthLoader` so it should be removed from here.

cognite_toolkit/cdf_tk/load.py
```diff
@@ -374,9 +382,12 @@ class AuthLoader(Loader[int, Group, GroupList]):
     )
 
     def __init__(
-        self, client: CogniteClient, target_scopes: Literal["all", "resource_scoped_only", "all_scoped_only"] = "all"
+        self,
+        client: CogniteClient,
+        mode: Literal["load", "all_only"] = "load",
+        target_scopes: Literal["all", "resource_scoped_only", "all_scoped_only"] = "all",
```
doctrino, line 388: `target_scopes` is obsolete if you start to use `mode`. However, I think you should keep `target_scope` and remove the `mode`.

cognite_toolkit/cdf_tk/load.py
```diff
@@ -412,18 +424,23 @@ def load_file(self, filepath: Path, ToolGlobals: CDFToolConfig) -> Group:
...
-                    ]
+                    if self.mode != "all_only":
+                        values["scope"]["datasetScope"]["ids"] = [
+                            ToolGlobals.verify_dataset(ext_id)
+                            for ext_id in values.get("scope", {}).get("datasetScope", {}).get("ids", [])
+                        ]
+                    else:
+                        # If we are running a clean, this is a no-op, but it needs to be valid
+                        values["scope"]["all"] = {}
+                        values["scope"].pop("datasetScope")
```
doctrino, line 435: if you pop all scopes, you need to skip the `capability`

cognite_toolkit/cdf_tk/load.py
```diff
@@ -412,18 +424,23 @@ def load_file(self, filepath: Path, ToolGlobals: CDFToolConfig) -> Group:
...
         return Group.load(raw)
```
doctrino, line 436: We need to handle empty dict here now, as if we skip scopes -> we need to skip capabilities -> we may need to skip the group (in case it only has dataSetScope capabilities.

doctrino (commented):
What is the intention with the new `mode` variable in the `Loader` interface?

cognite_toolkit/cdf_tk/load.py
gregertw: I'm popping only the datasetScope, not "all" that I set on the line before. I'm swapping out datasetscope with all scope.

cognite_toolkit/cdf_tk/load.py
gregertw: We are not skipping scopes, we are replacing datasetscopes with scope all. The whole point is to avoid that parsing of group with datasetscopes fail if datasets don't exist in the CDF project, but we need all the groups as we don't know which ones exist, only that the groups with datasetscopes where the dataset does not exist, will not exist themselves.

cognite_toolkit/cdf_tk/load.py
```diff
@@ -402,11 +401,12 @@ def fixup_resource(local: T_Resource, remote: T_Resource) -> T_Resource:
     def create_loader(
         cls,
         ToolGlobals: CDFToolConfig,
-        mode: Literal["load", "all_only"] = "load",
-        target_scopes: Literal["all", "resource_scoped_only", "all_scoped_only"] = "all",
+        target_scopes: Literal[
+            "all", "all_ignore_dataset", "all_scoped_ignore_dataset", "resource_scoped_only", "all_scoped_only"
```
doctrino, line 405: Any reason why we need `all_ignore_dataset` and `all_scoped_ignore_dataset`? These sounds exactly the same. 

In addition, is there a case where you would like to create all other scopes except for `dataset`? Isn't `all_scoped_only` enough?

tests/test_approval_modules_snapshots/cdf_infield_location.yaml
```diff
@@ -7,47 +7,6 @@ DataSet:
   externalId: ds_infield_default_location_source
   name: Infield default location source data set
 Group:
-- capabilities:
```
doctrino, line 2: I agree, it looks like the test before was wrong and that module `cdf_infield_location` should not generate any capability that is `all` scoped.

doctrino, line 2: I agree, it looks like the test before was wrong and that module `cdf_infield_location` should not generate any capability that is `all` scoped.

cognite_toolkit/cdf_tk/load.py
doctrino: I read through you argument in slack once more, the above comment was a first impression, I suggest we have four options for target scope, `all`, `resource_scoped_only`, `all_scoped_only`, `skipped_validation`. The `skipped_validation` will then be used for the clean command. Wdyt?

PR #182, opened 2023-11-28
Pr/dependency bug

ronpal:
## Description

## Checklist:
- [ ] Tests added/updated.
- [ ] Documentation updated.


cognite_toolkit/cdf.py
```diff
@@ -307,6 +309,32 @@ def deploy(
             verbose=ctx.obj.verbose,
         )
 
+    if "raw" in include and (directory := (Path(build_dir) / "raw")).is_dir():
```
gregertw, line 312: Didn't @doctrino pull out data_sets to be created first in another PR? Is it then necessary to split out these into the two separate? Because of the RAW dependency? @doctrino, I guess we should rather introduce a LOADER_ORDER ordered set to use so we don't have to duplicate so much code?

cognite_toolkit/cdf_tk/load.py
```diff
@@ -403,16 +406,42 @@ def get_required_capability(cls, ToolGlobals: CDFToolConfig) -> Capability:
...
+    def replace_scope_ids(self, capability: dict, ToolGlobals: CDFToolConfig):
+        if len(capability.get("scope", {}).get("datasetScope", {}).get("ids", [])) > 0:
+            capability["scope"]["datasetScope"]["ids"] = [
+                ToolGlobals.verify_dataset(ext_id)
+                for ext_id in capability.get("scope", {}).get("datasetScope", {}).get("ids", [])
+            ]
+
+        if len(capability.get("scope", {}).get("extractionPipelineScope", {}).get("ids", [])) > 0:
+            capability["scope"]["extractionPipelineScope"]["ids"] = [
+                ToolGlobals.verify_extraction_pipeline(ext_id)
```
gregertw, line 418: Here we get into the same issue as I had in #180 if we want to do a clean command and the target scope does not exist. We want to be able to load all extraction pipelines, also the ones where we don't have the scope stored in the CDF project as we need a complete list to be used for deletion (a later pipeline may be present and should be deleted).

cognite_toolkit/cdf_tk/load.py
```diff
@@ -403,16 +406,42 @@ def get_required_capability(cls, ToolGlobals: CDFToolConfig) -> Capability:
...
+
         for capability in raw.get("capabilities", []):
-            for _, values in capability.items():
-                if len(values.get("scope", {}).get("datasetScope", {}).get("ids", [])) > 0:
-                    values["scope"]["datasetScope"]["ids"] = [
-                        ToolGlobals.verify_dataset(ext_id)
-                        for ext_id in values.get("scope", {}).get("datasetScope", {}).get("ids", [])
-                    ]
-        return Group.load(raw)
+            for acl, values in capability.items():
```
gregertw, line 431: Not sure what this section is relative to what I have done in #180 , but this can become really messy to merge. But maybe the pattern in #180 can be reused?

gregertw (commented):
I wasn't aware that you were touching some of the same code as I was in #180. We have been working on totally different problems, but touched on some of the same root challenges. #180 is pretty big and pretty much done, so we should probably merge that one first, and then I can help with merging in main.

ronpal:
> I wasn't aware that you were touching some of the same code as I was in #180. We have been working on totally different problems, but touched on some of the same root challenges. #180 is pretty big and pretty much done, so we should probably merge that one first, and then I can help with merging in main.

No prob, let's do that!

cognite_toolkit/cdf.py
ronpal: Yes, I think it is ugly. Anders and I discussed it yesterday. We have to do  `auth without recourceScope` then `dataset` then `raw` then `extraction_pipelines` before `FDM`and `the rest`, and then finish `auth with resourceScope`. But it violates DRY bigtime. Hardcoding a "load order" might fix this particular case, but it wouldn't surprise me if there were cases where the order has to be different. Anyways, open to suggestions.

PR #187, opened 2023-11-29
Refactor; Data Model Loaders

doctrino:
## Description
Combined the `delete_space` and `delete_containers` into `drop_data` . This keeps the same functionality from a `CLI` perspective, and enables standardization of the `load_data_models`. 

## Checklist:
- [x] Tests added/updated.
- [ ] Documentation updated.
- [x] Run Demo Job Locally.


cognite_toolkit/experimental/cdf_asset_source_model/data_models/2.ModelSpace.space.yaml
```diff
@@ -0,0 +1,3 @@
+space: {{model_space}}
```
doctrino, line 1: Now spaces must be defined explicitly

cognite_toolkit/modules/cdf_infield_location/data_models/infield_apm_app_config.node.yaml
```diff
@@ -1,23 +1,25 @@
 autoCreateDirectRelations: True
 skipOnVersionConflict: False
 replace: True
```
doctrino, line 3: Changed the format for nodes to match the API, to avoid this custom format.

cognite_toolkit/cdf_tk/load.py
```diff
@@ -853,397 +1174,9 @@ def drop_load_resources(
...
-    explicit_space_list = [s.space for s in cognite_resources_by_type["space"]]
-    space_list = list({r.space for _, resources in cognite_resources_by_type.items() for r in resources})
-
-    implicit_spaces = [SpaceApply(space=s, name=s, description="Imported space") for s in space_list]
-    for s in implicit_spaces:
-        if s.name not in [s2.name for s2 in cognite_resources_by_type["space"]]:
-            print(
-                f"  [bold red]ERROR[/] Space {s.name} is implicitly defined and may need it's own {s.name}.space.yaml file."
-            )
-            cognite_resources_by_type["space"].append(s)
```
doctrino, line 971: Implicit spaces only gave this warning, and was not rejected.

doctrino:
> Are you actually trying this out locally beyond running tests? When I do build on a vanilla install, the build fails on missing externalId. I believe this is because the check_yaml_semantics() doesn't expect multiple nodes, just a single.

Yes, if not I have forgotten it. What is the sequence you are doing? I have tested it again now, and cannot seem to trigger this one.

CHANGELOG.cdf-tk.md
```diff
@@ -43,13 +44,18 @@ Changes are grouped as follows:
...
+
 
 ### Fixed
 
 - `cdf-tk clean` not supporting `--include` properly.
 - `cdf-tk clean` not working properly for data models with data.
 - Fix group deletion on use of clean command to actually delete groups.
 
+### Removed
+- Support for `.graphql` deployment.
```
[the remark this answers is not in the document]
doctrino: You, are right we did not. Will remove it.

cognite_toolkit/cdf.py
```diff
@@ -268,15 +256,17 @@ def deploy(
     # once with all_scoped_skipped_validation and once with resource_scoped_only
     selected_loaders = {
         LoaderCls: LoaderCls.dependencies
-        for folder_name, LoaderCls in LOADER_BY_FOLDER_NAME.items()
+        for folder_name, loader_classes in LOADER_BY_FOLDER_NAME.items()
         if folder_name in include and folder_name != "auth" and (build_path / folder_name).is_dir()
+        for LoaderCls in loader_classes
```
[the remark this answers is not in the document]
doctrino: It does not check the `build` forlder, only what type of resources are possible to run with the toolkit.

cognite_toolkit/cdf.py
```diff
@@ -291,38 +281,15 @@ def deploy(
...
-        )
-        if ToolGlobals.failed:
-            print("[bold red]ERROR: [/] Failure to load data models as expected.")
-            exit(1)
-    if CDFDataTypes.instances.value in include and (models_dir := Path(f"{build_dir}/data_models")).is_dir():
-        load_nodes(
-            ToolGlobals,
-            directory=models_dir,
-            dry_run=dry_run,
+    for LoaderCls in TopologicalSorter(selected_loaders).static_order():
```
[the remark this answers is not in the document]
doctrino: Difference now is that each resource loader have a dependency class variable used to determined the order. To me that is not less clear when you are in the context of a specific resource, instead it makes it more clear. 

The added bonus is that you do not need to repeat the `drop_load_resources` calls for each type of resource we support, and then repeat them again in the opposite order for the delete command.

cognite_toolkit/cdf_tk/load.py
```diff
@@ -845,6 +1168,10 @@ def drop_load_resources(
     else:
         filepaths = [file for file in path.glob("**/*")]
 
+    if loader.filename_pattern:
```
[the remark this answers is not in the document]
doctrino: Added an extra comment to explain it.

cognite_toolkit/cdf_tk/load.py
```diff
@@ -220,8 +220,9 @@ class Loader(ABC, Generic[T_ID, T_Resource, T_ResourceList]):
     list_cls: type[CogniteResourceList]
     dependencies: frozenset[Loader] = frozenset()
 
-    def __init__(self, client: CogniteClient):
+    def __init__(self, client: CogniteClient, ToolGlobals: CDFToolConfig):
         self.client = client
+        self.ToolGlobals = ToolGlobals
```
doctrino, line 225: The `ToolGlobals` was passed into `create` and `load_resource` and now I needed it in `delete`, so I think it is better to have it as part of the state of the Loader.

cognite_toolkit/cdf_tk/load.py
```diff
@@ -835,11 +910,237 @@ def create(
...
+    @classmethod
+    def get_id(cls, item: SpaceApply) -> str:
+        return item.space
+
+    def delete(self, ids: Sequence[str], drop_data: bool) -> int:
+        if not drop_data:
+            print("  [bold]INFO:[/] Skipping deletion of spaces as drop_data flag is not set...")
+            return 0
+        print("[bold]Deleting existing data...[/]")
+        for space in ids:
```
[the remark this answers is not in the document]
doctrino: Yes, that is checked on the line above, the if-statement before: `print("  [bold]INFO:[/] Skipping deletion of spaces as drop_data flag is not set...")`.

In the dependencies, you see that space <-containers <- nodes <- edges, so when you call this for delete, it will go in the opposite direction. We have to then delete the extra instances that are in the space, which are not governed by the toolkit.

This requires that we remove `drop_data` from `deploy` as we discussed before. Will do it in this PR.

cognite_toolkit/cdf_tk/load.py
```diff
@@ -612,16 +693,14 @@ def load_resource(self, filepath: Path, ToolGlobals: CDFToolConfig, dry_run: boo
                     f"Could not find sql file belonging to transformation {filepath.name}. Please run build again."
                 )
         transformation.query = sql_file.read_text()
-        transformation.data_set_id = ToolGlobals.data_set_id
+        transformation.data_set_id = self.ToolGlobals.data_set_id
```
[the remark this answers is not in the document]
doctrino: Agreed, fixed.

.github/workflows/demo.yml
```diff
@@ -46,9 +46,12 @@ jobs:
       - name: "Allow some time for data modeling to finish syncing of deletions"
         run: |
           sleep 30
+      - name: "Clean the templates"
+        run: |
+          cdf-tk clean --env=demo ./build
```
doctrino, line 51: with the removal of `drop_data`, I assume we know want to clean the demo project before redeploying.

gregertw: Yes, but I would prefer to push us a bit here and try to use only --drop and targeted clean, i.e. only clean --include for instances. Or even better later, not use --drop to prefer the upsert and then targeted do drop.

CHANGELOG.cdf-tk.md
```diff
@@ -18,6 +18,12 @@ Changes are grouped as follows:
 ## [TBD] - 2023-12-TBD
 ### Added
 - Warnings if a configuration file is using `snake_case` when then resource type is expecting `camelCase`.
+- Added support for validation of `space` for data models.
+### Removed
+- In the `deploy` command `drop_data` option has been removed. To drop data, use the `clean` command instead.
+### Changed
+- Require all spaces to be explicitly defined as separate .space.yaml file.
+- The `data_set_id` for `Transformations` must now be set explicitly in the yaml config file for the `Transformation`.
```
gregertw, line 26: We should mention the key to use.

gregertw (approved):
Just two minor comments

PR #201, opened 2023-12-07
Added groups capability to avoid error message in missing location co…

tesande:
Added groups capability for infield groups


ronpal (approved):
Approved in agreement with the InField team

PR #248, opened 2024-01-02
[CDF-20588] MyPy Part 2

doctrino:
## Description

Adding `mypy` check to the `cognite_toolkit/cdf_tk/bootstrap.py`. I have done a minimal amount of changes, and instead tried to just add type hints to reflect the actual types that are being used. 

## Checklist:
- [ ] Tests added/updated.
- [ ] Run Demo Job Locally.
- [ ] Documentation updated.

### DIFF
--- a/cognite_toolkit/cdf_tk/bootstrap.py
+++ b/cognite_toolkit/cdf_tk/bootstrap.py
@@ -16,6 +16,7 @@ from __future__ import annotations
 import os
 from dataclasses import dataclass
 from pathlib import Path
+from typing import cast
 
 from cognite.client import CogniteClient
 from cognite.client.data_classes.capabilities import (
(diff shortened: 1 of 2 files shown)


cognite_toolkit/cdf_tk/bootstrap.py
```diff
@@ -32,34 +33,34 @@
...
-    cdf_url: str = None
-    token_url: str = None
-    tenant_id: str = None
-    audience: str = None
-    scopes: str = None
+    cluster: str | None
+    project: str | None
+    token: str | None
+    client_id: str | None
+    client_secret: str | None
```
doctrino, line 40: I set these to `None` to reflect the behavior as it is today. The other option, is to raise an exception in line 54-58 if any of these are missing from the environment.

gregertw: Let's go for None for now. If we raise an exception, we might want to use the typer support for prompting if a command line option is not set either from env var or explicitly.

cognite_toolkit/cdf_tk/bootstrap.py
```diff
@@ -32,34 +33,34 @@
...
     vars = AuthVariables(
-        cluster=os.environ.get("CDF_CLUSTER", None),
-        project=os.environ.get("CDF_PROJECT", None),
-        token=os.environ.get("CDF_TOKEN", None),
-        client_id=os.environ.get("IDP_CLIENT_ID", None),
-        client_secret=os.environ.get("IDP_CLIENT_SECRET", None),
-        cdf_url=os.environ.get("CDF_URL", None),
-        tenant_id=os.environ.get("IDP_TENANT_ID", None),
-        audience=os.environ.get("IDP_AUDIENCE", None),
-        scopes=os.environ.get("IDP_SCOPES", None),
```
gregertw, line 59: Any particular reason for removing the explicitness of setting the var to None? It's default, but the less you need to remember about a specific function, the better IMO

gregertw (approved):
LGTM

PR #249, opened 2024-01-02
Bugfix: Skipping auth

doctrino:
## Description

Ref: https://cognitedata.slack.com/archives/C05RW13C6TX/p1703245850911189

From changelog:

### Fixed

- Fixed bug in `cdf-tk deploy` where auth groups with a mix of all and resource scoped capabilities skipped 
  the all scoped capabilities. This is now fixed.

## Checklist:
- [x] Tests added/updated.
- [ ] Run Demo Job Locally.
- [ ] Documentation updated.
- [x] Changelogs updated in [CHANGELOG.cdf-tk.md](https://github.com/cognitedata/cdf-project-templates/blob/main/CHANGELOG.cdf-tk.md).

### DIFF
--- a/cognite_toolkit/cdf.py
+++ b/cognite_toolkit/cdf.py
@@ -341,7 +341,7 @@ def deploy(
         # Last, we need to get all the scoped access, as the resources should now have been created.
         print("[bold]EVALUATING auth resources scoped to resources...[/]")
         result = deploy_or_clean_resources(
-            AuthLoader.create_loader(ToolGlobals, target_scopes="resource_scoped_only"),
+            AuthLoader.create_loader(ToolGlobals, target_scopes="all"),
             directory,
             **arguments,
         )
--- a/cognite_toolkit/cdf.py
+++ b/cognite_toolkit/cdf.py
@@ -338,7 +338,8 @@ def deploy(
             exit(1)
 
     if "auth" in include and (directory := (Path(build_dir) / "auth")).is_dir():
-        # Last, we need to get all the scoped access, as the resources should now have been created.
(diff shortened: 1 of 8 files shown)


cognite_toolkit/cdf.py
```diff
@@ -341,7 +341,7 @@ def deploy(
         # Last, we need to get all the scoped access, as the resources should now have been created.
         print("[bold]EVALUATING auth resources scoped to resources...[/]")
         result = deploy_or_clean_resources(
-            AuthLoader.create_loader(ToolGlobals, target_scopes="resource_scoped_only"),
+            AuthLoader.create_loader(ToolGlobals, target_scopes="all"),
```
doctrino, line 344: This is the bugfix. The issue is that we are creating new Groups and not updating existing group capabilities. Since we had `resource_scoped_only`, we are filtering out the all scoped capabilities and then deleting and recreating the groups with only resource scoped capabilities. Meaning the all scoped capabilities created on line 319 are lost, and replaced with groups which has only the resource scoped capabilities.

gregertw: I think you need to update comments as well.
I believe the correct behaviour is to do as I initially implemented? I.e. the diff mechanism to ensure that we are not deleting when not necessary. This mechanism seems to crude. Without having mapped out what the various target_scopes actually do in the various scenarioes (clean, deploy) and for different existing states, I must admit I'm unable to immediate see that changes this to "all" will not have unintended effects.

gregertw (commented):
Added a comment on the fix

gregertw (approved):

PR #256, opened 2024-01-02
Add markdownlint to pre-commit

gregertw:


doctrino (approved):
The less time we have to worry about style the better, so all for linting.

PR #273, opened 2024-01-09
[CDF-20641] Generation of Config.yaml Redesign

doctrino:
# Description

**Problem**; Regeneration of `config.yaml` when for example writing `cdf-tk init --upgrade` overwrites all set variables and comments. 

**Issue**: This turned out to be a design flaw. 

**Refactoring**: I redesigned the functions used to generate the `config.yaml` into a single class `ConfigYAML` with appropriate `.load` and `dump` methods. 

**Solution** The newly design class accounts for existing values in the existing `config.yaml` and will avoid overwriting them and their comments. 

## Checklist

- [ ] Tests added/updated.
- [ ] Run Demo Job Locally.
- [ ] Documentation updated.
- [ ] Version bumped.


cognite_toolkit/cdf_tk/templates.py
```diff
@@ -447,293 +447,281 @@ def build_config(
...
-                    ConfigEntry(
-                        key=key,
-                        module=module,
-                        path=path_to,
-                        last_value=value,
-                        current_value=None,
-                    )
+    @classmethod
+    def load(cls, directory: Path | Sequence[Path], existing_config_yaml: str | None = None) -> ConfigYAML:
+        """Loads the config.yaml file from a module directory and all its subdirectories.
```
gregertw, line 535: Is this an accurate description of what this method does? From a module directory? Sort of confused me. Looking at the code it looks like you mean default.config.yaml?

tests/test_approval_modules_snapshots/cdf_oid_example_data.yaml
```diff
@@ -27,7 +27,7 @@ Datapoints:
   - pi_163582
   - pi_160781
   - pi_191092
-  dataframe: 42209005129637194408210069515272386610088795145294684171391799791354039528511
+  dataframe: 79727527026717646501089764226733539859120513138805835720018417467989011529987
```
gregertw, line 30: This keeps changing??

cognite_toolkit/cdf_tk/templates.py
```diff
@@ -447,293 +447,281 @@ def build_config(
...
+        return "\n".join(out_lines)
 
     def __str__(self) -> str:
         total_variables = len(self)
         lines = []
         if removed := self.removed:
-            lines.append(f"Removed {len(removed)} variables from config.yaml: {[str(r) for r in removed]}")
+            lines.append(f"Untracked {len(removed)} variables in config.yaml: {[str(r) for r in removed]}")
         if added := self.added:
             lines.append(f"Added {len(added)} variables to config.yaml: {[str(a) for a in added]}")
```
gregertw, line 647: This user experience isn't pleasent (you get a [var1, var2, ...] over multiple lines.

gregertw (commented):
I did init, then edited config.yaml:
- Removed some modules
- Added some comments
- Changed some variables

I did upgrade and observed the following:
- my comments disappeared. I think this is a very valid use case
- A confusing message was shown as part of the output: "Will upgrade modules and files in place, config.yaml files will not be touched."
- The messages "  All default variables from the modules have been upgraded.
  Please check you config.yaml file for new default variables that may need to be changed." --> are these still valid as we don't change values that have been edited. Also, you should get a list now, so you don't have to check yourself?
- my var changes were intact
- all modules I had deleted were just added
- the output was cluttered (just a list dump)
- the message for each variable was confusing. Would be better with something like: "Variable Y in module X was added with value "<value>"

This is such a great improvement that I suggest you take the additional time to add a flag to init "--upgrade --add-new-modules" and not add modules that are not present in config.yaml without it.

Also, I would seriously evaluate adding a "--override-vars" flag to overwrite all values with defaults as this will greatly simplify identifying what is being upgraded.

tests/test_approval_modules_snapshots/cdf_oid_example_data.yaml
doctrino: Yes, it is a bug in the snapshot testing. I have a task on it, but have not gotten to it yet, so for now just regen when needed.

cognite_toolkit/cdf_tk/templates.py
doctrino: You are right I have removed them. I have some extra print if the user adds --verbose.

cognite_toolkit/cdf_tk/templates.py
doctrino: Improved it now.

gregertw (approved):
Great! Works nicely now

PR #276, opened 2024-01-11
[CDF-20672] Consistent YAML file endings.

doctrino:
# Description
I changed the file endings to all flies as well, but seems that does not show up as diffs in GitHub

## Checklist

- [x] Tests added/updated.
- [ ] Run Demo Job Locally.
- [ ] Documentation updated.
- [ ] Version bumped.


gregertw (commented):
Isn't this a bit upside down way of supporting CRLF vs LF? It is not a standard thing to do (to think about and change this for files). It also puts the burden on contributors to change the line endings instead of the diff being ignorant of CRLF vs LF. Also, some editors will change the endings (dependent on platform), so this seems fairly messy to have to deal with.
Is there something I don't see here?

PR #312, opened 2024-02-02
Gtw/run_function_locally

gregertw:
# Description

Add support for run function --local to run function locally in a virtual environment.
## Checklist

- [x] Tests added/updated.
- [ ] Run Demo Job Locally.
- [ ] Documentation updated.
- [x] Changelogs updated in [CHANGELOG.cdf-tk.md](https://github.com/cognitedata/cdf-project-templates/blob/main/CHANGELOG.cdf-tk.md).
- [ ] Version bumped.


cognite_toolkit/cdf_tk/templates/_templates.py
```diff
@@ -301,7 +301,9 @@ def process_function_directory(
                         )
                     destination = build_dir / "functions" / f"{func.external_id}"
                     if destination.exists():
-                        print(f"        [bold yellow]ERROR:[/] Function {func.external_id} is duplicated.")
+                        print(
+                            f"        [bold yellow]ERROR:[/] Function {func.external_id} is duplicated. If this is unexpected, you want want to use '--clean'."
```
doctrino, line 305:
```suggestion
                            f"        [bold red]ERROR:[/] Function {func.external_id} is duplicated. If this is unexpected, you want want to use '--clean'."
```

cognite_toolkit/common_function_code/tool.py
```diff
@@ -129,23 +128,20 @@ def environment_variables(self) -> dict[str, str | None]:
         return {**self._environ.copy(), **os.environ}
 
     def as_string(self) -> str:
-        environment = self._environ.copy()
+        environment = os.environ
```
doctrino, line 131: Wouldn't this cause the environment to be changed? If you don't do a `.copy`?

cognite_toolkit/cdf_tk/run.py
```diff
@@ -146,3 +158,222 @@ def run_transformation(ToolGlobals: CDFToolConfig, external_ids: str | list[str]
...
+        if not isinstance(root_values, dict):
+            continue
+        for group_module, group_values in root_values.items():
+            if not isinstance(group_values, dict):
+                continue
+            for module_name in group_values:
+                if Path(source_path / f"{root_module}/{group_module}/{module_name}/functions/{external_id}").is_dir():
+                    config.environment.selected_modules_and_packages = [f"{module_name}"]
+                    found = True
+                    break
```
doctrino, line 190: I guess this can be a method on `BuildConfigYAML` with an appropriate name. As it is now, it not very readable.

cognite_toolkit/cdf_tk/run.py
```diff
@@ -146,3 +158,222 @@ def run_transformation(ToolGlobals: CDFToolConfig, external_ids: str | list[str]
...
+        return False
+    build_config(
+        build_dir=Path(build_dir),
+        source_dir=source_path,
+        config=config,
+        system_config=system_config,
+        clean=True,
+        verbose=False,
+    )
+    environment_dir = Path(source_path / f"./.venv.{external_id}")
```
doctrino, line 202: Since we already have the concept of `environment` in the toolkit. 
```suggestion
    virtual_env_dir = Path(source_path / f"./.venv.{external_id}")
```

doctrino (commented):
Nice one!

I am a bit unsure about the workflow. Now it seems you run the a custom build command inside the run command to have this ready. Why not just require that the function you want to test has been built?

gregertw:
> Nice one!
> 
> I am a bit unsure about the workflow. Now it seems you run the a custom build command inside the run command to have this ready. Why not just require that the function you want to test has been built?

Yes, we need to inject a custom main function into the directory, so instead of trying to clean up (and risk errors on the way), it is cleaner this way. Also, it allows to run tests easier. Finally, it is a much smoother work process with one command instead of two.

cognite_toolkit/cdf_tk/run.py
gregertw: Agreed, but it's such a special case thing to modify the environment to only load one package, so I don't feel it belongs in BuildConfigYAML?

cognite_toolkit/common_function_code/tool.py
gregertw: Great catch!

PR #459, opened 2024-04-17
[CDF-21408] convert system exits to raising exceptions

haakonvt:
# DRAFT

## Checklist

- [ ] Tests added/updated.
- [ ] Run Demo Job Locally.
- [ ] Documentation updated.
- [ ] Version bumped.


cognite_toolkit/_cdf.py
```diff
@@ -56,39 +65,28 @@
...
     # From Typer version 0.11 -> 0.12, breaks if you have an existing installation.
-    print(
-        "'cognite-toolkit' uses a dependency named 'typer'.",
-        "From 'typer' version 0.11 -> 0.12 there was a breaking change if you have an existing "
-        "installation of 'typer'.",
-        "Workaround is to uninstall 'typer-slim', and then, reinstall 'typer':",
-        "'pip uninstall typer-slim'",
-        "'pip install typer'",
-        sep="\n",
+    raise ToolkitError(
```
doctrino, line 75: Here we might have to do something with `sentry`. If we raise regular errors they will be picked up by `sentry` and reported. This type of error we do not want `sentry` to report on.

cognite_toolkit/_cdf.py
```diff
@@ -56,39 +65,28 @@
...
-run_app = typer.Typer(
-    pretty_exceptions_short=False, pretty_exceptions_show_locals=False, pretty_exceptions_enable=False
-)
-pull_app = typer.Typer(
-    pretty_exceptions_short=False, pretty_exceptions_show_locals=False, pretty_exceptions_enable=False
-)
-dump_app = typer.Typer(
-    pretty_exceptions_short=False, pretty_exceptions_show_locals=False, pretty_exceptions_enable=False
-)
+app = typer.Typer(**default_typer_kws)
```
doctrino: 🏌

cognite_toolkit/_cdf.py
```diff
@@ -381,8 +377,8 @@ def deploy(
             if result:
                 results[result.name] = result
             if ToolGlobals.failed:
-                print(f"[bold red]ERROR: [/] Failure to clean {LoaderCls.folder_name} as expected.")
-                exit(1)
+                raise ToolkitCleanResourceError(f"Failure to clean {loader_cls.folder_name} as expected.")
```
doctrino, line 380: This type of error has often turned out to b e difficult to debug. Is it a later task to raise it were `ToolGlobals.failed` is set to `True` instead of doing it here. Wdyt?

cognite_toolkit/_cdf_tk/templates/_templates.py
```diff
@@ -85,64 +88,69 @@ def build_config(
     return source_by_build_path
 
 
+# TODO: Refactor this function into validation parts for each 'resource_type':
```
doctrino, line 91: Amen. Has been in the backlog for a long time.

doctrino (commented):
I think the `exception.py` module is incorrectly placed. 

There are two things that I think needs to be addressed

1. Ensure that there is a good print out for the user when the user does a mistake. I am in doubt whether we should include a long traceback for this case.
2. We use `sentry` to automatically send reports if there is an exception raised in toolkit. We need to ensure that there is no report sent for user mistakes, but only when there is an actual bug in the toolkit.

cognite_toolkit/_cdf.py
haakonvt: Addressed by adding a `before_send=sentry_exception_filter` function to filter out our own exceptions

doctrino (commented):
This is a great quality-of-improvement. I think the main missing piece, is that when a user does an error (similar to a 400 type error on an API) the traceback should not be shown (maybe if you set a --verbose flag)

Right now you get the full traceback. In addition, would be nice with some nice rich formatting.

doctrino (approved):

PR #540, opened 2024-05-14
Refactor `check_yaml_semantics` function

haakonvt:
# Tech Debt Reduction Initiative

### DIFF
--- a/cognite_toolkit/_cdf_tk/templates/_templates.py
+++ b/cognite_toolkit/_cdf_tk/templates/_templates.py
@@ -230,11 +230,8 @@ def _get_ext_id_and_type_from_parsed_yaml(
     elif resource is Resource.WORKFLOWS:
         return _extract_ext_id_workflows(*args)
 
-    elif resource is Resource.OTHER:
-        return _extract_ext_id_other(resource, *args)
-
     else:
-        return None, ""
+        return _extract_ext_id_other(resource, *args)
 
 
 def _check_yaml_semantics_auth(ext_id: str, filepath_src: Path, verbose: bool) -> None:
--- a/cognite_toolkit/_cdf_tk/templates/_templates.py
+++ b/cognite_toolkit/_cdf_tk/templates/_templates.py
@@ -370,7 +370,7 @@ def check_yaml_semantics(parsed: dict | list, filepath_src: Path, filepath_build
     if parsed is None or filepath_src is None or filepath_build is None:
         raise ToolkitYAMLFormatError
 
-    resource = Resource(filepath_build.parent.name)
+    resource = Resource(filepath_src.parent.name)
     ext_id, ext_id_type = _get_ext_id_and_type_from_parsed_yaml(resource, parsed, filepath_src)
 
     if ext_id is None:
--- a/cognite_toolkit/_cdf_tk/templates/_templates.py
+++ b/cognite_toolkit/_cdf_tk/templates/_templates.py
@@ -104,16 +104,133 @@ def build_config(
     return source_by_build_path
 
 
-class ResourceType(Enum):
-    DATA_MODELS = "data_models"
+class Resource(Enum):
     AUTH = "auth"
+    DATA_SETS = "data_sets"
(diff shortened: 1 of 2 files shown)


cognite_toolkit/_cdf_tk/templates/_templates.py
```diff
@@ -103,186 +104,280 @@ def build_config(
     return source_by_build_path
 
 
-# TODO: Refactor this function into validation parts for each 'resource_type':
-def check_yaml_semantics(parsed: dict | list, filepath_src: Path, filepath_build: Path, verbose: bool = False) -> bool:
-    """Check the yaml file for semantic errors
+class Resource(Enum):
```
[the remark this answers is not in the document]
doctrino: Something like
```python
def test_resource_enum_is_up_to_date() -> None:
    resource_loaders = {loader.folder_name for loader in RESOURCE_LOADER_LIST}
    enum_members = {member.value for member in Resource}

    missing = resource_loaders - enum_members

    assert not missing, f"Missing entry in ResourceEnum please update with: {missing}"

```

cognite_toolkit/_cdf_tk/templates/_templates.py
```diff
@@ -103,186 +104,280 @@ def build_config(
...
+    @classmethod
+    def _missing_(cls, value: object) -> Resource:
+        """Returns Resource.OTHER for any other value."""
+        assert isinstance(value, str)
+        return cls.OTHER
+
+
+def _extract_ext_id_dm_spaces(parsed: dict | list, filepath_src: Path) -> tuple[str | None, str]:
+    if isinstance(parsed, list):
+        raise ToolkitYAMLFormatError(f"Multiple spaces in one file {filepath_src} is not supported.")
```
doctrino, line 129: Actually we should support this, the rest of the code is written and tested for it. So this seems to be an old assumption. 

Nice to get a cleanup so we see this, because we need to remove this asap and support list of all resources. Ref @ronpal

cognite_toolkit/_cdf_tk/templates/_templates.py
```diff
@@ -103,186 +104,280 @@ def build_config(
...
+        return _extract_ext_id_workflows(*args)
+
+    else:
+        return _extract_ext_id_other(resource, *args)
+
+
+def _check_yaml_semantics_auth(ext_id: str, filepath_src: Path, verbose: bool) -> None:
+    parts = ext_id.split("_")
+    if len(parts) < 2:
+        if ext_id == "applications-configuration":
```
doctrino, line 240: Did we really have such a specific warning in the original code?

cognite_toolkit/_cdf_tk/templates/_templates.py
```diff
@@ -103,186 +104,280 @@ def build_config(
...
+    sql_file1 = filepath_src.parent / f"{filepath_src.stem}.sql"
+    if not sql_file1.exists():
+        # Next try to find the sql file next to the yaml file with the external_id as filename
+        sql_file2 = filepath_src.parent / f"{ext_id}.sql"
+        if not sql_file2.exists():
+            print(f"      {WARN_YELLOW} could not find sql file:")
+            print(f"                 [bold]{sql_file1.name}[/] or ")
+            print(f"                 [bold]{sql_file2.name}[/]")
+            print(f"               Expected to find it next to the yaml file at {sql_file1.parent}.")
+            raise ToolkitYAMLFormatError
```
doctrino, line 269: We now supports sql inside the transformation file. But when adding that support, must have missed this

doctrino (commented):
This is great, uncovering multiple issues already. The only critical comment is to use the build path to get the resource name. Except for that mostly optional suggestions.

doctrino (approved):

cognite_toolkit/_cdf_tk/templates/_templates.py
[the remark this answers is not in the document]
haakonvt: Agreed, added!

PR #550, opened 2024-05-16
Deploy command

ronpal:
# Description

Moved deploy into DeployCommand class

## Checklist

- [ ] Tests added/updated.
- [ ] Run Demo Job Locally.
- [ ] Documentation updated.
- [ ] Version bumped.


tests/tests_unit/utils.py
```diff
@@ -77,7 +77,6 @@ def fake_load_yaml_inject_variables(
     monkeypatch.setattr(
         "cognite_toolkit._cdf_tk.templates.data_classes._project_directory.read_yaml_file", fake_read_yaml_file
     )
-    monkeypatch.setattr("cognite_toolkit._cdf.read_yaml_file", fake_read_yaml_file)
```
ronpal, line 80: This was a bit strange, but it suddenly caused tests to fail. Is it ok to remove?

doctrino: If tests are passing then all good. I guess it was there as we probably used to do reading of yaml files in the `_cdf.py` module. But with you moving that code it is not longer necessary. You can consider to update the path to 
`monkeypatch.setattr("cognite_toolkit._cdf_tk.commands.deploy.read_yaml_file", fake_read_yaml_file)`

doctrino: There is a lager story behind this. When you mock in Python, you have to mock were the function is **used** and not were it is created. Now you are using the `read_yaml_file` in a new place, so I changed my mind from above, you should update the path.

doctrino (approved):
No Warnings??

Pre-approving, but you should update the mock path.

PR #561, opened 2024-05-24
[CDF-21088] 🐫 Warning on duplicated resources

doctrino:
# Description

I chose not to skip duplicates in the build step, and instead leave that for the deploy step. The reason is that skipping becomes complicated as a single YAML file can contain multiple resources, and thus if one of them is a duplicate, what do you do with the rest? It gets complicated to just partially skip a file. 

This is simpler in the deploy step, as there you load all items from all files, and then remove the duplicates. I think just a warning is sufficient for the build step.

## Checklist

- [ ] Tests added/updated.
- [ ] Run Demo Job Locally.
- [ ] Documentation updated.
- [ ] Version bumped.

### DIFF
--- a/cognite_toolkit/_cdf_tk/commands/build.py
+++ b/cognite_toolkit/_cdf_tk/commands/build.py
@@ -7,7 +7,7 @@ import shutil
 import sys
 import traceback
 from collections import ChainMap, defaultdict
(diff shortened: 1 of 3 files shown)


haakonvt (approved):
LGTM

PR #564, opened 2024-05-26
😬Cleanup loaders

doctrino:
# Description

* Made `build_dir` an init argument of all Loaders, and store it as `resource_build_path` with the appended `folder_name`. This simplifies the initialization of loaders.
* The `find_files` now defaults to the `resource_build_path`. Simplifiying the signauter of `clean_resources` and `deploy_resources`.
* Removed class attribute `_display_name`. Loaders that want to overwrite this just overwrites the `display_name` property.
* Replaced `_is_equal_custom` with a `are_equal` method with a default implementation, that is overwritten for subclasses that needs special compaison. 
* Reordered method in `ResourceLoaders` based on whether the subclass must or can overwrite the methods. To make it easier to understand how to write a resource loader.

## Checklist

- [ ] Tests added/updated.
- [ ] Run Demo Job Locally.
- [ ] Documentation updated.
- [ ] Version bumped.

### DIFF
--- a/cognite_toolkit/_cdf_tk/commands/clean.py
+++ b/cognite_toolkit/_cdf_tk/commands/clean.py
@@ -1,7 +1,6 @@
 import traceback
 from graphlib import TopologicalSorter
 from pathlib import Path
-from typing import cast
 
 import typer
 from cognite.client.data_classes._base import T_CogniteResourceList
@@ -10,7 +9,7 @@ from cognite.client.utils.useful_types import SequenceNotStr
 from rich import print
 from rich.panel import Panel
 
-from cognite_toolkit._cdf_tk.commands._base import ToolkitCommand
+from cognite_toolkit._cdf_tk.commands._base import LoaderCommand
 from cognite_toolkit._cdf_tk.exceptions import (
     ToolkitCleanResourceError,
     ToolkitNotADirectoryError,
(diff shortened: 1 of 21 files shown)


cognite_toolkit/_cdf_tk/commands/clean.py
```diff
@@ -309,13 +306,12 @@ def execute(
...
             if not issubclass(loader_cls, ResourceLoader):
                 continue
-            loader = loader_cls.create_loader(ToolGlobals)
+            loader = loader_cls.create_loader(ToolGlobals, build_dir)
             if type(loader) is DataSetsLoader:
                 self.warn(ToolkitNotSupportedWarning(feature="Dataset clean."))
                 continue
             result = self.clean_resources(
                 loader,
-                build_path / loader_cls.folder_name,
```
doctrino, line 318: This is an example for a simplified call to `clean_resources` as the loadrs already knows the build path it is not necessary any more

cognite_toolkit/_cdf_tk/commands/deploy.py
```diff
@@ -192,32 +186,29 @@ def execute(
...
         self,
         loader: ResourceLoader,
-        path: Path,
         ToolGlobals: CDFToolConfig,
         dry_run: bool = False,
         has_done_drop: bool = False,
         has_dropped_data: bool = False,
         verbose: bool = False,
     ) -> ResourceDeployResult | None:
-        loader.build_path = path
```
doctrino, line 219: This looks strange, you are setting an attribute on an input argument (mutating an input argument) with another input argument. 

This is one of the motivation for removing the `path` (which is the build path of the resource`, which is set in the init of the loader.

cognite_toolkit/_cdf_tk/commands/clean.py
```diff
@@ -275,25 +274,23 @@ def _remove_duplicates(
 
 class CleanCommand(CleanBaseCommand):
     def execute(
-        self, ctx: typer.Context, build_dir: str, build_env_name: str, dry_run: bool, include: list[str]
+        self, ctx: typer.Context, build_dir_raw: str, build_env_name: str, dry_run: bool, include: list[str]
```
ronpal, line 277: Hva menes med "raw"? At det er en str?

doctrino: yes, that is correct. With our current configuration of mypy it is not allowed to reuse a variable with another type.

PR #565, opened 2024-05-26
[CDF-21572, prep CDF-21544] 🫧 Simplify auth loader

doctrino:
# Description

This is a clean up of the special case used for the `AuthLoader` in the clean and deploy command, which is also a prep for SecurityCategories. 

* Rename `AuthLoader` to `GroupLoader`. Better to match the resource name, and with SecurityCategories coming in the same folder `AuthLoader` will be confusing.
* Used subclassing of `GroupLoader` do deal with the two cases `GroupAllScopedLoader` and `GroupResourceScopedLoader`. This is a more elegant way of dealing with the dependencies between the two types of loaders, which leads the clean and deploy operation to no longer have a special cause for the `GroupLoader`. 
* Avoid including loaders for resources which are not present due to the Topological sort. This is better as you do not get empty rows in the Deploy/Clean table. Switched the wording in the warning `ToolkitDependenciesIncludedWarning` as a result.

## Checklist

- [ ] Tests added/updated.
- [ ] Run Demo Job Locally.
- [ ] Documentation updated.
- [ ] Version bumped.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,7 +15,7 @@ repos:
       - id: ruff-format
         args:
           - --line-length=120
-    rev: v0.4.4
+    rev: v0.4.5
 
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v4.6.0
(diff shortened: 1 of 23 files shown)


ronpal:
I love this improvement. But should I think about GroupLoader as an abstract class/baseclass that you don't use directly? If so, should it either be marked or named as such?

ronpal (approved):
Zen

PR #579, opened 2024-05-29
Feature flag to replace normal init

ronpal:
# Description

Taking a shot at a simple feature flag that enables an alternate init command

## Checklist

- [ ] Tests added/updated.
- [ ] Run Demo Job Locally.
- [ ] Documentation updated.
- [ ] Version bumped.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,16 @@ It supports three different modes of operation:
    bundled with templates useful for getting started with Cognite Data Fusion, as well as for specific use cases
    delivered by Cognite or its partners. You can also create your own templates and share them.
 
+## Usage
+
+Install the Toolkit by running:
+
+```bash
+pip install cognite-toolkit
+```
+
+Then run `cdf-tk --help` to get started with the interactive command-line tool.
+
 ## For more information
 
 More details about the tool can be found at
--- a/cognite_toolkit/_cdf.py
+++ b/cognite_toolkit/_cdf.py
@@ -31,6 +31,7 @@ from cognite_toolkit._cdf_tk.load import (
     NodeLoader,
     TransformationLoader,
 )
+from cognite_toolkit._cdf_tk.prototypes import featureflag
 from cognite_toolkit._cdf_tk.templates import (
     COGNITE_MODULES,
 )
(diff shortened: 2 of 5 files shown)


README.md
ronpal, line 1: (Unrelated really)

cognite_toolkit/_cdf_tk/prototypes/featureflag.py
```diff
@@ -0,0 +1,22 @@
+import os
+from functools import lru_cache
+
+import dotenv
+
+
+@lru_cache(maxsize=128)
```
ronpal, line 7: Have no idea if neccessary or the right maxsixe for a boolean, but since i read from dotenv it seems smart to cache?

cognite_toolkit/_cdf_tk/prototypes/featureflag.py
```diff
@@ -0,0 +1,22 @@
...
+
+    Args:
+        flag (str): The feature flag to check.
+
+    Returns:
+        bool: True if the feature flag is enabled, False otherwise.
+    """
+    dotenv.load_dotenv()
+    if os.environ.get(flag, "false").lower() == "true":
+        print(f"Feature flag {flag} is enabled.")
```
ronpal, line 20: something that eventually could be logged to Mixpanel, Sentry or somewhere else. Unless we replace it with Unleash.

cognite_toolkit/_cdf_tk/prototypes/interactive_init.py
ronpal, line 1: Disregard the command parameters for now, they are the same as the original init for the sake of testing the feature flag.

cognite_toolkit/_cdf.py
```diff
@@ -453,7 +458,7 @@ def auth_verify(
         raise ToolkitValidationError("Failure to verify access rights.")
 
 
-@_app.command("init")
```
doctrino, line 456: Your solution works, an alternative would be 
```python
if featureflag.enabled("FF_INTERACTIVE_INIT"):
    from cognite_toolkit._cdf_tk.prototypes.interactive_init import InteractiveInit

    _app.command("init")(InteractiveInit().interactive)
else:
    @_app.command("init")
    def main_init(...)
```

doctrino: Not sure if there is a value to having the old `init` available on `cdf-tk _init`.

cognite_toolkit/_cdf_tk/prototypes/featureflag.py
doctrino, line 20: Lets use print from rich

cognite_toolkit/_cdf_tk/prototypes/featureflag.py
```diff
@@ -0,0 +1,22 @@
...
+    """
+    Check if a feature flag is enabled.
+
+    Args:
+        flag (str): The feature flag to check.
+
+    Returns:
+        bool: True if the feature flag is enabled, False otherwise.
+    """
+    dotenv.load_dotenv()
```
doctrino, line 18: I don't think this is necessary, in `_cdf.py` in line `187-192` we do load the environment file. This will make the `lru` cache obsolete.

doctrino (approved):
Good introduction, a few suggestions.

PR #666, opened 2024-06-18
[CDF-21817] 😈Check Toolkit Group Correctly setup

doctrino:
# Description

image

## Checklist

- [ ] Tests added/updated.
- [ ] Run Demo Job Locally.
- [ ] Documentation updated.
- [ ] Version bumped.

### DIFF
--- a/cognite_toolkit/_cdf_tk/utils.py
+++ b/cognite_toolkit/_cdf_tk/utils.py
@@ -38,6 +38,7 @@ from cognite.client.credentials import CredentialProvider, OAuthClientCredential
 from cognite.client.data_classes import CreatedSession
 from cognite.client.data_classes.capabilities import Capability, SecurityCategoriesAcl
 from cognite.client.data_classes.data_modeling import View, ViewId
+from cognite.client.data_classes.iam import TokenInspection
 from cognite.client.exceptions import CogniteAPIError, CogniteAuthError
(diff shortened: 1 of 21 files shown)


cognite_toolkit/_cdf_tk/utils.py
```diff
@@ -498,19 +505,6 @@ def initialize_from_auth_variables(self, auth: AuthVariables) -> None:
         )
         self._update_environment_variables()
 
-    def reinitialize_client(self) -> None:
```
doctrino, line 501: Not used so I deleted it

cognite_toolkit/_cdf_tk/utils.py
```diff
@@ -748,37 +719,6 @@ def verify_extraction_pipeline(self, external_id: str, skip_validation: bool = F
             )
         return -1
 
-    def verify_spaces(self, space: str | list[str]) -> list[str]:
```
doctrino, line 751: Removed this one. The reason is that verify_data_set/secutiry_categories etc are for replacing an externalId with an internal ID. This is not necessary for spaces. Thus this check makes cdf-tk deploy too strict as it then makes, for example, containers requiring READ access scoped to all.

Instead, in the deploy method there is a check on capabilities that is correctly scoped, such that for deploying a container we only need read and write access scoped to a specific dataset.

PR #707, opened 2024-06-26
[CDF-21918] 🖐Allow type test and qa

doctrino:
# Description

Upgraded `cognite-SDK` to `7.52` which introduced `immutable` as properties for views and containers. Thus the need to update test data.

## Checklist

- [ ] Tests added/updated.
- [ ] Run Demo Job Locally.
- [ ] Documentation updated.
- [ ] Version bumped.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,7 +14,7 @@ include = ["cognite_toolkit/*", "cognite_toolkit/**/*"]
 [tool.poetry.dependencies]
 python = "^3.9"
 python-dotenv = "^1.0.0"
-cognite-sdk = {version = "^7.51.0", extras = ["pandas"]}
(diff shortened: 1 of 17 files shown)


pyproject.toml
```diff
@@ -24,6 +24,7 @@ cognite-logger = "^0.6"
 questionary = "^2.0.1"
 mixpanel = "^4.10.1"
 packaging = "^24.1"
+typing-extensions = "^4.0"
```
doctrino, line 27: This is required by the cognite-sdk, so does not put any extra dependencies for us.

ronpal (approved):
As mentioned on Slack, I think we should releaste this but also revisit the use of environment.type based on what kind of behaviour we want to encourage.

PR #807, opened 2024-08-01
Gtw/add_devcontainer

gregertw:
# Description

Added .devcontainer config to steamline the dev experience for IDEs that support dev containers.
Also, the config serves as a reference for how to set up alternative environments.

## Checklist

- [ ] Tests added/updated.
- [ ] Run Demo Job Locally.
- [x] Documentation updated.
- [ ] Version bumped.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,18 @@
+FROM python:3.12.4-slim
+
+RUN apt-get update && apt-get install -y git gcc python3-dev procps pipx
+
+# Create the user
+RUN groupadd --gid 1000 vscode
+RUN useradd --uid 1000 --gid 1000 -m vscode --shell /bin/bash
+
+ENV PATH=$PATH:/home/vscode/.local/bin
+ENV SENTRY_ENABLED=false
+
+# Install poetry as the vscode ser
+RUN su -l vscode -c "pipx ensurepath && pipx install poetry==1.8.3"
+
+USER vscode
+# Set VSCode as the core editor for git
+RUN git config --global core.editor "code --wait"
+RUN git config --global --add safe.directory /workspaces/toolkit
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.12.4-slim
+FROM python:3.11.9-slim
 
 RUN apt-get update && apt-get install -y git gcc python3-dev procps pipx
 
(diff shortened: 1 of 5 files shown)


.devcontainer/Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM python:3.12.4-slim
```
doctrino, line 1: Even though we support 3.12, it is recommended to develop on 3.11. 
```suggestion
FROM python:3.11.9-slim
```

There is a few reason for this. 
1. 3.12 allows syntax that is not allowed in earlier versions, i.e., quotes in f-strings.
2. When testing functions, 3.11 is the most recent versions that is allowed.

CONTRIBUTING.md
```diff
@@ -84,6 +84,20 @@ See [tests](tests/README.md) for more information on how to run tests.
...
 
+In order to develop `cdf-tk` you need to set up a development environment. You need a working python
+installation and a virtual environment. We recommend using `poetry` to set up the environment as this is
+the package tool that the toolkit repo uses also to create the installable python package.
+
+When developing, you should use `cdf-tk-dev.py` to run the toolkit. This script will set the environment and paths
+correctly (to avoid running the installed cdf-tk package) and also set the `SENTRY_ENABLED` environment
+variable to `false` to avoid sending errors to Sentry.
+In .vscode/launch.json you will see a number of examples of debugging configurations that you can use to debug.
+If you use VSCode or another IDE supporting devcontainers, the very easiest way to set up the environment is to
```
doctrino, line 95:
```suggestion
If you use VSCode or another IDE supporting devcontainers, the easiest way to set up the environment is to
```

doctrino (approved):
I am approving, but I am uncertain about the value for a devcontainer in this project.

As I understand devcontainers, they are useful for complex dev setups, with external dependencies such as databases, which we do not have in this project. In addition, since cdf-tk should run on different OSes, it is useful that we also develop it on Windows and Mac, to hopefully pick up some of the same issues our users will run into.

.devcontainer/Dockerfile
gregertw: I think this is a very good reason for why a devcontainer is a good idea :) Although you don't need it, it is a very good reference for the environment that PRs should support.

gregertw:
The main values are the following:

1. We are documenting in code a reference environment to quickly be able to set up a dev environment the way it should be done. Good for on-boarding of new devs and to create a shared agreement.
2. Lower the barriers to creating PRs as you don't have the challenges of different linting, different OS, different installed packages.
3. Creates a reference environment to be used if there are discrepancies between what works in your local env, CI/CD, and other devs' environments
4. Is a recommended "feature" for Cognite product repos as a common way of making it easier to move across teams and contribute

PR #865, opened 2024-08-27
[CDF-22349] 🦮 Introduce Config

doctrino:
# Description

## Checklist

- [ ] Tests added/updated.
- [ ] Run Demo Job Locally.
- [ ] Documentation updated.
- [ ] Version bumped.


cdf.toml
```diff
@@ -0,0 +1,39 @@
...
+featureflags = [
+    "import",
+]
+
+[modules]
+# This is the version of the modules. It should not be changed manually.
+# It will be updated by the 'cdf module upgrade' command.
+version = "0.2.18"
+
+[packages]
```
doctrino, line 16: An alternative is to hardcode the packages into the CLI, however, I do not like that option.

cdf.toml
```diff
@@ -0,0 +1,39 @@
+[cdf]
```
doctrino, line 1:
```suggestion
[cdf]
project = "my_project"
```

cdf.toml
doctrino, line 16:
```suggestion
[modules.packages]
```

cdf.toml
```diff
@@ -0,0 +1,32 @@
...
+project_dir = "cognite_toolkit"
```
ronpal, line 2: Should we consider another word than "project_dir"? I know we've used that until now, but project can be related to both a process and a CDF instance, none of which we are talking about here. How about work_dir or similar?

cdf.toml
```diff
@@ -0,0 +1,32 @@
...
+    "cdf_auth_readwrite_all",
+    "cdf_apm_base",
+    "cdf_infield_common",
+    "cdf_infield_location",
+]
+example_pump = [
+    "example_pump_asset_hierarchy",
+    "cdf_asset_source_model",
+    "example_pump_data_model",
+]
```
ronpal, line 32: I think this is the way forward. Looking at the interactive init and the [manifest files](https://github.com/cognitedata/toolkit/blob/main/cognite_toolkit/_cdf_tk/prototypes/_packages/quickstart/manifest.yaml), the only thing missing (?) is a more explanatory text for each module to print in the interaction. But that can be solved in other ways. How about this? 
```suggestion
data_model_examples = {
    title: "Data model examples to learn from", 
    cdf_auth_readwrite_all: "Access groups that shouldn't be here",
    cdf_oid_example_data: "Actual sample data from the Valhall platform",
    cdf_apm_simple_data_model: "This is out of fashion anyway"
},
cdf_demo_infield = {
    title: "End-to-end setup of InField demo",
    cdf_auth_readwrite_all: "redundant groups",
    cdf_oid_example_data: "Actual sample data from the Valhall asset",
    cdf_apm_base: ""
    cdf_infield_common: "",
    cdf_infield_location: "",
}

...```

cdf.toml
doctrino: I agree, but I think we do not yet have a good alternative.

ronpal (approved):

PR #920, opened 2024-09-09
[CDF-22492] 👷‍♂️ Fix creation config.dev.yaml

doctrino:
# Description

Testing `cdf modules init` found multiple issues

* Deleted `default.environments.yaml` as it is no longer used.
* Added `module.toml` to each builtin module as for example selecting Examples yielded nothing.
* Renamed `organization_dir` to `default_organization_dir` after finding a bug in `build| as a result of misuse`
* Remove unused parameter `arg_package`
* Replaced `InitCofngiYAML.load_selected_defaults` with `load_defaults` as it did not include all `defautl.config.yaml` from the source folder.
* Merged `SelectableModule` into `ModuleLocation` as we need access to the modules relative path, and no need to have two data classes for the same concept.
* Made `ModuleToml` immutable such that it can be a property of the immutable `ModuleLocation`.
* Changed data structure `Packages` from a `Sequence` to a `Mapping` to better match usage.
* In `ModulesCmd.init` removed `_copy_all` and instead reuse the `_create` method as `copy_all` did not create `config.[env].yam` files.
* Some refactoring of `ModulesCmd.init` to reuse for `--all` flag and interactive, and make the flow clearer.

# Checklist

- [ ] Tests added/updated.
- [ ] Run Demo Job Locally.
- [ ] Documentation updated.
- [ ] Version bumped.


cognite_toolkit/_cdf_tk/apps/_modules_app.py
```diff
@@ -52,7 +45,7 @@ def init(
             typer.Option(
                 "--clean",
                 "-a",
-                help="Clean target 'modules' directory if it exists",
+                help="Clean target 'organization' directory if it exists",
```
ronpal, line 48: The code deletes orgdir/modules, not orgdir

```suggestion
                help="Clean target directory if it exists",
```

ronpal (changes requested):
Nice. Small inaccuracy in a prompt, otherwise good

ronpal (approved):

PR #1158, opened 2024-11-01
[CDF-23090, CDF-23107] 🦍Create RAW Database before Table

doctrino:
# Description

Bug is that 
```yaml
dbName: db_complete_org
tableName: asset
```
causes to fail on an empty project as `database` is not created, and fails when trying to create the table. In addition, discovered that there was a bug in the mocking leading ApprovalClient not to pick up RAW Database or Table, which probably explains how this bug got through in the first place.

Realized that the easiest solution was to fix the duplicated bug as well, as fixing the above bug, leads to sourcesystem as well as a few other builtin modules to complain about duplicates. 

## Checklist

- [ ] Tests added/updated.
- [ ] Run Demo Job Locally.
- [ ] Documentation updated.
- [ ] Version bumped.


cognite_toolkit/_cdf_tk/builders/_raw.py
```diff
@@ -23,14 +24,22 @@ def build(
...
                 except KeyError:
+                    seen_databases.add(tuple(item.items()))
                     entry_by_loader[RawDatabaseLoader].append(item)
                 else:
                     entry_by_loader[RawTableLoader].append(item)
+                    db_item = RawDatabaseLoader.dump_id(RawDatabase(table_id.db_name))
+                    hashable_db_item = tuple(db_item.items())
+                    if hashable_db_item not in seen_databases:
+                        seen_databases.add(hashable_db_item)
+                        entry_by_loader[RawDatabaseLoader].append(db_item)
```
doctrino, line 41: This fixes the original bug

tests/data/complete_org/modules/my_example_module/raw/staging.Database.yaml
```diff
@@ -1 +0,0 @@
-dbName: db_complete_org
```
doctrino, line 1: Test data no longer necessary, it is covered by the Table

cognite_toolkit/_cdf_tk/commands/build.py
```diff
@@ -544,27 +545,28 @@ def check_built_resource(
...
                         )
                 else:
-                    warning_list.append(MissingRequiredIdentifierWarning(source_path, element_no, tuple(), error.args))
+                    warning_list.append(MissingRequiredIdentifierWarning(source.path, element_no, tuple(), error.args))
 
             if identifier:
                 identifier_kind_pairs.append((identifier, item_loader.kind))
                 if first_seen := self._ids_by_resource_type[item_loader].get(identifier):
-                    warning_list.append(DuplicatedItemWarning(source_path, identifier, first_seen))
+                    if first_seen.hash != source.hash:
```
doctrino, line 556: This is the fix that stop complaining about identical resources.

ronpal (approved):

PR #1313, opened 2024-12-20
[CDF-23593] 😃 Refactor dump method

doctrino:
# Description

Changes the `ResourceLoaders`

Removed the following method
```python
    def are_equal(
        self,
        local: T_WriteClass,
        cdf_resource: T_WritableCogniteResource,
        return_dumped: bool = False,
        ToolGlobals: CDFToolConfig | None = None,
    ) -> bool | tuple[bool, dict[str, Any], dict[str, Any]]:
    ...
```

Changed the signature of the following three methods
```python
    def load_resource_file(
        self, filepath: Path, ToolGlobals: CDFToolConfig, is_dry_run: bool = False
    ) -> T_WriteClass | T_CogniteResourceList: ...

    def load_resource(
        self, resource: dict[str, Any] | list[dict[str, Any]], is_dry_run: bool = False, filepath: Path | None = None
    ) -> T_WriteClass | T_CogniteResourceList: ...

    def dump_resource(self, resource: T_WritableCogniteResource, local: T_WriteClass) -> dict[str, Any]: ...
```
To this:
```python
    def load_resource_file(
        self, filepath: Path, environment_variables: dict[str, str | None] | None = None
    ) -> list[dict[str, Any]]: ...

    def load_resource(self, resource: dict[str, Any], is_dry_run: bool = False) -> T_WriteClass: ...

    def dump_resource(self, resource: T_WritableCogniteResource, local: dict[str, Any]) -> dict[str, Any]: ...
```

Notice that this signature has fewer arguments than the original. It simplifies the implementations, in addition, by returning the `list[dict[str, Any]]` in the `load_resource_file` the methods are more flexible which means that they can be used in the `cdf modules pull` command.

Finally, realizes that loading from file, and figuring out what to create, update, and remain unchange, had to be done in one step, as now we dump the CDF resources into the same format as the local file.

Introduced a `ResourceWorker` to do this. note I am not happy with the name, I think it should ideally be changed to `ResourceLoader` and rename the `ResourceLoaders` to something like `StandardizedResource` as this does a lot more than just loading individual files, dumping, CRUD, iteration, authorization as well.

```python
class ResourceWorker
    def __init__(self,  loader: ResourceLoader):
        self.loader = loader

    def load_files(
        self, sort: bool = True, directory: Path | None = None, read_modules: list[ReadModule] | None = None
    ) -> list[Path]: ...

    def load_resources(
        self,
        filepaths: list[Path],
        return_existing: bool = False,
        environment_variables: dict[str, str | None] | None = None,
        is_dry_run: bool = False,
        verbose: bool = False,
    ) -> (
        tuple[T_CogniteResourceList, T_CogniteResourceList, T_CogniteResourceList, list[T_ID]]
        | tuple[T_WritableCogniteResourceList, list[T_ID]]
    ): ...
```

## Checklist

- [ ] Tests added/updated.
- [ ] Run Demo Job Locally.
- [ ] Documentation updated.
- [ ] Version bumped.


CHANGELOG.cdf-tk.md
```diff
@@ -25,6 +25,7 @@ Changes are grouped as follows:
 ### Fixed
 
 - No more warning about missing `.env` file when running in `Google Cloud Build`.
+- When deploying a `Sequence` resource, Cognite Toolkit now replaces `dataSetExternalId` with `dataSetId`.
```
doctrino, line 28: Found this bug in the refactoring.

cognite_toolkit/_cdf_tk/client/api/lookup.py
```diff
@@ -82,7 +83,7 @@ def external_id(
         id: int | Sequence[int],
     ) -> str | list[str]:
         ids = [id] if isinstance(id, int) else id
-        missing = [id_ for id_ in ids if id not in self._reverse_cache]
+        missing = [id_ for id_ in ids if id_ not in self._reverse_cache]
```
doctrino, line 86: Another bug introduced in last refactoring.

ronpal (approved):
> Introduced a ResourceWorker to do this. note I am not happy with the name, I think it should ideally be changed to ResourceLoader and rename the ResourceLoaders to something like StandardizedResource as this does a lot more than just loading individual files, dumping, CRUD, iteration, authorization as well.

Agree that we should rename, let's revisit later. Apart from that, great progress 🚀

PR #1371, opened 2025-01-15
Create separate oneshot session per transformation

dmivankov:


cognite_toolkit/_cdf_tk/commands/run.py
```diff
@@ -650,8 +646,12 @@ def run_transformation(self, client: ToolkitClient, external_ids: str | list[str
         if transformations is None or len(transformations) == 0:
             print(f"[bold red]ERROR:[/] Could not find transformation with external_id {external_ids}")
             return False
-        nonce = NonceCredentials(session_id=session.id, nonce=session.nonce, cdf_project_name=client.config.project)
         for transformation in transformations:
+            session = client.iam.sessions.create(session_type="ONESHOT_TOKEN_EXCHANGE")
+            if session is None:
+                print("[bold red]ERROR:[/] Could not get a oneshot session.")
+                return False
+            nonce = NonceCredentials(session_id=session.id, nonce=session.nonce, cdf_project_name=client.config.project)
```
doctrino, line 654: We do not expose more than running one transformation at a time, so the user would never run into this issue.

doctrino (approved):

PR #1472, opened 2025-02-20
[CDF-24158] Fixing pull for transformations

ronpal:
# Description

Fixed an error in pulling transformation sql.

## Changelog

- [x] Patch
- [ ] Minor
- [ ] Skip

### Fixed

- Pulling transformations into could fail because the internal build step added .Query. as suffix to the sql file. This is now fixed.

## templates

No changes.


cognite_toolkit/_cdf_tk/builders/_transformation.py
```diff
@@ -30,8 +34,10 @@ def build(
         for source_file in source_files:
             loaded = source_file.loaded
             if loaded is None:
-                # Not a YAML file
-                continue
+                if accept_invalid_files and source_file.source.path.suffix == ".yaml":
+                    loaded = {}
```
ronpal, line 38: Ignoring that some files are empty (yaml.safe_load returns None)

cognite_toolkit/_cdf_tk/commands/build.py
```diff
@@ -586,6 +598,9 @@ def check_built_resource(
     ) -> tuple[WarningList[FileReadWarning], list[tuple[Hashable, str]]]:
         warning_list = WarningList[FileReadWarning]()
 
+        if not parsed:
+            parsed = {}
```
ronpal, line 602: Unsure if this will have side effects yet

cognite_toolkit/_cdf_tk/commands/pull.py
```diff
@@ -599,12 +603,20 @@ def _get_local_resource_dict_by_id(
...
         environment_variables: dict[str, str | None],
+        allow_invalid_files: bool,
     ) -> dict[T_ID, dict[str, Any]]:
         unique_destinations = {r.destination for r in resources if r.destination}
         local_resource_by_id: dict[T_ID, dict[str, Any]] = {}
         local_resource_ids = set(resources.identifiers)
         for destination in unique_destinations:
-            resource_list = loader.load_resource_file(destination, environment_variables)
+            try:
+                resource_list = loader.load_resource_file(destination, environment_variables)
```
ronpal, line 613: Invalid files will fail again even if I managed to force through an invalid build (this is when the pull command tries to load)

cognite_toolkit/_cdf_tk/builders/_base.py
```diff
@@ -60,7 +60,11 @@ def build_dir(self) -> Path:
 
     @abstractmethod
     def build(
-        self, source_files: list[BuildSourceFile], module: ModuleLocation, console: Callable[[str], None] | None = None
+        self,
+        source_files: list[BuildSourceFile],
+        module: ModuleLocation,
+        console: Callable[[str], None] | None = None,
+        accept_invalid_files: bool = False,
```
doctrino, line 67: Any idea is to instead call it something like this
```suggestion
        validation: Literal["identifier", "full"] = "full",
```

ronpal: Much better! Implemented.

tests/test_unit/test_cdf_tk/test_commands/test_pull.py
```diff
@@ -519,3 +522,16 @@ def test_to_write_content(
...
+        cmd = BuildCommand(silent=True, skip_tracking=True, print_warning=False)
+
+        res = cmd.build_modules(
+            modules=input_modules,
+            build_dir=build_tmp_path,
+            on_error="raise",
+            validation="identifier",
+            variables=BuildVariables([]),
+        )
+        assert res is not None
```
doctrino, line 537: This is empty. There is no test data.

cognite_toolkit/_cdf_tk/utils/auth.py
```diff
@@ -21,7 +21,7 @@
 from cognite_toolkit._cdf_tk.utils import humanize_collection
 from cognite_toolkit._version import __version__
 
-Provider: TypeAlias = Literal["entra_id", "auth0", "other"]
+Provider: TypeAlias = Literal["entra_id", "auth0", "other", "cdf"]
```
doctrino, line 24: Bad merge? Why was `cdf` added back?

cognite_toolkit/_cdf_tk/loaders/_base_loaders.py
```diff
@@ -328,6 +328,10 @@ def load_resource(self, resource: dict[str, Any], is_dry_run: bool = False) -> T
         """Loads the resource from a dictionary. Can be overwritten in subclasses."""
         return self.resource_write_cls._load(resource)
 
+    def load_minimum_resource(self, resource: dict[str, Any]) -> T_ID:
```
doctrino, line 331: I don't think this is necessary. The `self.get_id` supports dict` directly. 
So you can do 
```python
my_id = loader.get_id({"externalId": "some-id"})
```

cognite_toolkit/_cdf_tk/commands/build.py
```diff
@@ -268,11 +274,16 @@ def build_modules(
...
                 try:
-                    built_module_resources = self._build_module_resources(module, build_dir, module_variables, verbose)
+                    built_module_resources = self._build_module_resources(
+                        module, build_dir, module_variables, verbose, validation
+                    )
                 except ToolkitError as err:
                     if on_error == "raise":
                         raise
 
+                    if validation == "identifier":
```
doctrino, line 284: Will not this lead to a silent skip on_error =="continue". The code below logs what the issue was with building the module resources.

cognite_toolkit/_cdf_tk/commands/build.py
```diff
@@ -355,9 +367,7 @@ def _build_module_resources(
                     continue
 
                 file_warnings, identifiers_kind_pairs = self.check_built_resource(
```
doctrino, line 369: I would expect this to take in the validation level

cognite_toolkit/_cdf_tk/builders/_base.py
```diff
@@ -178,8 +187,16 @@ def build(
                 if warning is not None:
                     yield [warning]
                 continue
-            destination_path = self._create_destination_path(source_file.source.path, loader.kind)
 
+            if validation == "identifier":
```
doctrino, line 191: This seems to be duplicated in many builders. What about just making an `xIdentifierBuilder`? and use that if the validation in `identifier`? 

This also seems to be duplicating the `.check_built_resource` that is in the build command. Why do you need to do this here at all?

doctrino (changes requested):
See comments, but two main things
1. There is no placeholder test data.
2. Suggest introducing an IdentifierBuilder to handle building for all resources when the validation is identifying. I think this should not look up the ID, as that is done in the `check_built_resource` in the BuildCommand.

cognite_toolkit/_cdf_tk/builders/_transformation.py
```diff
@@ -19,12 +19,15 @@ class TransformationBuilder(Builder):
...
+        self,
+        source_files: list[BuildSourceFile],
+        module: ModuleLocation,
+        console: Callable[[str], None] | None = None,
     ) -> Iterable[BuildDestinationFile | list[ToolkitWarning]]:
         query_files = {
             source_file.source.path: source_file
             for source_file in source_files
-            if source_file.source.path.suffix == ".sql"
+            if source_file.source.path.name.endswith(".sql") or source_file.source.path.name.endswith(".Query.sql")
```
ronpal, line 30: This is the actual fix. Build added Kind (Query) which is ok for Deploy, but caused the builder not to find it during pull (second build)

CHANGELOG.cdf-tk.md
```diff
@@ -15,6 +15,12 @@ Changes are grouped as follows:
 - `Fixed` for any bug fixes.
 - `Security` in case of vulnerabilities.
 
+## TBD
+
+### Fixed
+
+- Kind in Query files broke pull command. This is now fixed.
+
```
doctrino, line 23: Lets merge this after the new release process is in place. Move this to the description of the PR.

tests/test_unit/test_cdf_tk/test_loaders/test_base_loaders.py
```diff
@@ -325,14 +325,3 @@ def test_unique_kind_by_folder(self):
...
-
-class TestLoaders:
-    def test_unique_display_names(self, env_vars_with_client: EnvironmentVariables):
-        name_by_count = Counter(
-            [loader_cls.create_loader(env_vars_with_client.get_client()).display_name for loader_cls in LOADER_LIST]
-        )
-
-        duplicates = {name: count for name, count in name_by_count.items() if count > 1}
-
-        assert not duplicates, f"Duplicate display names: {duplicates}"
```
doctrino, line 338: Why is this removed?

cognite_toolkit/_cdf_tk/builders/_transformation.py
doctrino: I am confused about this. I don't understand how the second term will ever come into effect. If a file has suffixed `.sql` then `source_file.source.path.name.endswith(".sql")` will be true, but if it does not, the expression `source_file.source.path.name.endswith(".Query.sql")` will never be true.

cognite_toolkit/_cdf_tk/builders/_datamodels.py
```diff
@@ -30,31 +30,28 @@ def build(
...
                     yield [warning]
                 continue
 
+            extra_sources: list[SourceLocation] | None = None
             destination_path = self._create_destination_path(source_file.source.path, loader.kind)
 
-            extra_sources: list[SourceLocation] | None = None
             if loader is GraphQLLoader:
-                # The GraphQL must be copied over instead of added to the DML field as
-                # it is hashed in the deployment step and used to determine if the DML has changed.
```
doctrino, line 47: This is useful information. Why is it removed?

doctrino (commented):
I am very confused by this. I cannot see the test that fixes the issue, I cannot see the code that fixes it. Are the commits missing?

cognite_toolkit/_cdf_tk/builders/_transformation.py
ronpal: You're right. The second is superfluous. It should be enough to replace 

`if source_file.source.path.suffix == ".sql"
`
with 

`if source_file.source.path.name.endswith(".sql")
`

I'll fix it.

tests/test_unit/test_cdf_tk/test_loaders/test_base_loaders.py
ronpal: By mistake. Restored.

cognite_toolkit/_cdf_tk/builders/_datamodels.py
ronpal: By mistake. Restored.

PR #1483, opened 2025-02-26
[CDF-24185] 🚓 Allow CDF auth provider

doctrino:
# Description

## Before
There were no way to set ENV variables to authenticate with CDF as the provider.

## After
You can use CDF as the provider if you set your env file to 
```env
LOGIN_FLOW=client_credentials
PROVIDER=cdf
CDF_CLUSTER=
CDF_PROJECT=
IDP_CLIENT_ID=
IDP_CLIENT_SECRET=
```

We are not exposing this to the end user as this is only in private alpha used by GSS.

## Checklist

- [ ] Prefixed the PR title with the Jira issue number on the form `[CDF-12345]`.
- [ ] Tests added/updated.
- [ ] Run Demo Job Locally.
- [ ] Documentation updated.

[CDF-12345]: https://cognitedata.atlassian.net/browse/CDF-12345?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cognite_toolkit/_cdf_tk/utils/auth.py
+++ b/cognite_toolkit/_cdf_tk/utils/auth.py
@@ -111,7 +111,7 @@ class EnvironmentVariables:
                 "entra_id": "https://login.microsoftonline.com/{IDP_TENANT_ID}/oauth2/v2.0/token",
                 "auth0": "https://<my_auth_url>/oauth/token",
             },
-            required=all_providers(flow="client_credentials", exclude="entra_id"),
+            required=all_providers(flow="client_credentials", exclude={"entra_id", "cdf"}),
             optional=frozenset([("entra_id", "client_credentials")]),
         ),
     )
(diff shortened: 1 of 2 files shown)


ronpal (approved):

finnag (commented):
What do you mean by "We are not exposing this to the end user as this is only in private alpha used by GSS" -- this looks like a public repository?

doctrino:
> t do you mean by "We are not exposing this to the end user as this is only in private alpha used by GSS" -- this looks like a public repository?

When the user call `cdf init` they will interactively get options to select provider. Here, we will only give `entra_id`, `auth0`, and `other`. Not `cdf`, thus not exposing it.

finnag (approved):

PR #1487, opened 2025-02-27
[CDF-24194]🤖 New release process

doctrino:
# Description

After a discussion with the risk-reviewer team during the office hours 26/02, we are recommended to change our release process.

The changes are as follows:
* Changelog is no longer in the code base - only available in GitHub releases. This is to avoid unnecessary merge conflict of the changelog.
* A changelog entry should be part of the PR description. The PR description is now expected to follow the template.
* The release process controls the versions, so the placeholder `0.0.0` is used in the main branch and only replaced just before release.


ronpal (approved):
Will be interesting!

larshagencognite (approved):

PR #1489, opened 2025-02-28
[CDF-24184] 🐕‍🦺Hint view direct relation error

doctrino:
# Description

The API gives the following error message, which is misleading as the referenced view is part of the requst. The issue is reported to DMS team
image.

## Changelog

- [x] Patch
- [ ] Minor
- [ ] Skip

### Improved

- Error message when deploying views in which reverse direct relation is missing target.

## templates

No changes.


doctrino:
Solved in #1490

PR #1490, opened 2025-03-01
[CDF-24184] 🚍 Strongly coupled views.

doctrino:
# Description

Data models with strongly coupled views through implements and reverse direct relations are tricky to deploy as the API returns 400 complaining about views missing which are part of the request.

image

## Changelog

- [x] Patch
- [ ] Minor
- [ ] Skip

### Improved

-  Running `cdf deploy` for highly coupled views, the Toolkit now tries to recover if the API returns 400 complaining about missing views which are part of the model.

## templates

No changes.


cognite_toolkit/_cdf_tk/loaders/_resource_loaders/datamodel_loaders.py
```diff
@@ -601,38 +602,77 @@ def create(self, items: Sequence[ViewApply]) -> ViewList:
             if self._is_auto_retryable(e1):
                 # Fallback to creating one by one if the error is auto-retryable.
                 return self._fallback_create_one_by_one(items, e1)
-            elif self._is_deployment_order(e1, set(self.get_ids(items))):
-                return self._fallback_create_one_by_one(self._topological_sort(items), e1, warn=False)
```
doctrino, line 605: This is handled by the new method, so we no longer need it.

ronpal (approved):

BugGambit:
@erlendvollset what do you think here? Something that should be fixed in the API?

doctrino:
> @erlendvollset what do you think here? Something that should be fixed in the API?

Starbase is already notified, and yes this is definitely something that should be fixed in the API. However, we currently have a customers that is waiting for this fix. So suggest we merge this one, as that will solve the immediate problem.

BugGambit (approved):

PR #1495, opened 2025-03-06
full overhaul of all bootcamp materials for version 3

petreeb:
# Description

Version 3 of the Bootcamp is ready. This is the newest material for the bootcamp that the participants need to get started. There are intentional errors and misconfigurations that the participants are required to fix.

## Changelog

- [x] Patch
- [ ] Minor
- [ ] Skip

### Changed

No changes

### Changed

- Changed Bootcamp module from v2 to v3


ronpal (commented):
@petreeb do you see any good reason why we can't introduce scoped access here? It seems like a good opportunity to get participants into good habits from the start

petreeb:
@ronpal I've adjusted the variables so they are as scoped as possible. What do you think?

larshagencognite (approved):
risk review ok

PR #1496, opened 2025-03-07
Missing internal ID lookup 

doctrino:
# Description

We no longer raise an error when looking up external IDs from internal IDs. Instead we give a warning. The reasoning is that every time we look up an external ID from internal, it is from a resource already deployed to CDF and not a user fault in local configuration. Thus, we give warning instead.

## Changelog

- [x] Patch
- [ ] Minor
- [ ] Skip

### Fixed

- When running `cdf deploy/pull`, the Toolkit no longer raise an `ResourceRetrievalError` if a CDF resource is referencing a non-existing resource. For example, if a group has an `TimeSeriesAcl` scoped to a timeseries that no longer exists. 

## templates

No changes.


dbrattli (approved):
:unicorn:

PR #1500, opened 2025-03-11
[CDF-24233, CDF-24234] 🚶🏻Dump command wrong model issue

doctrino:
# Description

Interactive selection of data models can lead to the selection of a model that is different than you selected. 
### Before
image

## Changelog

- [x] Patch
- [ ] Minor
- [ ] Skip

### Fixed

- When running `cdf dump datamodel` in interfactive mode, the Toolkit now always dumps the selected data model. Earlier versions, depending on the choices the user made, Toolkit could end up not dumping the correct model.

## templates

No changes.

### DIFF
--- a/tests/test_unit/test_cli/test_packages.py
+++ b/tests/test_unit/test_cli/test_packages.py
@@ -1,4 +1,4 @@
-from collections.abc import Callable, Sequence
+from collections.abc import Sequence
 from pathlib import Path
 from typing import Any
 
@@ -12,51 +12,7 @@ from cognite_toolkit._cdf_tk.commands import BuildCommand, ModulesCommand
 from cognite_toolkit._cdf_tk.constants import BUILTIN_MODULES_PATH
 from cognite_toolkit._cdf_tk.data_classes import Package, Packages
 from cognite_toolkit._cdf_tk.tk_warnings import DuplicatedItemWarning, TemplateVariableWarning
-
-
-class MockQuestion:
-    def __init__(self, answer: Any, choices: list[Choice] | None = None) -> None:
-        self.answer = answer
-        self.choices = choices
-
-    def ask(self) -> Any:
-        if isinstance(self.answer, Callable):
-            return self.answer(self.choices)
-        return self.answer
-
-
-class MockQuestionary:
-    def __init__(self, module_target: str, monkeypatch: MonkeyPatch, answers: list[Any]) -> None:
-        self.module_target = module_target
-        self.answers = answers
-        self.monkeypatch = monkeypatch
-
-    def select(self, *_, choices: list[Choice], **__) -> MockQuestion:
-        return MockQuestion(self.answers.pop(0), choices)
-
-    def confirm(self, *_, **__) -> MockQuestion:
-        return MockQuestion(self.answers.pop(0))
-
-    def checkbox(self, *_, choices: list[Choice], **__) -> MockQuestion:
-        return MockQuestion(self.answers.pop(0), choices)
-
-    def text(self, *_, **__) -> MockQuestion:
-        return MockQuestion(self.answers.pop(0))
-
-    def __enter__(self):
-        for method in [self.select, self.confirm, self.checkbox, self.text]:
-            self.monkeypatch.setattr(f"{self.module_target}.questionary.{method.__name__}", method)
-        return self
-
-    def __exit__(self, *args):
-        self.monkeypatch.undo()
-        return False
-
-    @staticmethod
-    def select_all(choices: list[Choice]) -> list[str]:
-        if not choices:
-            return []
-        return [choice.value for choice in choices]
+from tests.test_unit.utils import MockQuestionary
 
 
 def get_packages() -> list[ParameterSet]:
--- a/tests/test_unit/utils.py
+++ b/tests/test_unit/utils.py
@@ -12,7 +12,7 @@ import typing
 from datetime import date, datetime
 from pathlib import Path
 from types import UnionType
-from typing import IO, Any, Literal, Optional, TypeVar, get_args, get_origin
+from typing import IO, Any, Callable, Literal, Optional, TypeVar, get_args, get_origin
 
 from _pytest.monkeypatch import MonkeyPatch
 from cognite.client import CogniteClient
@@ -55,6 +55,7 @@ from cognite.client.data_classes.workflows import (
 )
 from cognite.client.testing import CogniteClientMock
 from cognite.client.utils.useful_types import SequenceNotStr
+from questionary import Choice
 
 from cognite_toolkit._cdf_tk._parameters.get_type_hints import _TypeHints
 from cognite_toolkit._cdf_tk.client.data_classes.location_filters import LocationFilterScene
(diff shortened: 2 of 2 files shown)


cognite_toolkit/_cdf_tk/commands/dump_resource.py
```diff
@@ -90,54 +90,56 @@ def __init__(self, client: ToolkitClient, identifier: DataModelId | None = None)
         self.space_ids: set[str] = set()
 
     def _interactive_select(self) -> DataModelId:
-        include_global = False
-        spaces = self.client.data_modeling.spaces.list(limit=-1, include_global=include_global)
-        selected_space: str = questionary.select(
-            "In which space is your data model located?", [space.space for space in spaces]
-        ).ask()
-
```
doctrino, line 98: This listing of spaces include spaces without a data model (typically instances spaces), thus I changed it to only display the spaces that contain a data model.

tests/test_unit/test_cli/test_packages.py
```diff
@@ -12,51 +12,7 @@
...
-        self.answer = answer
-        self.choices = choices
-
-    def ask(self) -> Any:
-        if isinstance(self.answer, Callable):
-            return self.answer(self.choices)
-        return self.answer
-
-
-class MockQuestionary:
```
doctrino, line 28: Pure move to reuse in different test

cognite_toolkit/_cdf_tk/commands/dump_resource.py
```diff
@@ -90,54 +90,59 @@ def __init__(self, client: ToolkitClient, identifier: DataModelId | None = None)
...
-        if not data_model_ids:
-            raise ToolkitMissingResourceError(f"No data models found in space {selected_space}")
+        available_spaces = sorted({model.space for model in data_model_ids})
+        if not available_spaces:
+            raise ToolkitMissingResourceError("No data models found")
+        if len(available_spaces) == 1:
+            selected_space = available_spaces[0]
+        else:
+            selected_space = questionary.select("In which space is your data model located?", available_spaces).ask()
```
doctrino, line 102: Edge case, but do not ask about space if there is only one

cognite_toolkit/_cdf_tk/commands/dump_resource.py
```diff
@@ -90,54 +90,59 @@ def __init__(self, client: ToolkitClient, identifier: DataModelId | None = None)
...
-            or not questionary.confirm(
-                f"Would you like to select a different version than {selected_data_model.version} of the data model",
-                default=False,
-            ).ask()
-        ):
-            self.data_model = data_models[0]
+        }
+        if len(models_by_version) == 1:
+            self.data_model = retrieved_models[0]
+            return selected_data_model
```
doctrino, line 132: Earlier we asked for version, even if there is only one. Now, unless there are multiple versions of a model, we do not mention version.

tests/test_unit/approval_client/client.py
```diff
@@ -795,6 +796,19 @@ def return_values(*args, **kwargs):
 
             return read_list_cls(existing_resources[resource_cls.__name__], cognite_client=client)
 
+        def return_data_models(
+            ids: DataModelIdentifier | Sequence[DataModelIdentifier], inline_views: bool = False
+        ) -> DataModelList:
```
doctrino, line 801: Needed a more specific retrieve method for data models in the mock client.

ronpal (approved):

cognite_toolkit/_cdf_tk/apps/_dump_app.py
```diff
@@ -79,7 +79,7 @@ def dump_datamodel_cmd(
         """This command will dump the selected data model as yaml to the folder specified, defaults to /tmp."""
         selected_data_model: Union[DataModelId, None] = None
         if data_model_id is not None:
-            if len(data_model_id) <= 2:
+            if len(data_model_id) < 2:
                 raise ToolkitRequiredValueError(
                     "Data model ID must have at least 2 parts: space, external_id, and, optionally, version."
```
rbtcollins, line 84: Either external_id is optional, or the condition is wrong.

cognite_toolkit/_cdf_tk/commands/dump_resource.py
```diff
@@ -90,54 +90,59 @@ def __init__(self, client: ToolkitClient, identifier: DataModelId | None = None)
...
+        )
 
         selected_data_model: DataModelId = questionary.select(
-            "Which data model would you like to dump?", [Choice(f"{model!r}", value=model) for model in data_model_ids]
+            "Which data model would you like to dump?",
+            [
+                Choice(f"{model_id!r}", value=model_id)
+                for model_id in sorted(data_model_ids, key=lambda model: model.as_tuple())
+            ],
         ).ask()
```
rbtcollins, line 113: Above you implicitly select the space when only 1 exists; why do you not implicitly select the model when only one model exists?

cognite_toolkit/_cdf_tk/commands/dump_resource.py
```diff
@@ -90,54 +90,59 @@ def __init__(self, client: ToolkitClient, identifier: DataModelId | None = None)
...
+            # This happens if the data model is removed after the list call above.
+            raise ToolkitMissingResourceError(f"Data model {selected_data_model} not found")
+        if len(retrieved_models) == 1:
+            self.data_model = retrieved_models[0]
+            return selected_data_model
+        models_by_version = {
+            model.version: model
+            for model in retrieved_models
             if (model.space, model.external_id) == (selected_data_model.space, selected_data_model.external_id)
             and model.version is not None
```
rbtcollins, line 128: Is the condition on (space, external_id) needed, given you pass space and external_id into the `retrieve` call?

tests/test_unit/approval_client/client.py
```diff
@@ -820,10 +834,11 @@ def files_retrieve(id: int | None = None, external_id: str | None = None) -> Fil
                 return return_value(external_id=external_id)
 
         def data_model_retrieve(ids, *args, **kwargs):
-            id_list = list(ids) if isinstance(ids, Sequence) else [ids]
+            id_set = set(ids) if isinstance(ids, Sequence) else {ids}
```
rbtcollins, line 837: This idiom is generally an antipattern in Python: it is moving more and more to well formed types and easier reasoning about interfaces.

Not for this PR, but I encourage having e.g. ids : Sequence, and then when calling with one id, just pass in [id]

tests/test_unit/utils.py
```diff
@@ -464,3 +465,48 @@ def _type_checking(cls) -> dict[str, Any]:
             "NumpyFloat64Array": NumpyFloat64Array,
             "NumpyObjArray": NumpyObjArray,
         }
+
+
+class MockQuestion:
```
rbtcollins, line 470: these are fine, but it seems like you could equally use mock() with little downside and less code

cognite_toolkit/_cdf_tk/apps/_dump_app.py
doctrino: This condition should triggers on an incorrect input. It is incorrect if there is less than 2 values. 

For example
```python
# Should raise
if len(["just_space"]) < 2 -> Raise

# Should not raise
if len(["space", "ExternalId"])
```

cognite_toolkit/_cdf_tk/commands/dump_resource.py
doctrino: You are correct.

cognite_toolkit/_cdf_tk/commands/dump_resource.py
doctrino: The method is for dumping a datamodel. Then, I want the user to explicit select one. The reason for first selecting space is to avoid overwhelming the user with all models they have access to. If that is only one, we skip it.

tests/test_unit/approval_client/client.py
doctrino: This comes from the Cognite-SDK as we are mocking it. 

> This idiom is generally an antipattern in Python: it is moving more and more to well formed types and easier reasoning about interfaces.

For my curiosity, is this a general consensus? In the PySDK, we have a lot of beginner users. They typically, get easily confused by having to provide something like `client.data_modeling.data_models.retrieve([("my_space", "MyExternalId")])`. Thus it is an advantage that we can take both sequence and individual argumetns.

ronpal (approved):

tests/test_unit/approval_client/client.py
rbtcollins: It was certainly then general consensus in the larger Python community when I was very active in it a decade ago, and I haven't seen anything come in that changes that. If anything, more precision and type safety, not less.

You can, for instance, have 
```
def  get_one(self, space: SpaceId, model: ExternalId) -> Model:
    models = self.retrieve([(space, model)])
    if len(models) == 1:
        return models[0]
    if not len(models):
        raise KeyError((space, model))
    raise RuntimeError(f"More than one model matched {space}:{model}, {model.id for model in models}")
```

or something.

NumPy has super powerful, but also super confusion polymorphic behaviour in this space, but compensates with very very clear documentation - and it grants super powers, which this trivial situation doesn't really.

cognite_toolkit/_cdf_tk/apps/_dump_app.py
rbtcollins: indeed, sorry!

tests/test_unit/approval_client/client.py
```diff
@@ -795,6 +796,19 @@ def return_values(*args, **kwargs):
 
             return read_list_cls(existing_resources[resource_cls.__name__], cognite_client=client)
 
+        def return_data_models(
+            ids: DataModelIdentifier | Sequence[DataModelIdentifier], inline_views: bool = False
```
rbtcollins, line 800: This is the same antipattern I commented on somewhere else.  I'm not going to block this pr, but I would like to see us not doing this are a regular practice.

rbtcollins:
:unicorn:

PR #1501, opened 2025-03-14
Remove devcontainer

doctrino:
# Description

After discussion with @rbtcollins, decide to remove the container as the default container for the space is better, making this one just noise.  

## Changelog

- [ ] Patch
- [ ] Minor
- [x] Skip


ronpal (approved):
:zen:

simenl (approved):
risk review ok

PR #1502, opened 2025-03-18
[CDF-24288] 🪸Dump resource issue

doctrino:
# Description

All loaders (standardized interface to work with CDF resources) have a `dump_resource` method. In some of the loaders, this method assumes that the local representation of the resource has all required fields. This is wrong as, for example, the `cdf modules pull` only requires the identifier fields to be present. 

In the first commit, created a test that tries do dump Response/Read format of a resource together with local dict with only the identifier keys. This triggers all the implementations of the `dump_resource` were this wrong assumption is made.

## Changelog

- [x] Patch
- [ ] Minor
- [ ] Skip

### Fixed

- When using `cdf modues pull`, if a local resource is missing required parameters, the Toolkit no longer raises a `AttributeError` or `KeyError`. Instead, the resource is fetched from CDF and written to the local configuration YAML.

## templates

No changes.

### DIFF
--- a/tests/test_unit/test_cdf_tk/test_loaders/test_base_loaders.py
+++ b/tests/test_unit/test_cdf_tk/test_loaders/test_base_loaders.py
@@ -485,6 +485,29 @@ class TestResourceLoaders:
 
         assert expected_strings.issubset(sensitive_strings), f"Expected {expected_strings} but got {sensitive_strings}"
 
+    @pytest.mark.parametrize(
+        "loader_cls",
+        [
+            loader_cls
+            for loader_cls in RESOURCE_LOADER_LIST
+            if loader_cls != {HostedExtractorSourceLoader, HostedExtractorDestinationLoader}
+        ],
+    )
+    def test_dump_resource_with_local_id(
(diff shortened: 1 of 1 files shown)


ronpal (dismissed):
🚀

tests/test_unit/approval_client/client.py
```diff
@@ -124,7 +127,10 @@ def external_id(
         try:
             return self._reverse_cache[id] if isinstance(id, int) else [self._reverse_cache[i] for i in id]
         except KeyError:
-            raise RuntimeError(f"{type(self).__name__} does not support reverse lookup before lookup")
+            if self._allow_reverse_lookup:
+                return "".join(random.choices(string.ascii_letters + string.digits, k=10))
```
larshagencognite, line 131: The name `_allow_reverse_lookup` was confusing, I read it as `__generate_random_external_id_on_lookup_failure`

doctrino: Thanks. I think there is some renaming needed in the client in the code base as well, but will save that for another PR. Did you suggestion for now.

larshagencognite (approved):
risk review ok

ronpal (approved):

PR #1503, opened 2025-03-19
🎾fix: wrong default in dumping containers

doctrino:
# Description

I did some experimenting on a case, and ended up having to delete and recreate the test containers. This triggered the following bug, which is currently causing the tests on the main branch to fail.

**Risk Reviewer** As this is currently blocking all other PRs it would be nice to get some speed on this one. 

## Changelog

- [x] Patch
- [ ] Minor
- [ ] Skip

### Fixed

- A container with the `nullable` property not set will no longer always be considered changed when running `cdf deploy`.

## templates

No changes.


ronpal (approved):

finnag (approved):
:unicorn:

PR #1505, opened 2025-03-19
[CDF-24336] warn if requirements.txt in streamlit apps are missing useful dependencies

ronpal:
# Description

When Streamlit apps are managed by the Toolkit, the `requirements.txt` content is used by Fusion to load the correct dependencies. A user suggested that we issue a warning if the default dependencies (Pyodide and Cognite SDK) were missing. 

## Changelog

- [x] Patch
- [ ] Minor
- [ ] Skip

### Fixed

- If default dependencies are missing from a Streamlit app's requirements.txt file, the Toolkit issues a warning.

## templates

No changes.


doctrino (approved):
Good stuff, the failing test is caused by a PR I did yesterday. I needed to update the test data  such that it is not producing warnings. Merging in main and run the tests twice and I think you should be good :)

bdemirel (approved):
LGTM, I will request another risk review since I'm onboarding.

cemreyavuz (approved):
:unicorn:

PR #1506, opened 2025-03-19
[CDF-24358] 🥡 cdf dump datamodel --include-global flag

doctrino:
# Description

Put behind an alpha flag, planned for the `0.5` release. 

## Changelog

- [x] Patch
- [ ] Minor
- [ ] Skip

### Added

- [alpha feature] The alpha flag `dump-dm-global` changes the behavior of cdf dump datamodel to skip global resources. The motivation for this change is the following observed user flow. First, a data model is developed outside Toolkit (neat/CDF UI). Then, this command is used to create the configuration that are checked into version control history. In this case, global resources should not be part of the configuration. For example, a CogniteCore extension model should not include the Cognite containers and views in the local configurations as these are controlled by Cognite.

## templates

No changes.


cognite_toolkit/_cdf_tk/apps/_dump_app.py
```diff
@@ -23,7 +23,10 @@ class DumpApp(typer.Typer):
     def __init__(self, *args: Any, **kwargs: Any) -> None:
         super().__init__(*args, **kwargs)
         self.callback(invoke_without_command=True)(self.dump_main)
-        self.command("datamodel")(self.dump_datamodel_cmd)
+        if Flags.DUMP_DM_GLOBAL:
+            self.command("datamodel")(self.dump_datamodel_cmd_v5)
+        else:
+            self.command("datamodel")(self.dump_datamodel_cmd)
```
ronpal, line 29: Maybe you'd want to explain why it is called ...`_v5`? I assume it is because it is because it will be part of the 0.5 release?

ronpal (approved):

cognite_toolkit/_cdf_tk/apps/_dump_app.py
doctrino: Yes that is my thinking. It is a temporary way to introduce this feature behind an alpha flag.

tests/test_unit/test_cli/test_behavior.py
```diff
@@ -369,6 +368,113 @@ def test_dump_datamodel(
...
+        created_time=1,
+        name=None,
+        description=None,
+        implements=None,
+        writable=True,
+        used_for="node",
+        is_global=False,
+        filter=None,
+    )
+    default_prop_args: dict[str, Any] = dict(
```
christjt, line 387: The use of `Any` is discurraged: https://cognitedata.slack.com/archives/CDJUS0FAB/p1742557756225569. Could we add some safer type annotations (also the case for other Any usages in this PR)?

ronpal (approved):

finnag (approved):
:unicorn:

PR #1509, opened 2025-03-22
[CDF-24237] 👨‍🏫 Refactor dump data

doctrino:
# Description

**Context** The `cdf dump asset/timeseries` commands are very similar, but have separate implementations. 
This PR reimplements them with a shared implementation. No functionality is added, except one inconsistency between the two commands is fixed as mentioned below.

## Changelog

- [x] Patch
- [ ] Minor
- [ ] Skip

### Fixed

- When `cdf dump timeseries` writes to multiple csv/parquet files, all files now have the same columns.

## templates

No changes.


cognite_toolkit/_cdf_tk/commands/dump_data.py
```diff
@@ -0,0 +1,439 @@
...
+    def create_iterator(self, hierarchies: list[str], data_sets: list[str], limit: int | None) -> Iterator[AssetList]:
+        return self.client.assets(
+            chunk_size=1000,
+            asset_subtree_external_ids=hierarchies or None,
+            data_set_external_ids=data_sets or None,
+            limit=limit,
+        )
+
+    def key(self, item: Asset, hierarchies: list[str], data_sets: list[str]) -> str:
+        if hierarchies and hierarchies:
```
ronpal, line 87:
```suggestion
        if hierarchies:
```

cognite_toolkit/_cdf_tk/commands/dump_data.py
```diff
@@ -0,0 +1,439 @@
...
+            asset_external_id = self.client.lookup.assets.external_id(item.root_id or 0)
+            data_set_external_id = self.client.lookup.data_sets.external_id(item.data_set_id or 0)
+            if asset_external_id and data_set_external_id:
+                return f"{asset_external_id}.{data_set_external_id}"
+            elif asset_external_id:
+                return asset_external_id
+            elif data_set_external_id:
+                return data_set_external_id
+            return ""
+        elif hierarchies:
```
ronpal, line 97: Not sure I get this control structure: 

if hierarchies and hierarchies:
....
elif hierarchies:

ronpal (commented):
See comment in dump_data.py

cognite_toolkit/_cdf_tk/commands/dump_data.py
doctrino: Nice catch

cognite_toolkit/_cdf_tk/commands/dump_data.py
doctrino, line 87:
```suggestion
        if hierarchies and data_sets:
```

larshagencognite (approved):
risk review ok, as this follows the structure of the previous code. The code does contain rather large, complex functions that would probably benefit from being broken down and tested individually.

PR #1510, opened 2025-03-24
[CDF-24370] 🧑‍💻Unchanged Streamlit app detected as changed

doctrino:
# Description

The StreamlitLoader (Loader = standardized interface for CDF resources) does not account for `theme=Light` set on the server side.

## Changelog

- [x] Patch
- [ ] Minor
- [ ] Skip

### Fixed

- When running `cdf deploy`, Streamlit apps without the parameter `theme=Light` are no longer always considered changed.

## templates

No changes.

### DIFF
--- a/tests/data/complete_org/modules/my_example_module/streamlit/myapp.Streamlit.yaml
+++ b/tests/data/complete_org/modules/my_example_module/streamlit/myapp.Streamlit.yaml
@@ -3,7 +3,6 @@ name: MySuperApp
 creator: doctrino@github.com
 description: This is a super app
 published: true
-theme: Light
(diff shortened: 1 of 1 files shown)


tests/data/complete_org/modules/my_example_module/streamlit/myapp.Streamlit.yaml
```diff
@@ -3,7 +3,6 @@ name: MySuperApp
 creator: doctrino@github.com
 description: This is a super app
 published: true
-theme: Light
```
doctrino, line 6: Removing default value to trigger bug

larshagencognite (approved):
risk review ok

PR #1511, opened 2025-03-25
[CDF-24339] Ignore environment variable syntax in workflow version tasks 

doctrino:
# Description

We use the same syntax for environment variables, `${ENV_VAR}`, as the workflow task definition parameters `${workflow.output.someKey}`. 

## Changelog

- [x] Patch
- [ ] Minor
- [ ] Skip

### Fixed

- Using a reference syntax in a `WorkflowVersion` task no longer gives a warning that an environment variable is missing or tries to replace the jsonPath with an environment variable.

## templates

No changes.

### DIFF
--- a/tests/test_integration/test_loaders/test_resource_loaders.py
+++ b/tests/test_integration/test_loaders/test_resource_loaders.py
@@ -1,5 +1,7 @@
 from asyncio import sleep
 from contextlib import suppress
+from pathlib import Path
+from unittest.mock import MagicMock
 
 import pytest
 from cognite.client import CogniteClient
@@ -15,6 +17,7 @@ from cognite.client.data_classes import (
     FunctionSchedulesList,
     FunctionScheduleWrite,
     FunctionScheduleWriteList,
+    FunctionTaskParameters,
     GroupWrite,
     LabelDefinitionWrite,
     TimeSeriesWrite,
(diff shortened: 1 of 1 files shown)


ronpal (commented):
What happens if a user actually wants to use env vars? Is that a legit case?

```
workflowExternalId: myWorkflow
version: v1
workflowDefinition:
  description: Two tasks with reference
  tasks:
  - externalId: myTask2
    type: function
    parameters:
      function:
        externalId: fn_first_function
  - externalId: myTask2
    type: function
    parameters:
      function:
        externalId: ${myTask1.output.nextFunction}
        data: ${myenvvar}
    dependsOn:
    - externalId: myTask1
```

doctrino:
> function:
>         externalId: ${myTask1.output.nextFunction}
>         data: ${myenvvar}

We simply do not support it.

ronpal:
> We simply do not support it.

Fair enough.

ronpal (approved):

tests/test_integration/test_loaders/test_resource_loaders.py
```diff
@@ -647,3 +653,41 @@ def test_dump_cdf_group_with_invalid_reference(self, toolkit_client: ToolkitClie
...
+version: v1
+workflowDefinition:
+  description: Two tasks with reference
+  tasks:
+  - externalId: myTask2
+    type: function
+    parameters:
+      function:
+        externalId: fn_first_function
+  - externalId: myTask2
```
psalaberria002, line 670: Should both tasks have the same externalId?

psalaberria002 (approved):
:unicorn:

PR #1519, opened 2025-03-28
Release 0.5.0

doctrino:
# Description

Release `0.5.0`. This is done by removing the alpha flags for the 4 features that are planned for v5.

## Changelog

- [ ] Patch
- [x] Minor
- [ ] Skip

### Changed

- The `cdf dump datamodel` no longer includes data models that are global (in system space). You can get the original behavior by using the `--include-global` flag.
- Toolkit now stores a hash of the credentials of Workflow/Transformation/Function in the resources such that the resource is updated when the credentials change.
- For Workflow/Transformations/Function Toolkit no longer falls back to Toolkit credentials when validation-type != 'dev' in the `config.[env].yaml`. 

### Added
- Toolkit now support for `cdf dump workflow/transformation/group/node`.

## templates

No changes.

### DIFF
--- a/cdf.toml
+++ b/cdf.toml
@@ -9,10 +9,6 @@ module-repeat = true
 dump-extended = true
 populate = true
 strict-validation = true
-# Setting this to true wil change all snapshots for WorkflowTriggers/FunctionSchedules.
-# For simplicity, (avoid keeping track of two sets of snapshots), we keep this to false in development.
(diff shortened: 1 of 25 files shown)


ronpal (approved):
Let's go!

psalaberria002 (approved):
:unicorn:

PR #1521, opened 2025-03-31
[CDF-24411] 	🐛 Fix skip workflow trigger

doctrino:
# Description

The base class `CogniteResourceList` in the PySDK takes an iterable, but it exhausted the iterable in a checking, thus it practice it has to take a sequence.

## Changelog

- [x] Patch
- [ ] Minor
- [ ] Skip

### Fixed

- When running `cdf dump workflow` the Toolkit now includes triggers if the selected workflow has triggers.
## templates

No changes.


ronpal (approved):

psalaberria002 (approved):
:unicorn:

PR #1526, opened 2025-04-01
[CDF-24437] Agent CRUD Alpha support

ronpal:
# Description

Alpha feature that uses the Agent CRUD API: https://pr-2829.specs.preview.cogniteapp.com/20230101-alpha.json.html#tag/Agents/operation/main_api_v1_projects__projectName__ai_agents_post

## Changelog

- [x] Patch
- [ ] Minor
- [ ] Skip

### Added

- agent resource type support (alpha flag)

## templates

No changes.

### DIFF
--- a/cognite_toolkit/_cdf_tk/client/_toolkit_client.py
+++ b/cognite_toolkit/_cdf_tk/client/_toolkit_client.py
@@ -5,6 +5,9 @@ from typing import Literal, cast
 from cognite.client import ClientConfig, CogniteClient
 from cognite.client.credentials import CredentialProvider
 
+from cognite_toolkit._cdf_tk.feature_flags import FeatureFlag, Flags
+
+from .api.agents.agents import AgentsAPI
 from .api.dml import DMLAPI
 from .api.location_filters import LocationFiltersAPI
 from .api.lookup import LookUpGroup
@@ -76,6 +79,8 @@ class ToolkitClient(CogniteClient):
         self.dml = DMLAPI(self._config, self._API_VERSION, self)
         self.verify = VerifyAPI(self._config, self._API_VERSION, self)
         self.lookup = LookUpGroup(self._config, self._API_VERSION, self)
+        if FeatureFlag.is_enabled(Flags.AGENTS):
+            self.agents = AgentsAPI(self._config, self._API_VERSION, self)
 
     @property
     def config(self) -> ToolkitClientConfig:
--- a/cognite_toolkit/_cdf_tk/client/_toolkit_client.py
+++ b/cognite_toolkit/_cdf_tk/client/_toolkit_client.py
@@ -5,8 +5,6 @@ from typing import Literal, cast
 from cognite.client import ClientConfig, CogniteClient
 from cognite.client.credentials import CredentialProvider
 
-from cognite_toolkit._cdf_tk.feature_flags import FeatureFlag, Flags
-
 from .api.agents.agents import AgentsAPI
 from .api.dml import DMLAPI
 from .api.location_filters import LocationFiltersAPI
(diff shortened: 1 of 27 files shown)


tests/test_unit/test_cli/test_build_deploy_snapshots/cdf_entity_matching.yaml
```diff
@@ -161,7 +161,7 @@ Workflow:
 WorkflowTrigger:
 - externalId: entity_matching_trigger
   metadata:
-    cognite-toolkit-auth-hash: f13a9e42
+    cognite-toolkit-auth-hash: 9d622119
```
doctrino, line 164: Why is this changed? This seems to be unrelated to the PR.

cognite_toolkit/_cdf_tk/client/api/agents/agents.py
```diff
@@ -0,0 +1,107 @@
...
+        headers = {"cdf-version": "alpha"}
+        return self._create_multiple(
+            list_cls=AgentList,
+            resource_cls=Agent,
+            items=agent,
+            input_resource_cls=AgentWrite,
+            headers=headers,
+        )
+
+    def retrieve(self, external_id: str | SequenceNotStr[str]) -> AgentList:
```
doctrino, line 61: In the PySDK, we use overload to return `Agent | AgenList` depending on whether it is a sequence of strings or a single strings. Since this is only used by Toolkit, we always have sequence of strings, so suggest the following.
```suggestion
    def retrieve(self, external_id: SequenceNotStr[str]) -> AgentList:
```

cognite_toolkit/_cdf_tk/client/_toolkit_client.py
```diff
@@ -76,6 +79,8 @@ def __init__(self, config: ToolkitClientConfig | None = None) -> None:
         self.dml = DMLAPI(self._config, self._API_VERSION, self)
         self.verify = VerifyAPI(self._config, self._API_VERSION, self)
         self.lookup = LookUpGroup(self._config, self._API_VERSION, self)
+        if FeatureFlag.is_enabled(Flags.AGENTS):
```
doctrino, line 82: Personal preference: No need for the feature flag here. This is internal so will not be exposed to the end user anyways. You already remove the loader if the flag is not enabled.

cognite_toolkit/_cdf_tk/loaders/_resource_loaders/agent_loaders.py
```diff
@@ -0,0 +1,99 @@
...
+    ) -> Capability | list[Capability]:
+        return []
+
+    def load_resource_file(
+        self, filepath: Path, environment_variables: dict[str, str | None] | None = None
+    ) -> list[dict[str, Any]]:
+        loaded = super().load_resource_file(filepath, environment_variables)
+        for item in loaded:
+            # the API always returns instructions
+            if "instructions" not in item:
```
doctrino, line 51: Hmm, suggest not doing this and handling in the `.dump_resource` instead. As this will have push an empty instructions into the agent config if you do cdf modules pull.

cognite_toolkit/_cdf_tk/loaders/_resource_loaders/agent_loaders.py
```diff
@@ -0,0 +1,99 @@
...
+            return len(ids)
+        except CogniteAPIError:
+            deleted = 0
+            for id in ids:
+                try:
+                    self.client.agents.delete(id)
+                    deleted += 1
+                except CogniteAPIError:
+                    pass
+            return deleted
```
doctrino, line 91: Nitpick, it is best practise to only have the code line(s) that can trigger the exception inside the try... except block
```suggestion
        try:
            self.client.agents.delete(ids)
        except CogniteAPIError:
            deleted = 0
            for id in ids:
                try:
                    self.client.agents.delete(id)
                    deleted += 1
                except CogniteAPIError:
                    pass
            return deleted
        else:
            return len(ids)
```

doctrino (commented):
Good stuff, I think you need to revert the snapshot changes.

larshagencognite (approved):
🦄

PR #1530, opened 2025-04-03
[CDF-24340] 😵‍💫 Report label update correctly

doctrino:
# Description

## Before
image

## After
image

Note there is also a soft delete issue for labels, but suggest we do not address it.
image

## Changelog

- [x] Patch
- [ ] Minor
- [ ] Skip

### Fixed

- When running `cdf deploy` on label resources, changes are no longer incorrectly reported as update, but instead delete + create.

## templates

No changes.


ronpal (approved):

thorkildcognite:
risk review ok

thorkildcognite:
Fixed by doing the merge now... It seems I need to get some training on how to use a mouse again.

PR #1531, opened 2025-04-04
[CDF-24461] 🙃  Support Infield Configuration Part 1

doctrino:
# Description

Today Toolkit implicitly supports Infield configurations through supporting Nodes. This is causing complexity on the user side as the Infield configuration contains for example `dataSetId` which the user is then responsible for looking up across CDF projects. This PR adds explicit support for the Infield Configuration. 

This PR adds the data class for the APM Config node.

## Changelog

- [ ] Patch
- [ ] Minor
- [x] Skip


ronpal (approved):

psalaberria002 (approved):
:unicorn: 

How is this gonna be kept up to date when the APM schema changes? Could we use protos somehow instead of hardcoding all of these?

PR #1535, opened 2025-04-13
repo: accommodate different letter casings in --host= option

ivartj:
# Description

I was surprised to see no `.github/workflows/deploy.yaml` file after calling `cdf-tk repo init --host=Github`. The file is only created when the option matches "GitHub" with a capital H.

## Changelog

- [x] Patch
- [ ] Minor
- [ ] Skip

### Changed

- Currently the `--host=` option value is ignored if it does not case-sensitively match a value in the `REPOSITORY_HOSTING` array. This change makes the matching case-insensitive and falls back to the "Other" repository hosting option if no match is found.

## templates

No changes.


doctrino (approved):
Good catch

PR #1540, opened 2025-04-23
[CDF-24561] 👷 Introduce Toolkit Resource Classes

doctrino:
# Description

This PR introduces Toolkit resource classes. The difference between these and the write/read (request/response) class in the cognit-sdk, it that these are based on pydantic and matches the YAML file format that toolkit expects. This is similar to the write (request) format in the PySDK with the following two differences:

* Pydantic based enables giving much better error messages to the end user when they are not following the spec. See #1537 for a prototype. 
* Toolkit does small adjustments to write (request) format, for example, it uses `dataSetExternalId` instead of `dataSetId`, i.e., using external IDs instead of internal IDs. Another example is credentials, Toolkit allows the uses to specify `clientId+clientSecret` and Toolkit will create the `nonced. 

I am only adding the TimeSeries here to instroduce a simple example, I also do not use it yet in the build command. That will be a separate PR.

## Changelog

- [ ] Patch
- [ ] Minor
- [x] Skip


cognite_toolkit/_cdf_tk/resource_classes/timeseries.py
```diff
@@ -0,0 +1,15 @@
+from .base import ToolkitResource
+
+
+class TimeSeriesTK(ToolkitResource):
+    externalId: str | None = None
```
ronpal, line 5: externalId should be mandatory

doctrino: fixed

cognite_toolkit/_cdf_tk/resource_classes/timeseries.py
```diff
@@ -0,0 +1,15 @@
+from .base import ToolkitResource
+
+
+class TimeSeriesTK(ToolkitResource):
```
ronpal, line 4:
```suggestion
class TimeSeriesResource(ToolkitResource):
```
Too long/verbose?

cognite_toolkit/_cdf_tk/resource_classes/__init__.py
```diff
@@ -0,0 +1,16 @@
...
+1. Toolkit resource classes (this module): Represent the YAML configuration format the Toolkit uses.
+2. Write resource classes (from cognite-sdk): Represent the write/request format the Cognite resources.
+3. Read resource classes (from cognite-sdk): Represent the read/response format the Cognite resources.
+"""
+
+from .base import ToolkitResource
+from .timeseries import TimeSeriesTK
+
+__all__ = [
+    "TimeSeriesTK",
```
ronpal, line 14:
```suggestion
    "TimeSeriesResource",
```

ronpal (dismissed):
Naming convention doesn't sit right with me, But I can live with it.

Aashutosh-cognite (approved):
LGTM. WRT to naming conventions, in my opinion `TimeSeriesResource` is a good way because these represent dataset resource types.

ronpal (dismissed):

ronpal (approved):

larshagencognite (approved):
🦄

PR #1541, opened 2025-04-23
[CDF-24566] 🚅 Move to UV

doctrino:
# Description

Replace poetry with uv as the package manager.

## Changelog

- [ ] Patch
- [ ] Minor
- [x] Skip


psalaberria002 (approved):
🦄

PR #1543, opened 2025-04-24
[CDF-24563] Topological sort to fix dependencies between location filters

ronpal:
# Description

Due to the Location API design, nested locations have to be created in hierarchical order and one by one to make sure that parent exists before child. Also, when retrieving locations on the list endpoint, they are nested. Since our lookup depended on iterating the list, we didn't find child locations.  

Screenshot 2025-04-30 at 09 30 49

## Changelog

- [x] Patch
- [ ] Minor
- [ ] Skip

### Fixed

- Creating nested LocationFilters would fail under some circumstances. This is now fixed.
- The report of deployed changes would be inaccurate for nested LocationFilters. This is now fixed.  

## templates

No changes.


cognite_toolkit/_cdf_tk/loaders/_resource_loaders/location_loaders.py
```diff
@@ -157,13 +183,24 @@ def create(self, items: LocationFilterWrite | LocationFilterWriteList) -> Locati
 
         created = []
         for item in items:
+            if item._parent_external_id and item.parent_id == -1:
+                item.parent_id = self.client.lookup.location_filters.id(item._parent_external_id)
```
ronpal, line 187: Deferred lookup

cognite_toolkit/_cdf_tk/client/api/lookup.py
```diff
@@ -249,7 +250,7 @@ def _lookup(self) -> None:
     def _id(self, external_id: SequenceNotStr[str]) -> dict[str, int]:
         if not self._has_looked_up:
             self._lookup()
-        return {external_id: self._cache[external_id] for external_id in external_id}
+        return {external_id: self._cache[external_id] for external_id in external_id if external_id in self._cache}
```
doctrino, line 253: I am not sure this is the correct solution. Seeing how this is used when `lookup.id(...)` is called it is assumed that each external ID will result in an internal ID. I think this is worth a discussion.

cognite_toolkit/_cdf_tk/client/api/lookup.py
```diff
@@ -276,16 +277,28 @@ def _read_acl(self) -> Capability:
...
     def _read_acl(self) -> Capability:
         return LocationFiltersAcl(
             [LocationFiltersAcl.Action.Read],
             scope=LocationFiltersAcl.Scope.All(),
         )
 
+    def _id(self, external_id: SequenceNotStr[str]) -> dict[str, int]:
+        if not self._has_looked_up:
+            self._lookup()
+        return {external_id: self._cache[external_id] for external_id in external_id if external_id in self._cache}
```
doctrino, line 300: Same issue here, I think we need to have a discussion on how to handle the case when an `externalID` is missing in CDF. I think it should give a warning to the user and not pass by silently.

cognite_toolkit/_cdf_tk/client/data_classes/location_filters.py
```diff
@@ -160,6 +161,7 @@ def __init__(
         self.asset_centric = asset_centric
         self.views = views
         self.data_modeling_type = data_modeling_type
+        self._parent_external_id = _parent_external_id
```
doctrino, line 164: A small comment on why this is here

cognite_toolkit/_cdf_tk/loaders/_resource_loaders/location_loaders.py
```diff
@@ -56,6 +59,22 @@ class LocationFilterLoader(
 
     subfilter_names = ("assets", "events", "files", "timeseries", "sequences")
 
+    def __init__(self, client: ToolkitClient, build_dir: Path | None, console: Console | None = None) -> None:
+        self.client = client
```
doctrino, line 63: missing a super call
```suggestion
super().__init__(client, build_dir, console)
```

cognite_toolkit/_cdf_tk/loaders/_resource_loaders/location_loaders.py
```diff
@@ -56,6 +59,22 @@ class LocationFilterLoader(
...
+        self.resource_build_path: Path | None = None
+
+        # location filters may have to be deployed in a specific order since the API doesn't
+        # support parentExternalId so we need to deploy them one by one
+        if build_dir and build_dir.is_file():
+            self.resource_build_path = build_dir
```
doctrino, line 69: I don't understand this modification of the resource_build_path. It should never be a file always a directory

cognite_toolkit/_cdf_tk/loaders/_resource_loaders/location_loaders.py
```diff
@@ -102,7 +121,14 @@ def safe_read(self, filepath: Path | str) -> str:
...
     def load_resource(self, resource: dict[str, Any], is_dry_run: bool = False) -> LocationFilterWrite:
         if parent_external_id := resource.pop("parentExternalId", None):
-            resource["parentId"] = self.client.lookup.location_filters.id(parent_external_id, is_dry_run)
+            # TODO: this will draw blanks if the parent is in the current build but has not been deployed yet
+            #
+            try:
+                resource["parentId"] = self.client.lookup.location_filters.id(parent_external_id, is_dry_run)
+            except ResourceRetrievalError:
+                resource["parentId"] = -1
+                resource["_parentExternalId"] = parent_external_id
```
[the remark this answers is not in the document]
doctrino: This comment is worth putting into the code :)

cognite_toolkit/_cdf_tk/loaders/_resource_loaders/location_loaders.py
doctrino: Here I think it is worth a longer comment that we have to deploy a location filter one by one. In case the create endpoint support multiple in the future, it is good to know why we want to do it this way.

doctrino (commented):
Very good work. I like the work arounds, both single file (all though not sure that is necessary as long as you sort them well) and the lookup later solution.

doctrino (approved):
Let's go!

cognite_toolkit/_cdf_tk/builders/_location.py
```diff
@@ -0,0 +1,103 @@
...
+            yield BuildDestinationFile(
+                path=destination_path,
+                loaded=location,
+                loader=loader,  # type: ignore[arg-type]
+                source=build_source_file.source,
+                extra_sources=None,
+                warnings=warnings,
+            )
+
+    def _create_file_path(self, source_path: Path, index: int, kind: str) -> Path:
```
finnag, line 85: This function needs a test, and if the test is good enough, maybe you don't need to describe what it tries to do.

cognite_toolkit/_cdf_tk/builders/_location.py
```diff
@@ -0,0 +1,103 @@
...
+                        location_hierarchy_graph[ext_id].append(parent_external_id)
+
+            warnings = WarningList[FileReadWarning]()
+
+        ordered_locations = []
+        try:
+            for external_id in TopologicalSorter(location_hierarchy_graph).static_order():
+                (location, _) = location_by_external_id[external_id]
+                ordered_locations.append(location)
+        except CycleError:
```
finnag, line 66: should be a test somewhere that tests for cycleerror

cognite_toolkit/_cdf_tk/builders/_location.py
```diff
@@ -0,0 +1,103 @@
...
+        """
+        filestem = source_path.stem
+        # Get rid of the local index
+        filestem = INDEX_PATTERN.sub("", filestem)
+
+        # Increment to ensure we do not get duplicate filenames when we flatten the file
+        # structure from the module to the build directory.
+        self.resource_counter = index
+
+        filename = f"{self.resource_counter}.{filestem}"
```
finnag, line 97: An earlier comment says

> 3. create a new file for each location where the prefix index ensures deployment order

Does this mean this counter needs to sort correctly, and that it can fail if you get into double digits? Does it need to be 0-prefixed to be safe, for example?

cognite_toolkit/_cdf_tk/builders/_location.py
```diff
@@ -0,0 +1,103 @@
...
+        if not filename.casefold().endswith(kind.casefold()):
+            filename = f"{filename}.{kind}"
```
finnag, line 99: what kind of filename do you start with if the stem ends with kind? It would be at least somewhat interesting to see the test case that exercises this code

finnag (changes requested):
This code would be in a happier place with some tests. 
Added some comments where clarification is needed.

cognite_toolkit/_cdf_tk/builders/_location.py
ronpal: Added test_location_builder_detect_cyclic_references and test_location_builder_detect_self_reference in unit tests

doctrino (approved):

finnag (approved):
:unicorn:

PR #1545, opened 2025-04-25
[CDF-24589] 👮 Use pydantic class for error validation

doctrino:
# Description

Use the new `pydantic` class to replace the `get_paramter_spec` for giving error message. So far we only support `TimeSeries`. 

#### Before
image

#### After
image

## Changelog

- [x] Patch
- [ ] Minor
- [ ] Skip

### Improved

- Improved warnings if a `TimeSeries` configuration is not following the expected spec.

## templates

No changes.


ronpal (approved):
Smooth!

psalaberria002 (approved):
:unicorn:

PR #1546, opened 2025-04-29
🧑‍🔧Fix failing test

doctrino:
# Description

We have a functionality `cdf run function local` which creates a Python virtual environment to execute a CogniteFunction locally. This is causing issues as we are running the test suite in with 8 workers and have two tests for this functionality. This leads them to compete for the same virtual environment which on Windows ends with `Permission` error and on Linux/Max `OSError: [Errno 26] Text file busy`. 

This PR solves the issue by letting the location of the virtual environment folder be an argument and set it differently for the two tests, such that they create one virtual environment each.

## Changelog

- [ ] Patch
- [ ] Minor
- [x] Skip


ronpal (approved):

larshagencognite (approved):
🦄

PR #1549, opened 2025-04-29
😐 Fix git ignore

doctrino:
# Description

The fix in #1546 introduced a fix to the virtual environment. This means all folders starting with `function_local_venvs` must ignored.

## Changelog

- [ ] Patch
- [ ] Minor
- [x] Skip


ronpal (approved):

larshagencognite (approved):
🦄

PR #1560, opened 2025-05-06
[CDF-23856] 🚶 Pydantic Group

doctrino:
# Description
Toolkit: CLI for managing CDF configuration as code. The CDF resources are configured in YAML files that are typically checked into version control. 

Context: Up until now we have used the `cognite-sdk` data classes when loading these YAML files. This gives poor error messages when there are errors in the YAML files. In addition, Toolkit has a few deviation from the API spec, the most common Toolkit always uses external IDs instead of internal IDs. We have made the decision to implement `pydantic` to be able to give better error messages to the user when they mistype their `YAML` files.

Reason for size of this PR: Our Groups API has a lot of capabilities.

## Changelog

- [x] Patch
- [ ] Minor
- [ ] Skip

### Improved

- Improved error message if group is not following the API specifications. 

## templates

No changes.

### DIFF
--- a/cognite_toolkit/_cdf_tk/resource_classes/capabilities.py
+++ b/cognite_toolkit/_cdf_tk/resource_classes/capabilities.py
@@ -1,0 +1,110 @@
+import sys
+from collections.abc import Sequence
+from types import MappingProxyType
+from typing import Any, ClassVar, Literal, cast
+
+from pydantic import BaseModel, ModelWrapValidatorHandler, field_validator, model_serializer, model_validator
+from pydantic_core.core_schema import SerializerFunctionWrapHandler
+
+from cognite_toolkit._cdf_tk.utils.collection import humanize_collection
+
+if sys.version_info < (3, 11):
+    from typing_extensions import Self
+else:
+    from typing import Self
+
+from .base import BaseModelResource
+
+
+class Scope(BaseModelResource):
+    _scope_name: ClassVar[str]
+
+    @model_validator(mode="wrap")
+    @classmethod
+    def find_scope_cls(cls, data: Any, handler: ModelWrapValidatorHandler[Self]) -> Self:
+        if isinstance(data, Scope):
+            return cast(Self, data)
+        if not isinstance(data, dict):
+            raise ValueError(f"Invalid scope data '{type(data)}' expected dict")
+
+        if cls is not Scope:
+            return handler(data)
+        name, content = next(iter(data.items()))
+        if name not in _SCOPE_CLASS_BY_NAME:
+            raise ValueError(
+                f"Invalid scope name '{name}'. Expected one of {humanize_collection(_SCOPE_CLASS_BY_NAME.keys(), bind_word='or')}"
+            )
+        cls_ = _SCOPE_CLASS_BY_NAME[name]
+        return cast(Self, cls_.model_validate(content))
+
+    @model_serializer(mode="wrap", when_used="always", return_type=dict)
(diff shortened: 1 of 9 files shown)


tests/test_unit/test_cdf_tk/test_commands/test_build.py
```diff
@@ -114,7 +114,7 @@ def test_build_no_warnings_when_space_exists_in_cdf(
         my_group = """name: gp_trigger_issue
 sourceId: '1234567890123456789'
 capabilities:
-- dataModelInstancesAcls:
+- dataModelInstancesAcl:
```
doctrino, line 117: Bug in test data

thorkildcognite:
Exemption given based on the overhead for specifications in the group tests.

cognite_toolkit/_cdf_tk/resource_classes/capabilities.py
```diff
@@ -0,0 +1,507 @@
...
+    dbs_to_tables: dict[str, list[str] | dict[Literal["tables"], list[str]]]
+
+
+class AssetRootIDScope(Scope):
+    _scope_name = "assetRootIdScope"
+    root_ids: list[str]
+
+
+class ExperimentsScope(Scope):
+    _scope_name = "experimentscope"
```
ronpal, line 98: Suggest sticking with singular 

```suggestion
class ExperimentScope(Scope):
    _scope_name = "experimentscope"
```

cognite_toolkit/_cdf_tk/resource_classes/capabilities.py
```diff
@@ -0,0 +1,507 @@
...
+    def include_capability_name(self, handler: SerializerFunctionWrapHandler) -> dict:
+        if self._capability_name is None:
+            raise ValueError("Capability name is not set")
+        serialized_data = handler(self)
+        return {self._capability_name: serialized_data}
+
+
+class AnalyticsAcl(Capability):
+    _capability_name = "analyticsAcl"
+    actions: list[Literal["READ", "EXECUTE", "LIST"]]
```
ronpal, line 168: More of a question since @Aashutosh-cognite brought it up earlier: is there a case for using enums instead of literal?

ronpal (dismissed):
Small comments/questions

ronpal (approved):

psalaberria002:
Please update the PR description to explain what this PR does. I feel like it is doing more than what the pr title and description say.

cognite_toolkit/_cdf_tk/resource_classes/capabilities.py
```diff
@@ -0,0 +1,507 @@
...
+class AnnotationsAcl(Capability):
+    _capability_name = "annotationsAcl"
+    actions: list[Literal["READ", "WRITE", "SUGGEST", "REVIEW"]]
+    scope: AllScope
+
+
+class AssetsAcl(Capability):
+    _capability_name = "assetsAcl"
+    actions: list[Literal["READ", "WRITE"]]
+    scope: AllScope | DataSetScope
```
psalaberria002, line 181: Are all of these classes manually created? Where is the source of truth for these? Could they be autogenerated from protos?

doctrino: These are manually created. Source of truth is the OPENAPI spec. They can unfortunately not be generated as Toolkit has some deviation, for example, all cases were the API expects an internal ID, Toolkit uses an external ID such that the same configuration can be used across multiple CDF projects.

psalaberria002 (approved):
:unicorn:

PR #1569, opened 2025-05-15
Allow upper case

doctrino:
# Description

Allow upper case when in the changelog below in the release process.

## Changelog

- [ ] Patch
- [ ] Minor
- [x] Skip


ronpal (approved):

rsjr (commented):
risk review ok

PR #1573, opened 2025-05-16
[CDF-24720] 🚶‍♂️‍➡️Profile command

doctrino:
# Description

This is a new alpha feature with the first use case to evaluate new CDF projects for upgrade.
image

Will extend it with writing to file and give more granular results as well as more resources. But stopping for now to keep the PR small.

## Changelog

- [x] Patch
- [ ] Minor
- [ ] Skip

### Added

- [alpha] New command `cdf profile`. Enabled with `profile = true` in the [alpha_flags] of section `cdf.toml` .

## templates

No changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -225,7 +225,6 @@ pyrightconfig.json
 
 ### VisualStudioCode ###
 .vscode/*
-!.vscode/settings.json
 !.vscode/tasks.json
 !.vscode/launch.json
 !.vscode/extensions.json
--- a/cognite_toolkit/_cdf_tk/loaders/_resource_loaders/function_loaders.py
+++ b/cognite_toolkit/_cdf_tk/loaders/_resource_loaders/function_loaders.py
@@ -139,7 +139,9 @@ class FunctionLoader(ResourceLoader[str, FunctionWrite, Function, FunctionWriteL
 
     @classmethod
     def _create_hash_values(cls, function_rootdir: Path) -> str:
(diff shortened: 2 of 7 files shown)


cognite_toolkit/_cdf_tk/utils/cdf.py
ronpal, line 1: Had no idea you could query directly like that. Nice!

ronpal (dismissed):

finnag (commented):
I am missing some context that is obvious to you. I would expect a new command to come with some documentation somewhere that talks about it?

Why is this command useful, how do you know what it does? 

Do you have a hidden wiki page somewhere telling people about the command and its various options and when to use them?

Is the output from --help enough for anyone to figure it out?

doctrino:
> I am missing some context that is obvious to you. I would expect a new command to come with some documentation somewhere that talks about it?

We update the docs after it is released here: https://docs.cognite.com/cdf/deploy/cdf_toolkit/api/alpha_flags

>
> Why is this command useful, how do you know what it does?

This command gives an overview over the metadata and labels for each of the asset-centric resources. This shows roughly how much unstructured data there is, which we use to determined the effort to move over to data modeling. 

> Do you have a hidden wiki page somewhere telling people about the command and its various options and when to use them?
> 
> Is the output from --help enough for anyone to figure it out?

Not at this stage, but this command is behind an alpha flag and is currently in progress. There will be more PRs once this one is merged.

finnag (approved):
:unicorn:

ronpal (approved):

finnag (approved):
:unicorn:

PR #1575, opened 2025-05-16
[CDF-24583] Pydantic Validation for location

Aashutosh-cognite:
# Description

This PR introduces pydantic model for Location Resource class that matches the YAML file format that toolkit expects.

## Changelog

- [ ] Patch
- [ ] Minor
- [x] Skip


doctrino (commented):
This is a tricky API as it is poorly documented. There are still a few more adjustments.

doctrino (approved):

cognite_toolkit/_cdf_tk/resource_classes/location.py
```diff
@@ -0,0 +1,67 @@
...
+    timeseries: AssetCentricResourceData | None = Field(default=None, description="Timeseries resource type data")
+    files: AssetCentricResourceData | None = Field(default=None, description="File resource type data")
+    sequences: AssetCentricResourceData | None = Field(default=None, description="Sequence resource type data")
+
+
+class LocationYAML(ToolkitResource):
+    external_id: str = Field(description="The external ID provided by the client.")
+    name: str = Field(description="The name of the location.")
+    description: str | None = Field(default=None, description="The description of the data set.", max_length=255)
+    parent_external_Id: str | None = Field(default=None, description="The external ID of the parent location.")
```
psalaberria002, line 52: is the capital I in parent_external_Id intentional? it is not consistent with other fields such as external_id in the same class.

Aashutosh-cognite: This is a typo, will fix it and will ping you once it is done :)

psalaberria002:
I don't like that the tests did not catch the typo. Is the case irrelevant for Pydantic?

doctrino:
> I don't like that the tests did not catch the typo. Is the case irrelevant for Pydantic?

This would not have impacted the users as we use an alias generator to get camelCase aliases which is what pydantic use to validate.

doctrino (approved):

psalaberria002 (approved):
🦄

PR #1576, opened 2025-05-16
Updating pydantic

ronpal:
# Description

the CLI fails if the pydantic version is lower than 2.11

## Changelog

- [x] Patch
- [ ] Minor
- [ ] Skip

### Fixed

- Upgraded pydantic version

## templates

No changes.


pyproject.toml
```diff
@@ -66,7 +66,7 @@ package = true
 show_missing = true
 
 [tool.pytest.ini_options]
-minversion = 8.0
+minversion = "8.0"
```
ronpal, line 69: Got complaints from local linter...

doctrino (approved):

rsjr (commented):
risk review ok

PR #1580, opened 2025-05-20
fix configuration value for workflows, fix bug with datapoints extractor

petreeb:
# Description

Fixed Bootcamp workflows using incorrect credentials, fixed bug introduced in last change for datapoints extractor.

## Changelog

- [ ] Patch
- [ ] Minor
- [ ] Skip

### Fixed

- Bootcamp config.test.yaml had incorrect credentials for workflow variables
- fixed bug that would miss list of datapoints when list was < 50

## templates

No changes.

### DIFF
--- a/cognite_toolkit/_builtin_modules/bootcamp/ice_cream_api/default.config.yaml
+++ b/cognite_toolkit/_builtin_modules/bootcamp/ice_cream_api/default.config.yaml
@@ -3,7 +3,7 @@ icapi_extractors_source_id: <OBJECT ID FOR ICAPI EXTRACTORS ENTRA GROUP>
 tokenUri: ${IDP_TOKEN_URL}
 cdfProjectName: ${CDF_PROJECT}
 scopes: ${IDP_SCOPES}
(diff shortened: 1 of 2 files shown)


David-Alvarez-Cognite (approved):

ronpal (approved):
LGTM

larshagencognite:
This seems like a good motivation for why we should have (and check) typing in these examples, no?

larshagencognite:
The fact that these samples will be copied around and used as best-practice for how to use CDF is not really a good motivation to avoid typing.

petreeb:
We tell the participants that they will never use our datapoints extractor outside of this course. I'll start working on typing with my next update but we need these bug fixes available for the next bootcamp on Monday in Japan.

doctrino:
> The fact that these samples will be copied around and used as best-practice for how to use CDF is not really a good motivation to avoid typing.

Yes, you are right. However, we are in the process of removing these templates from this repo and keep them separate. As this is needed in a bootcamp soon, it would be good to get this out now.

larshagencognite (approved):
🦄

PR #1584, opened 2025-05-21
[CDF-24131] Add optional pyarrow dependency

doctrino:
# Description

This is used when dumping to parquet.

## Changelog

- [x] Patch
- [ ] Minor
- [ ] Skip

### Added

- Optional pyarrow dependency. You can install this with `pip install cognite-toolkit[table]`. 

## templates

No changes.


rsjr (commented):
risk review ok

PR #1588, opened 2025-05-22
feat; extend to filtering of datasets

doctrino:
# Description

No change to the code that impacts the user. This is a prep commit to reduce an upcoming refactoring of the dump method. The dump method takes assets/time series and write them from CDF to csv/parquet. Then, it needs to have the metadata keys of all assets/timeseries to know the column headers. 

This PR changes the metadata_key_counts to filter for data_set_id and rootId such that it can be used in the dump method that needs this filtering.

## Changelog

- [ ] Patch
- [ ] Minor
- [x] Skip

### DIFF
--- a/cognite_toolkit/_cdf_tk/utils/cdf.py
+++ b/cognite_toolkit/_cdf_tk/utils/cdf.py
@@ -143,20 +143,33 @@ def read_auth(
 
 
 def metadata_key_counts(
-    client: ToolkitClient, resource: Literal["assets", "events", "files", "timeseries", "sequences"]
+    client: ToolkitClient,
+    resource: Literal["assets", "events", "files", "timeseries", "sequences"],
+    data_sets: list[int] | None = None,
+    hierarchies: list[int] | None = None,
 ) -> list[dict[str, int | str]]:
     """Get the metadata key counts for a given resource.
 
     Args:
         client: ToolkitClient instance
         resource: The resource to get the metadata key counts for. Can be one of "assets", "events", "files", "timeseries", or "sequences".
+        data_sets: A list of data set IDs to filter by. If None, no filtering is applied.
+        hierarchies: A list of hierarchy IDs to filter by. If None, no filtering is applied.
 
     Returns:
         A dictionary with the metadata keys as keys and the counts as values.
     """
+    where_clause = ""
+    if data_sets is not None and hierarchies is not None:
+        where_clause = f"\n         WHERE dataSetId IN ({''.join(map(str, data_sets))}) AND rootId IN ({','.join(map(str, hierarchies))})"
+    elif data_sets is not None:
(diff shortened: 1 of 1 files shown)


ronpal (approved):
🙌

tests/test_integration/test_utils/test_cdf.py
```diff
@@ -0,0 +1,56 @@
...
+        ]
+    )
+    existing = toolkit_client.assets.retrieve_multiple(external_ids=assets.as_external_ids(), ignore_unknown_ids=True)
+    if missing := [asset for asset in assets if asset.external_id not in set(existing.as_external_ids())]:
+        created = toolkit_client.assets.create(missing)
+        existing.extend(created)
+    return existing
+
+
+class TestMetadataKeyCounts:
```
larshagencognite, line 37: no testing for datasets (single vs. multiple), multiple hierarchies, combination of datasets and hierarchies

cognite_toolkit/_cdf_tk/utils/cdf.py
```diff
@@ -143,20 +143,33 @@ def read_auth(
...
         resource: The resource to get the metadata key counts for. Can be one of "assets", "events", "files", "timeseries", or "sequences".
+        data_sets: A list of data set IDs to filter by. If None, no filtering is applied.
+        hierarchies: A list of hierarchy IDs to filter by. If None, no filtering is applied.
 
     Returns:
         A dictionary with the metadata keys as keys and the counts as values.
     """
+    where_clause = ""
+    if data_sets is not None and hierarchies is not None:
+        where_clause = f"\n         WHERE dataSetId IN ({''.join(map(str, data_sets))}) AND rootId IN ({','.join(map(str, hierarchies))})"
```
larshagencognite, line 164: bug in joining data_sets, empty string used as separator

larshagencognite: would it make sense to avoid repeating the conditions (and allowing subtle differences) by doing something like
```
data_sets_condition = f"dataSetId IN ({','.join(map(str, data_sets))})" if data_sets is not None else None
hierarchies_condition = f"rootId IN ({','.join(map(str, hierarchies))})" if hierarchies is not None else None
conditions = [c for c in [data_sets_condition,hierarchies_condition] if c is not None]
where_clause = f"\n     WHERE {' AND '.join(conditions)}" if conditions else ""
```

doctrino, line 164:
```suggestion
        where_clause = f"\n         WHERE dataSetId IN ({','.join(map(str, data_sets))}) AND rootId IN ({','.join(map(str, hierarchies))})"
```

doctrino: I chose to make it more explicit and avoid the above as it is harder to read. So I prefer keeping it as is.

ronpal (approved):

PR #1589, opened 2025-05-22
[prep CDF-24745, CDF-24554, CDF-24555] 👷Prepare dump refactoring, TableWriter class.

doctrino:
# Description

**Context**: Toolkit supports dumping assets/timeseries to csv/parquet (and yaml). That command will be refactored to support non-asset centric resources such as Canvas as part of migration support. This PR introduces a TableWriter class with subclasses for writing to parquet/csv/yaml which will be used in the refactoring. It is pulled out in this PR to keep PR size small.

## Changelog

- [ ] Patch
- [ ] Minor
- [x] Skip

### DIFF
--- a/cognite_toolkit/_cdf_tk/utils/table_writers.py
+++ b/cognite_toolkit/_cdf_tk/utils/table_writers.py
@@ -1,0 +1,103 @@
+import csv
+from abc import abstractmethod
+from dataclasses import dataclass
+from pathlib import Path
+from types import MappingProxyType
+from typing import Any, ClassVar, Literal, TypeAlias
+
+from cognite_toolkit._cdf_tk.exceptions import ToolkitValueError
+from cognite_toolkit._cdf_tk.utils import humanize_collection, to_directory_compatible
+from cognite_toolkit._cdf_tk.utils.file import yaml_safe_dump
+
+FileFormat: TypeAlias = Literal["csv", "parquet", "yaml"]
+Rows: TypeAlias = list[dict[str, Any]]
+
+
+@dataclass
+class SchemaColumn:
+    name: str
+    type: str
+
+
+@dataclass
+class Schema:
+    display_name: str
+    folder_name: str
+    kind: str
+    format_: FileFormat
+    columns: list[SchemaColumn]
+
+
+class FileWriter:
+    # 128 MB
+    file_size = 128 * 1024 * 1024
+    encoding = "utf-8"
+    newline = "\n"
+    format: ClassVar[FileFormat]
+
+    def __init__(self, schema: Schema, output_dir: Path) -> None:
+        self.schema = schema
+        self.output_dir = output_dir
+
+    @abstractmethod
+    def write_rows(self, rows_group_list: list[tuple[str, Rows]]) -> None:
+        """Write rows to a file."""
+        raise NotImplementedError("This method should be implemented in subclasses.")
+
+    @classmethod
+    def load(cls, schema: Schema, output_directory: Path) -> "FileWriter":
+        write_cls = _TABLEWRITER_CLASS_BY_FORMAT.get(schema.format_)
+        if write_cls is None:
+            raise ToolkitValueError(
+                f"Unsupported format {schema.format_}. Supported formats are {humanize_collection(_TABLEWRITER_CLASS_BY_FORMAT.keys())}."
+            )
+
+        return write_cls(schema, output_directory)
+
+
+class ParquetWrite(FileWriter):
+    format = "parquet"
+
+    def write_rows(self, rows_group_list: list[tuple[str, Rows]]) -> None:
+        pass
+
+
+class CSVWriter(FileWriter):
+    format = "csv"
+
+    def write_rows(self, rows_group_list: list[tuple[str, Rows]]) -> None:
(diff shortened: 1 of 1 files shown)


cognite_toolkit/_cdf_tk/utils/table_writers.py
```diff
@@ -0,0 +1,243 @@
...
+        for prop in self.schema.columns:
+            pa_type = self._as_pa_type(prop.type, prop.is_array)
+            fields.append(pa.field(prop.name, pa_type, nullable=True))
+        return pa.schema(fields)
+
+    @staticmethod
+    def _check_pyarrow_dependency() -> None:
+        if importlib.util.find_spec("pyarrow") is None:
+            raise ToolkitMissingDependencyError(
+                "Writing to parquet requires pyarrow. Install with 'pip install cognite-toolkit[table]'"
```
ronpal, line 165: This fails in some terminals, enclosing in single or double quotes may be necessary: 

```suggestion
                "Writing to parquet requires pyarrow. Install with 'pip install \"cognite-toolkit[table]\"'"
```

ronpal (approved):

ronpal (approved):

cognite_toolkit/_cdf_tk/utils/table_writers.py
```diff
@@ -0,0 +1,243 @@
...
+from cognite_toolkit._cdf_tk.utils import humanize_collection, to_directory_compatible
+from cognite_toolkit._cdf_tk.utils.file import yaml_safe_dump
+
+if TYPE_CHECKING:
+    import pyarrow as pa
+    import pyarrow.parquet as pq
+
+FileFormat: TypeAlias = Literal["csv", "parquet", "yaml"]
+DataType: TypeAlias = Literal["string", "integer", "float", "boolean", "datetime", "date", "time", "json"]
+Rows: TypeAlias = list[dict[str, Any]]
```
larshagencognite, line 22: is `Any` correct here, or should we use a nested json type?

tests/test_unit/test_cdf_tk/test_utils/test_table_writers.py
```diff
@@ -0,0 +1,100 @@
...
+        with CSVWriter(example_schema, output_dir) as writer:
+            writer.write_rows([("group1", [{"column1": "value1", "column2": 1, "column3": 1.0, "column4": ["value"]}])])
+            writer.write_rows([("group1", [{"column1": "value2", "column3": 2.0}])])
+
+        csv_file = list(output_dir.rglob("*.csv"))
+        assert len(csv_file) == 1
+        actual_file = csv_file[0].read_text()
+        assert actual_file == ("column1,column2,column3,column4\nvalue1,1,1.0,['value']\nvalue2,,2.0,\n")
+
+    def test_write_yaml(self, example_schema: Schema, tmp_path: Path) -> None:
```
larshagencognite, line 44: Should verify that we handle corner cases correctly for yaml, like `false` vs `"false"`. Do we support yaml 1.1 parsers or only >1.2? Older parsers might treat `yes/no/on/off/y/n` etc. as a boolean, so it needs quotes in more cases.

cognite_toolkit/_cdf_tk/utils/table_writers.py
doctrino: Did a more specific type, however, has to keep the `object` and `list` as we do support `json`.

ronpal (dismissed):

cognite_toolkit/_cdf_tk/utils/table_writers.py
```diff
@@ -18,8 +19,8 @@
     import pyarrow.parquet as pq
 
 FileFormat: TypeAlias = Literal["csv", "parquet", "yaml"]
-DataType: TypeAlias = Literal["string", "integer", "float", "boolean", "datetime", "date", "time", "json"]
-Rows: TypeAlias = list[dict[str, Any]]
+DataType: TypeAlias = Literal["string", "integer", "float", "boolean", "json"]
+Rows: TypeAlias = list[dict[str, str | int | float | bool | datetime | date | object | list | None]]
```
larshagencognite, line 23: Is a recursive definition of a json value more correct here?

larshagencognite: where `JsonVal: TypeAlias = None | str | int | float | bool | dict[str, "JsonVal"] | list["JsonVal"]`

cognite_toolkit/_cdf_tk/utils/table_writers.py
```diff
@@ -20,7 +19,8 @@
 
 FileFormat: TypeAlias = Literal["csv", "parquet", "yaml"]
 DataType: TypeAlias = Literal["string", "integer", "float", "boolean", "json"]
-Rows: TypeAlias = list[dict[str, str | int | float | bool | datetime | date | object | list | None]]
+JsonVal: TypeAlias = None | str | int | float | bool | dict[str, "JsonVal"] | list["JsonVal"]
+Rows: TypeAlias = list[dict[str, str | int | float | bool | JsonVal | None]]
```
larshagencognite, line 23: I think the fact that this expands to duplicates of identical types is a bit strange, but not blocking.

PR #1591, opened 2025-05-23
[prep CDF-24745, CDF-24554, CDF-24555] 👷Prepare dump refactoring, Interactive Select

doctrino:
# Description

**Context**: Toolkit supports dumping assets/timeseries to csv/parquet (and yaml). Today this command has a lot of functionality that is needed in other commands. In this part, I recreate the interactive select logic used to select dataset/asset hierarchy to select which assets/timeseries to dump. I will reuse this logic in for example the profile command to do the same type of selection. 

## Changelog

- [ ] Patch
- [ ] Minor
- [x] Skip

### DIFF
--- a/cognite_toolkit/_cdf_tk/utils/interactive_select.py
+++ b/cognite_toolkit/_cdf_tk/utils/interactive_select.py
@@ -1,0 +1,125 @@
+from abc import abstractmethod
+from functools import lru_cache
+
+import questionary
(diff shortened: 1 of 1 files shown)


andersfylling:
:unicorn: 

> If this piece of code fails, the user will be unable to select data set/asset hierarchy to dump data from interactively.

Sounds good. Was just not certain if this is a tool expected to run in production systems - unfortunately I'm not all too familiar with this repo.

PR #1593, opened 2025-05-23
tests: added failing test

doctrino:
# Description

Adds support for dumping events to csv/parquet.

## Changelog

- [x] Patch
- [ ] Minor
- [ ] Skip

### Added

- [alpha] Support for dumping events, `cdf dump data events`. To enable this set the flag `dump_data = true`.

## templates

No changes.


ronpal (dismissed):
⚡

ronpal (dismissed):

cognite_toolkit/_cdf_tk/apps/_dump_app.py
```diff
@@ -517,3 +523,81 @@ def dump_timeseries_cmd(
...
+        client = EnvironmentVariables.create_from_environment().get_client()
+        if hierarchy is None and data_set is None:
+            hierarchy, data_set = EventInteractiveSelect(client).interactive_select_hierarchy_datasets()
+        cmd.run(
+            lambda: cmd.dump_table(
+                EventFinder(client, hierarchy or [], data_set or []),
+                output_dir,
+                clean,
+                limit,
+                format_,  # type: ignore [arg-type]
```
einarmo, line 600: Suggestion: You could avoid this type: ignore if you shifted the `is_supported_format` method up here. You could even give it a convenient signature like `validate_format(str) -> Literal['csv', 'parquet']`. It also looks to me like this would technically allow `yaml` as format, but that isn't documented anywhere?

cognite_toolkit/_cdf_tk/commands/dump_data.py
```diff
@@ -290,6 +299,51 @@ def _get_resource_columns(self) -> list[SchemaColumn]:
...
+            SchemaColumn(name="endTime", type="integer"),
+            SchemaColumn(name="type", type="string"),
+            SchemaColumn(name="subtype", type="string"),
+            SchemaColumn(name="description", type="string"),
+            SchemaColumn(name="assetExternalIds", type="string", is_array=True),
+            SchemaColumn(name="source", type="string"),
+        ]
+        data_set_ids = self.client.lookup.data_sets.id(self.data_sets) if self.data_sets else []
+        root_ids = self.client.lookup.assets.id(self.hierarchies) if self.hierarchies else []
+        metadata_keys = metadata_key_counts(self.client, "events", data_set_ids or None, root_ids or None)
```
einarmo, line 330: This is a pretty hefty way to order your keys. Is doing it this way really necessary? I realize this is already an established pattern, but maybe explain why? The key counts can change, so the ordering could change between dumps, and running a transformation to count keys is a pretty heavy method. Is it about knowing all the columns before-hand? This is running a preview so I don't think that will actually work in every case, since previews only run on a subset of the data (?). You'd need to run a full transformation.

einarmo (commented):
The metadata thing concerns me slightly, it seems like it may have some pitfalls. On a related note I'd like to see a test that shows how the dumper behaves if there are events with different metadata keys.

cognite_toolkit/_cdf_tk/commands/dump_data.py
doctrino: Unfortunately, yes. The aggregate endpoint for files (and all other asset-centric resources) lower cases the metadata keys. This is not what our users expects, as there is typically good information in the casing of metadata. Ref https://cognitedata.slack.com/archives/C019GLXD0T0/p1736754054881239

Yes this is for knowing all the columns before writing. The way preview is used here it actually runs on all rows, see the specified limits.

ronpal (approved):

cognite_toolkit/_cdf_tk/commands/dump_data.py
einarmo: Huh. I was not aware that previews actually had no upper limit. Tested it myself now and it seems you are right. I understand, this seems to be the most reasonable solution for this, though it feels fairly heavy.

ronpal (approved):

PR #1597, opened 2025-05-27
Expose new yaml classe in module

doctrino:
# Description

We have added several `pydantic` classes which will be used for validation. To avoid merge conflicts these have not yet been included in the `__init__` file of the module. This PR does that work for the latest pydantic classes that @Aashutosh-cognite has added.

## Changelog

- [ ] Patch
- [ ] Minor
- [x] Skip


larshagencognite (approved):
🦄

PR #1599, opened 2025-05-27
[CDF-24590] 👨‍👩‍👦 Prefer parent over grandparent

doctrino:
# Description

This issue casus the following when dumping a model from one Project and redeploying it to another:
image

## Changelog

- [x] Patch
- [ ] Minor
- [ ] Skip

### Fixed

- When running `cdf dump datamodel` of a data model deployed through the UI, Toolkit now orders the implements of the views from grandparent, parent, and so on. This is in case the grandparent has a direct relation property that the parent overwrites with a new source, the parent source is used instead of the grandparent source for a child view.

## templates

No changes.


ronpal (approved):
Smooth

larshagencognite (approved):
🦄

PR #1602, opened 2025-06-02
[CDF-24796] Pydantic validation for containers

Aashutosh-cognite:
# Description

This PR introduces pydantic model for Container Resource class that matches the YAML file format that toolkit expects.

## Changelog

- [ ] Patch
- [ ] Minor
- [x] Skip


cognite_toolkit/_cdf_tk/resource_classes/containers.py
```diff
@@ -0,0 +1,91 @@
...
+            raise ValueError(
+                f"'{val}' is a reserved container External ID. Reserved External IDs are: {humanize_collection(FORBIDDEN_CONTAINER_EXTERNAL_IDS)}"
+            )
+        return val
+
+    @field_validator("properties")
+    @classmethod
+    def validate_properties_identifier(cls, val: dict[str, str]) -> dict[str, str]:
+        """Validate properties Identifier"""
+        key_pattern = re.compile(CONTAINER_PROPERTIES_IDENTIFIER_PATTERN)
```
doctrino, line 74: Small optimization: Suggest doing this on a module level instead of recompiling this for each container you validate

cognite_toolkit/_cdf_tk/resource_classes/__init__.py
```diff
@@ -9,6 +9,7 @@
 
 from .asset import AssetYAML
 from .base import BaseModelResource, ToolkitResource
+from .containers import ContainerYAML
```
doctrino, line 12: Keep this as is, but I suggest doing this in a separate PR for multiple YAML resources to avoid merge conflicts when you make several of these types of PR in parallel.

cognite_toolkit/_cdf_tk/resource_classes/container_field_definitions.py
```diff
@@ -0,0 +1,361 @@
...
+
+
+_INDEX_DEFINITION_CLASS_BY_TYPE: MappingProxyType[str, type[IndexDefinition]] = MappingProxyType(
+    {c._index_type: c for c in IndexDefinition.__subclasses__()}
+)
+
+
+class PropertyTypeDefinition(BaseModelResource):
+    _property_type: ClassVar[str]
+    type: str
```
doctrino, line 140: Why do you have both a type classvar and an attribute type? Suggest removing `type` and only keep the `_property_type`. Look at how we do it in the `cognite-sdk`: https://github.com/cognitedata/cognite-sdk-python/blob/master/cognite/client/data_classes/data_modeling/data_types.py#L49

cognite_toolkit/_cdf_tk/resource_classes/container_field_definitions.py
```diff
@@ -0,0 +1,361 @@
...
+        description="Specifies that the data type is a list of values.",
+    )
+    max_list_size: int | None = Field(
+        default=None,
+        description="Specifies the maximum number of values in the list",
+    )
+    unit: dict[Literal["externalId", "sourceUnit"], str] | None = Field(
+        default=None,
+        description="The unit of the data stored in this property.",
+    )
```
doctrino, line 200: You can only have unit on `float32` and `float64`, so suggest moving this property down to those subclasses.

cognite_toolkit/_cdf_tk/resource_classes/container_field_definitions.py
```diff
@@ -0,0 +1,361 @@
...
+    _property_type = "text"
+    type: Literal["text"] = "text"
+    list: bool | None = Field(
+        default=None,
+        description="Specifies that the data type is a list of values.",
+    )
+    max_list_size: int | None = Field(
+        default=None,
+        description="Specifies the maximum number of values in the list",
+    )
```
doctrino, line 179: Suggest creating a `ListableProperty` with `list` and `max_list_size` and use that as a base class for all properties that can become listable.

cognite_toolkit/_cdf_tk/resource_classes/container_field_definitions.py
```diff
@@ -0,0 +1,361 @@
...
+
+    @model_serializer(mode="wrap", when_used="always", return_type=dict)
+    def serialize_index(self, handler: SerializerFunctionWrapHandler) -> dict:
+        serialized_data = handler(self)
+        return serialized_data
+
+
+class BtreeIndex(IndexDefinition):
+    _index_type = "btree"
+    index_type: Literal["btree"] = "btree"
```
doctrino, line 121: Same comment here, suggest we only keep the class variable.

cognite_toolkit/_cdf_tk/resource_classes/container_field_definitions.py
```diff
@@ -0,0 +1,361 @@
...
+
+    @model_serializer(mode="wrap", when_used="always", return_type=dict)
+    def serialize_constrain_definition(self, handler: SerializerFunctionWrapHandler) -> dict:
+        serialized_data = handler(self)
+        return serialized_data
+
+
+class UniquenessConstraintDefinition(ConstraintDefinition):
+    _constraint_type = "uniqueness"
+    constraint_type: Literal["uniqueness"] = "uniqueness"
```
doctrino, line 71: Suggest only keep class variable

doctrino (commented):
A few adjustments, and this will be ready.

cognite_toolkit/_cdf_tk/resource_classes/container_field_definitions.py
```diff
@@ -48,7 +47,10 @@ def find_constraint_definition_cls(cls, data: Any, handler: ModelWrapValidatorHa
             raise ValueError(f"Invalid constraint definition data '{type(data)}' expected dict")
 
         if cls is not ConstraintDefinition:
-            return handler(data)
+            data_copy = dict(data)
+            if "constraintType" in data_copy:
+                data_copy.pop("constraintType")
```
doctrino, line 52: Simplified
```suggestion
            data_copy.pop("constraintType", None)
```

doctrino (commented):
Good work I like the solution, just suggest you reduce 
```python
if "type" in data_copy:
    data_copy.pop("type")
```
to a one liner `data_copy.pop("type", None)`.

cognite_toolkit/_cdf_tk/constants.py
```diff
@@ -77,6 +77,49 @@
 HINT_LEAD_TEXT_LEN = 5
 EnvType: TypeAlias = Literal["dev", "test", "staging", "qa", "prod"]
 USE_SENTRY = "pytest" not in sys.modules and os.environ.get("SENTRY_ENABLED", "true").lower() == "true"
+SPACE_FORMAT_PATTERN = r"^[a-zA-Z][a-zA-Z0-9_-]{0,41}[a-zA-Z0-9]?$"
```
larshagencognite, line 80: This space format pattern looks buggy, is this the official definition? I would guess the intention of this pattern is to make sure the string does not end with `_` or `-`, but that is not how it is written.

The container pattern looks sane, as it has the right parentheses

larshagencognite (approved):
🦄

PR #1606, opened 2025-06-04
Use pydantic for transformation validation

doctrino:
# Description

Start using the newly created `TransformationYAML` for validation in the `cdf build` command.

## Changelog

- [x] Patch
- [ ] Minor
- [ ] Skip

### Improved

- In the `cdf build` command improved feedback if transformations configurations is not following the specifications.

## templates

No changes.

### DIFF
--- a/cognite_toolkit/_cdf_tk/loaders/_resource_loaders/transformation_loaders.py
+++ b/cognite_toolkit/_cdf_tk/loaders/_resource_loaders/transformation_loaders.py
@@ -79,6 +79,7 @@ from cognite_toolkit._cdf_tk.exceptions import (
     ToolkitYAMLFormatError,
 )
 from cognite_toolkit._cdf_tk.loaders._base_loaders import ResourceLoader
+from cognite_toolkit._cdf_tk.resource_classes import TransformationYAML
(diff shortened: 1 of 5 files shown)


cognite_toolkit/_cdf_tk/validation.py
```diff
@@ -138,7 +138,7 @@ def validate_resource_yaml_pydantic(
     warning_list: WarningList = WarningList()
     if isinstance(data, dict):
         try:
-            validation_cls.model_validate(data)
+            validation_cls.model_validate(data, strict=True)
```
doctrino, line 141: strict=True is necessary, if not string will be converted to `bool`.

cognite_toolkit/_cdf_tk/loaders/_resource_loaders/transformation_loaders.py
```diff
@@ -111,6 +112,7 @@ class TransformationLoader(
     list_cls = TransformationList
     list_write_cls = TransformationWriteList
     kind = "Transformation"
+    yaml_cls = TransformationYAML
```
doctrino, line 115: Setting this makes the `cdf build` command use this class for validation.

PR #1607, opened 2025-06-04
[CDF-24893, CDF-24894] Support external package libraries

ronpal:
# Description

[alpha] As ongoing efforts to decouple Toolkit from content (Templates, a.k.a Deployment Packs), this introduces the ability to set external "libraries", i.e. url to zip files that conform to a specific format. 

If the user includes a library in their `cdf.toml`, the Toolkit attempts to download and unpack it. If successful, the library contents is presented to the user in the interactive `modules add` and `modules init `commands. 

Screenshot 2025-06-12 at 13 38 46

Caveats: 
* We only support one library at a time for now, but the toml format suggests multiple. This discrepancy is by choice because we don't want to change the toml format later. 
* The context manager pattern is only implemented on ModulesCommand's init and add subcommands. It would be cleaner to follow the same pattern on all of them, that will follow in a later PR.

## Changelog

- [x] Patch
- [ ] Minor
- [ ] Skip

### Added

- [alpha] Support for external libraries

## templates

No changes.


LIBRARIES.md
```diff
@@ -0,0 +1,153 @@
...
+
+The library must be available with https. Authentication is not currently supported.
+
+To publish a library, create a repository with the following structure. It is strictly required for the
+Toolkit to recognize it and present it to the user in the cdf modules init or cdf modules add commands.
+
+```example:
+
+<root folder>
+├── package.toml
```
doctrino, line 24: This is sufficient for us to run the entire `cdf modules init/add` command, and only at the very end start the download process of what the user selects?

cognite_toolkit/build_info.dev.yaml
doctrino, line 1: This should not be committed.

cdf.toml
doctrino, line 1: Can you revert this? There is no reason to touch this file in this PR

cognite_toolkit/_cdf_tk/commands/modules.py
```diff
@@ -280,7 +307,9 @@ def init(
             organization_dir = Path(organization_dir_raw.strip())
 
         modules_root_dir = organization_dir / MODULES
-        packages = Packages().load(self._builtin_modules_path)
+
+        cdf_toml = CDFToml.load()
+        packages = self._get_library_packages(cdf_toml) or Packages.load(self._builtin_modules_path)
```
doctrino, line 312: I think we need an alpha flag here. We should not change the current behavior unless it is set.

cognite_toolkit/_cdf_tk/commands/modules.py
```diff
@@ -679,10 +708,67 @@ def add(self, organization_dir: Path) -> None:
             environments.append(default.environment.validation_type)
             build_env = default.environment.validation_type
 
+        cdf_toml = CDFToml.load()
         existing_module_names = [module.name for module in ModuleResources(organization_dir, build_env).list()]
-        available_packages = Packages().load(self._builtin_modules_path)
-
+        available_packages = self._get_library_packages(cdf_toml) or Packages.load(self._builtin_modules_path)
```
doctrino, line 713: Same here

cognite_toolkit/_cdf_tk/commands/modules.py
```diff
@@ -679,10 +708,67 @@ def add(self, organization_dir: Path) -> None:
             environments.append(default.environment.validation_type)
             build_env = default.environment.validation_type
 
+        cdf_toml = CDFToml.load()
```
doctrino, line 711: The CDFToml is  singleton you can import as well.

cognite_toolkit/_cdf_tk/data_classes/_packages.py
```diff
@@ -66,7 +66,7 @@ def load(
             root_module_dir: The module directories to load the packages from.
         """
 
-        package_definition_path = root_module_dir / "package.toml"
+        package_definition_path = next(root_module_dir.rglob("packages.toml"))
```
doctrino, line 69: Needs an alpha flag I think, or is `packages.toml` only in the builtint?

What happens if you don't find it?

cognite_toolkit/config.dev.yaml
doctrino, line 1: Should not be comitted.

cognite_toolkit/_cdf_tk/commands/modules.py
doctrino: Never mind. This is the pattern we use.

cdf.toml
```diff
@@ -10,10 +10,8 @@ module-repeat = true
 populate = true
 agents = false
 profile = true
-dump-extended = true
 infield = true
 dump-data = true
-migrate = true
```
doctrino, line 16: Lets not remove these

cognite_toolkit/_cdf_tk/commands/modules.py
```diff
@@ -680,9 +709,62 @@ def add(self, organization_dir: Path) -> None:
...
+                try:
+                    print(f"[green]Adding library {library_name}[/]")
+                    output_path = self._temp_download_dir / f"{library_name}.zip"
+                    self._download_and_unpack(library.url, output_path)
+                    return Packages().load(output_path.parent)
+                except Exception as e:
+                    print(f"[red]Failed to add library {library_name}: {e}[/red]")
+                    raise
+            # If no libraries are specified or the flag is not enabled, load the built-in modules
+        return Packages.load(self._builtin_modules_path)
```
doctrino, line 735: Small suggestion, just for readability.
```suggestion
            # If no libraries are specified or the flag is not enabled, load the built-in modules
        else:
            return Packages.load(self._builtin_modules_path)
```

cognite_toolkit/_cdf_tk/commands/modules.py
```diff
@@ -680,9 +709,62 @@ def add(self, organization_dir: Path) -> None:
...
         download_data = self._get_download_data(added_packages)
         self._create(organization_dir, added_packages, environments, "update", download_data)
+
+    def _get_available_packages(self) -> Packages:
+        """
+        Returns a list of available packages, either from the CDF TOML file or from external libraries if the feature flag is enabled.
+        If the feature flag is not enabled and no libraries are specified, it returns the built-in modules.
+        """
+
+        if Flags.EXTERNAL_LIBRARIES.is_enabled() and CDFToml.libraries:
```
doctrino, line 724: Should we raise an error here if we cannot find a library? Or do we have a default hardcoded one?

doctrino (approved):
Just a minor suggestion

cognite_toolkit/_cdf_tk/commands/modules.py
ronpal: I thought about it and figured that if the users intends to use an external library, we shouldn't silently fall back to default. It is better to be explicit.

doctrino (approved):

doctrino (approved):

finnag (changes requested):
Let's discuss the general concept a bit.  Correct me if i'm wrong, but this basically allows you to download a remote zip file, unpack it, and it will be executed in the context of the user on their own machine?

If this is the case, you should either enforce, or at at least enable, checksum validation. So you can specify targets like `eu.gcr.io/cognitedata/prometheuscommunity/stackdriver-exporter@sha256:5e85516e72f0a78b413b0128a5612c6007a4359c0bfa3492c8aa76115d5832f5` 
And then you verify the checksum before unpacking it.

ronpal:
> Let's discuss the general concept a bit. Correct me if i'm wrong, but this basically allows you to download a remote zip file, unpack it, and it will be executed in the context of the user on their own machine?
> 
> If this is the case, you should either enforce, or at at least enable, checksum validation. So you can specify targets like `eu.gcr.io/cognitedata/prometheuscommunity/stackdriver-exporter@sha256:5e85516e72f0a78b413b0128a5612c6007a4359c0bfa3492c8aa76115d5832f5` And then you verify the checksum before unpacking it.

cognite_toolkit/_cdf_tk/commands/modules.py
```diff
@@ -680,9 +710,102 @@ def add(self, organization_dir: Path) -> None:
...
+
+        except requests.exceptions.RequestException as e:
+            raise ToolkitError(f"Error downloading file from {url}: {e}") from e
+
+    def _validate_checksum(self, checksum: str, file_path: Path) -> None:
+        """
+        Compares the checksum of the downloaded file with the expected checksum.
+        """
+
+        if checksum.lower().startswith("sha256:"):
```
doctrino, line 775: I like this

doctrino (approved):
Great stuff

PR #1608, opened 2025-06-05
[CDF-24846] 📶Round 2 profiling command.

doctrino:
# Description

In upgrade work, we need a quick way to get an overview over the asset-centric resource in a project.

## Changelog

- [x] Patch
- [ ] Minor
- [ ] Skip

### Added

- [alpha] The `cdf profile asset-centric` command now also shows relationships and labels. In addition, the number of transformations that targets each resource type is shown in the aggregation table.

## templates

No changes.


cognite_toolkit/_cdf_tk/apps/_profile_app.py
```diff
@@ -24,7 +24,7 @@ def asset_centric(
         verbose: bool = False,
     ) -> None:
         """This command gives an overview over the metadata and labels for each of the asset-centric resources.
-        This shows roughly how much unstructured data there. This can, fox example, be used to determine the
+        This shows roughly how much unstructured data there. This can, for example, be used to determine the
```
ronpal, line 27:
```suggestion
        This shows an approximation of unstructured data count. This can, for example, be used to estimate the
```

tests/test_integration/test_commands/test_profile.py
```diff
@@ -6,9 +6,19 @@ class TestDumpResource:
...
+            "Assets",
+            "Events",
+            "Files",
+            "TimeSeries",
+            "Sequences",
+            "Relationships",
+            "Labels",
+        }
         total_count = sum(item["Count"] for item in results)
         assert total_count > 0
```
einarmo, line 20: Is it possible to have a slightly stronger assertion here? "There is something asset-centric in the project" is a bit thin.

doctrino: Do you have anything in mind? 

The intention of this test is to ensure that the asset-centric profiling runs end-to-end and does not raise any exceptions. Each individual part is tested more accurately as you see in the test below. Thus I prefer to keep it as is.

einarmo:
:unicorn:

PR #1612, opened 2025-06-06
Create toolkit client

doctrino:
# Description

This is a small helper function that is used in internal bug hunting/testing. Switching it to use the toolkit client instead cognite client.

## Changelog

- [ ] Patch
- [ ] Minor
- [x] Skip


thorkildcognite:
risk review ok

PR #1613, opened 2025-06-06
[CDF-24918] ⏱ Toolkit client set-pending-instance-id

doctrino:
# Description

**REVIEWER**: The core part of this PR is ~90 lines, the `.set_instance_ids()` method in the `ExtendedTimeSeriesAPI`. The rest is copy-paste from the CogniteSDK with addding `pending_instance_ids` to the `ExtendedTimeSeries` and same for the `.retrieve`, `.retrieve_multiple` and `.list` methods with setting `alpha` flag and returning `ExtendedTimeSeries`.

In the upgrade from `TimeSeries` to `CogniteTimeSeries`, we need to link the newly created `CogniteTimeSeries` nodes to existing asset-centric timeseries to avoid reingesting all datapoints. This is the first PR in building this tooling that can do that linking. The tooling will depend on an alpha feature in the `timeseries` endpoint which accepts setting a property `pendingInstanceId` that is used as follows.

1. Set `pendindInstanceId` on existing asset-centric timeseries
2. Create the `CogniteTimeSeries` with the node ID set in step 1.
3. The timeseries syncer will then link the newly created `CogniteTimeSeries` node to the existing asset-centri timeseries from step 1, instead of creating a new one.

This PR adds support for this endpoint into the internal `ToolkitClient` which is not exposed to the end-user. This will be used as a basis for a future `cdf migrate timeseries` command.

## Changelog

- [ ] Patch
- [ ] Minor
- [x] Skip

### DIFF
--- a/cognite_toolkit/_cdf_tk/client/api/extended_timeseries.py
+++ b/cognite_toolkit/_cdf_tk/client/api/extended_timeseries.py
@@ -6,7 +6,8 @@ from cognite.client._constants import DEFAULT_LIMIT_READ
 from cognite.client.data_classes.data_modeling import NodeId
 from cognite.client.data_classes.filters import Filter
 from cognite.client.data_classes.time_series import TimeSeriesFilter, TimeSeriesSort
-from cognite.client.utils._auxiliary import exactly_one_is_not_none
+from cognite.client.utils._auxiliary import exactly_one_is_not_none, split_into_chunks, unpack_items_in_payload
+from cognite.client.utils._concurrency import execute_tasks
 from cognite.client.utils._identifier import IdentifierSequence
 from cognite.client.utils._validation import prepare_filter_sort, process_asset_subtree_ids, process_data_set_ids
 from cognite.client.utils.useful_types import SequenceNotStr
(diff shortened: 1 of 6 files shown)


cognite_toolkit/_cdf_tk/client/api/extended_timeseries.py
```diff
@@ -0,0 +1,301 @@
...
+        """
+        identifiers = IdentifierSequence.load(ids=id, external_ids=external_id, instance_ids=instance_id).as_singleton()
+        return self._retrieve_multiple(
+            list_cls=ExtendedTimeSeriesList,
+            resource_cls=ExtendedTimeSeries,
+            identifiers=identifiers,
+            api_subversion="alpha",
+        )
+
+    def retrieve_multiple(
```
doctrino, line 126: This is a simple overwriting of the `.retrieve_multiple` method in the TimeSeriesAPI to return `ExtendedTimeSeries` and use the `alpha` flag.

cognite_toolkit/_cdf_tk/client/data_classes/extended_timeseries.py
```diff
@@ -0,0 +1,103 @@
...
+        unit: str | None = None,
+        unit_external_id: str | None = None,
+        asset_id: int | None = None,
+        is_step: bool | None = None,
+        description: str | None = None,
+        security_categories: Sequence[int] | None = None,
+        data_set_id: int | None = None,
+        created_time: int | None = None,
+        last_updated_time: int | None = None,
+        pending_instance_id: NodeId | None = None,
```
ronpal, line 55: This is missing in docs above

einarmo (commented):
:unicorn: I'm assuming that once this is stable and part of the SDK you will remove the copy-pasted SDK code?

doctrino:
> 🦄 I'm assuming that once this is stable and part of the SDK you will remove the copy-pasted SDK code?

This will never go into the PySDK. This is intended to be here in Toolkit until migration is complete.

PR #1615, opened 2025-06-07
[CDF-24919] 👷 Fix transformations.preview

doctrino:
# Description
The Python-SDK implementation of the `/transformations/run/query` has a few issues:
1. It is retried. By default up to ten times.
2. It has a client timeout typically set to 30 seconds. That means that even though you pass in a `timeout=240` in the request body, the client will timeout before the server timeout.

In the `cdf profile asset-centric` we (mis)use this endpoint to calculate how many unique metadata keys each asset-centric resource have (unfortunately `assets/aggregate` cannot do this as it 'standardizes the metadata keys and thus lose the actual count). In addition, we use it for doing relationships and label count. These calculation are longer running calculations for larger projects. What happens for a large project, as shown below, is that the PySDK tries starts the calculations, waits 30 seconds (client-side timeout), fails it and does a retry. This ends up with 10 attempts each 30 seconds long to do this computations.

This PR overwrites the `transformations.preview(...)` method (the one that implements `/transformations/run/query`), and forces it to not retry and sets the client-side timeout to server-side + 60 seconds to ensure it is the server side that will be the limiting factor. This enables the profiling of larger projects such as shown below. 

image

## Changelog

- [x] Patch
- [ ] Minor
- [ ] Skip

### Fixed

- [alpha] The `cdf profile asset-centric` is much less likely to time-out when calculating metadata key usage.

## templates

No changes.


cognite_toolkit/_cdf_tk/client/api/fixed_transformations.py
```diff
@@ -0,0 +1,124 @@
...
+from cognite.client.utils import _json
+from requests import Response
+
+
+class FixedTransformationsAPI(TransformationsAPI):
+    """This overrides the preview method in the transformations API to allow for longer timeouts and retries."""
+
+    _default_timeout_run_query = 240  # seconds, this is the maximum timeout for running queries in CDF
+
+    def preview(
```
doctrino, line 18: Note that this is well tested through the `tests/test_integration/test_utils/test_cdf.py` tests.

ronpal (dismissed):

ronpal (approved):

einarmo:
:unicorn:

PR #1620, opened 2025-06-08
[prep CDF-24921] 👌 Raw profile endpoint

doctrino:
# Description

**Context**: This is part of developing a `cdf profile raw` command. The motivation for this command is to support migration from assent-centric to data modeling by analyzing how data is flowing from raw into the different cdf resources such as assets, events, and files.

This PR introduced the `raw/profile` endpoint into the internal client used by Toolkit. This is a hidden endpoint that profiles raw tables up to 1 million rows. This is useful for the new `profile raw` command to list number of rows and columns and thus give an impression of the amount of, for example, assets that are coming from a specific endpoint. The limitation  of 1 million rows I plan to handle by just informing the user that a table have either x number of rows, or more than 1 million. For analyzing assets I expects this limitation will not be relevant for 95% of the cases.

## Changelog

- [ ] Patch
- [ ] Minor
- [x] Skip


ronpal (approved):

ronpal (approved):

ronpal (approved):

ronpal (dismissed):

einarmo (commented):
What did you base this PR on? This endpoint doesn't appear to be documented.

cognite_toolkit/_cdf_tk/client/data_classes/raw.py
```diff
@@ -56,3 +67,360 @@ class RawTableList(WriteableCogniteResourceList[RawTable, RawTable]):
...
+            "valueCounts" if camel_case else "value_counts": value_counts,
+            "count": self.count,
+        }
+
+
+@dataclass
+class NumberProfile(CogniteObject):
+    value_range: tuple[float, float]
+    distinct_count: int
+    value_counts: dict[float, int]
```
einarmo, line 139: This is `Map[String, Int]` in the profiler backend, it looks like, are you sure `float` is valid?

doctrino: You are correct, it returns a string and not float.

einarmo:
:unicorn:

ronpal (approved):

PR #1621, opened 2025-06-09
[prep CDF-24921] 🧮 Raw row count

doctrino:
# Description

This is stacked on #1620 and #1615

This adds a function for counting number of rows in a raw table. It will be used as part of the `cdf profile raw` command.

## Changelog

- [ ] Patch
- [ ] Minor
- [x] Skip


ronpal (approved):

rsjr:
As discussed on Slack, I had a quick chat with transformations team about this. Their recommendation is to not use the query endpoint to get raw row count, if possible. If you can check how to implement this using the profile API + limiting the usage, this will help to mitigate risk.

PR #1624, opened 2025-06-10
[prep CDF-24931] 😊 Introduce migrate app

doctrino:
# Description

**Context** We will support multiple migrations through Toolkit. Currently I plan for `timeseries/files/canvas/charts/infield`. This PR sets up the feature flag for migration along with the base command, such that upcoming PRs can stack on top of this one, and thus being independent of each other.

## Changelog

- [ ] Patch
- [ ] Minor
- [x] Skip

### DIFF
--- a/cdf.toml
+++ b/cdf.toml
@@ -12,6 +12,7 @@ agents = false
 profile = true
 infield = true
 dump-data = true
+migrate = true
 
 exit-on-warning = true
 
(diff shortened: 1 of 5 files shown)


ronpal (approved):
Pretty straightforward. I also support using "migrate" although the official name is "upgrade" since upgrade is used elsewhere in the toolkit (pip install --upgrade, cdf modules upgrade)

larshagencognite (approved):
🦄

PR #1626, opened 2025-06-10
[CDF-24932] 📊 Added statistics API

doctrino:
# Description

This adds the `statistics` endpoint to the internal `ToolkitClient`, i.e., not exposed to end-users.

This is a copy of this PR that has already been through risk-review: https://github.com/cognitedata/neat/pull/1163

**Context**: Toolkit will be used in migration of timeseries/files/canvas. All these requires writing nodes/edges to CDF in large amounts. We need the statistics endpoint (even though it is not yet even in alpha) to ensure that Toolkit cannot be used to fill up a CDF projects with nodes and edges, as that can cause data loss from applications such as Infield, Canvas, or other third-party applications. This has already happen in the past with Neat.

## Changelog

- [ ] Patch
- [ ] Minor
- [x] Skip

### DIFF
--- a/cognite_toolkit/_cdf_tk/client/_toolkit_client.py
+++ b/cognite_toolkit/_cdf_tk/client/_toolkit_client.py
@@ -7,10 +7,10 @@ from cognite.client.credentials import CredentialProvider
 
 from .api.agents.agents import AgentsAPI
 from .api.dml import DMLAPI
+from .api.extended_data_modeling import ExtendedDataModelingAPI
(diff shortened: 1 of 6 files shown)


ronpal (approved):

ronpal:
This is going to be invaluable 🙌

einarmo (commented):
:unicorn: again under the assumption that this will at some point be moved to the SDK.

ronpal (approved):

PR #1627, opened 2025-06-10
[CDF-24935] 🚅 Robustify and speed up instance writing.

doctrino:
# Description

This is a reimplementation of the `client.data_modeling.instances.apply(...)` method from the cognite-sdk overcoming the following issues
1. The `cognite-sdk` method is limited to 1 worker, even though the API support 4. This will be used to run migration which can be millions of nodes, then, we need the full power of the what the API supports.
2. The retry logic in simply backs off and tries again. Learning from the upgrade pilots turns out that when writing a lot of nodes, this often leads to exceptions. Instead, the new method backs off, splits the load into two and tries again with two requests.
3. Standardized the signature of the function to simply take instances instead of distinguishing between nodes and edges. 

Test on writing 1 million nodes

image

## Changelog

- [ ] Patch
- [ ] Minor
- [x] Skip


ronpal (approved):

cognite_toolkit/_cdf_tk/client/api/extended_data_modeling.py
```diff
@@ -8,3 +27,170 @@ class ExtendedDataModelingAPI(DataModelingAPI):
...
+
+        retry_tracker.status += 1
+        match result.status_code:
+            case 200 | 201 | 202 | 204:
+                pass
+            case 401:
+                self._raise_no_project_access_error(result)
+            case 429 | 502 | 503 | 504:
+                items = json_payload["items"]
+                if len(items) > 1 and retry_tracker.should_retry(
```
larshagencognite, line 172: should we never retry single items?

cognite_toolkit/_cdf_tk/client/api/extended_data_modeling.py
```diff
@@ -8,3 +27,170 @@ class ExtendedDataModelingAPI(DataModelingAPI):
...
+            str_format_element_fn=str_format_element,
+        )
+        created_resources = summary.joined_results(lambda res: res.json()["items"])
+
+        return InstancesApplyResultList._load(created_resources)
+
+    def _post_with_item_reduction_retry(
+        self,
+        url_path: str,
+        json_payload: dict[str, Any],
```
larshagencognite, line 123: use a proper json typealias here

cognite_toolkit/_cdf_tk/client/api/extended_data_modeling.py
```diff
@@ -8,3 +27,170 @@ class ExtendedDataModelingAPI(DataModelingAPI):
...
+    ) -> Response:
+        _, full_url = self._resolve_url("POST", url_path)
+        headers = self._configure_headers(
+            "application/json",
+            additional_headers=self._config.headers.copy(),
+        )
+        data: str | bytes
+        try:
+            data = _json.dumps(json_payload, allow_nan=False)
+        except ValueError as e:
```
larshagencognite, line 134: Are we sure this is the only thing that can throw `ValueError` here?

doctrino: Yes, `JSONDecodeError` is a subclass of ValueError.

larshagencognite: I guess one risk here is the error message changing, and users installing this lib with a newer version of the dependency from what is tested here.

cognite_toolkit/_cdf_tk/client/api/extended_data_modeling.py
larshagencognite: I think that is fine for the moment, hopefully this can be improved in the sdk at some point

cognite_toolkit/_cdf_tk/client/api/extended_data_modeling.py
```diff
@@ -8,3 +27,168 @@ class ExtendedDataModelingAPI(DataModelingAPI):
...
+            case 200 | 201 | 202 | 204:
+                pass
+            case 401:
+                self._raise_no_project_access_error(result)
+            case 429 | 502 | 503 | 504:
+                items = json_payload["items"]
+                if retry_tracker.should_retry(status_code=result.status_code, is_auto_retryable=True):
+                    # If we get a 429 or 5xx error, we reduce the number of items in the payload and retry.
+                    # This is to avoid overwhelming the API with too many items at once.
+                    half = len(items) // 2
```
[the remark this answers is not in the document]
doctrino: Nice catch.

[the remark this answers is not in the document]
larshagencognite: please add a test for it as well

ronpal (approved):

cognite_toolkit/_cdf_tk/client/api/extended_data_modeling.py
doctrino: That seems like a very small risk that the builtin `json` module stops making `JSONDecodeError` a subclass of `ValueError`.

larshagencognite: but if that does happen, it only means we return a more generic exception

larshagencognite (approved):
🦄

PR #1628, opened 2025-06-11
[CDF-24953] Fix node update

doctrino:
# Description

See changelog below

## Changelog

- [x] Patch
- [ ] Minor
- [ ] Skip

### Fixed

- When running `cdf deploy`, Toolkit no longer overwrites nodes that are existing. Instead they are updated. This means if there is a property that is not in the local configuration but is set in CDF, Toolkit will not overwrite. 

## templates

No changes.


cognite_toolkit/_cdf_tk/loaders/_resource_loaders/datamodel_loaders.py
```diff
@@ -1140,7 +1143,7 @@ def retrieve(self, ids: SequenceNotStr[NodeId]) -> NodeList:
 
     def update(self, items: NodeApplyList) -> NodeApplyResultList:
         result = self.client.data_modeling.instances.apply(
-            nodes=items, auto_create_direct_relations=False, replace=True
+            nodes=items, auto_create_direct_relations=True, replace=False
```
doctrino, line 1146: Fix

ronpal (approved):

PR #1630, opened 2025-06-14
 Hashing issue

doctrino:
# Description

Found this bug in the way we calculate hashes.

## Changelog

- [x] Patch
- [ ] Minor
- [ ] Skip

### Fixed

- Running `cdf deploy` on Windows will no longer give a difference hash than on Unix/Mac.

## templates

No changes.

### DIFF
--- a/tests/test_unit/test_cdf_tk/test_utils/test_hashing.py
+++ b/tests/test_unit/test_cdf_tk/test_utils/test_hashing.py
@@ -1,0 +1,13 @@
+from pathlib import Path
+
+from cognite_toolkit._cdf_tk.utils.hashing import calculate_bytes_or_file_hash, calculate_str_or_file_hash
+
+def test_hash_filepath_equals_windows_line_endings(tmp_path: Path) -> None:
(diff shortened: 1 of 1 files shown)


ronpal (approved):

cognite_toolkit/_cdf_tk/utils/hashing.py
```diff
@@ -46,10 +46,10 @@ def calculate_secure_hash(item: dict[str, Any], shorten: bool = False) -> str:
 
 def calculate_str_or_file_hash(content: str | Path, shorten: bool = False) -> str:
     if isinstance(content, str):
-        byte_content = content.encode("utf-8")
-    else:
-        byte_content = content.read_bytes()
-    return calculate_bytes_or_file_hash(byte_content, shorten)
```
finnag, line 52: You had two semi-overlapping functions,

`calculate_str_or_file_hash` and `calculate_bytes_or_file_hash`,  and they treated files differently?

Why not one function `calculate_hash` that takes str, bytes, or Path as input and does the right thing on all 3?

PR #1636, opened 2025-06-16
[CDF-24999]🐛 Instance space not required

doctrino:
# Description

Transformation instance space is not required for destination instances

image

## Changelog

- [x] Patch
- [ ] Minor
- [ ] Skip

### Fixed

- Toolkit no longer gives a warning if the `instanceSpace` is missing in a transformation configuration with destination set to `instances`

## templates

No changes.


muradsater (approved):

doctrino:
> How do we decide on space for instances if not required?

You can set it in the SQL.
```sql
SELECT 
  "my_space" as space
FROM 
...
```

larshagencognite (approved):
🦄

PR #1637, opened 2025-06-16
[CDF-24773] 🪲Build only identifiers

doctrino:
# Description

The issue solved in this PR is that the `cdf modules pull` command uses the `cdf build` command internally, but fails if a resource is missing a required field. Instead of solving these two bugs individually, I introduce a new example project with all resources with only identifiers and build that project. This should only raise warnings and not errors.

Note this PR does not fix the issue, but prepares for the fix.

## Changelog

- [ ] Patch
- [ ] Minor
- [x] Skip

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -310,3 +310,4 @@ my_*/
 !cognite_toolkit/_builtin_modules/custom/my_module/
 function_local_venvs*/
 qa_pending_id/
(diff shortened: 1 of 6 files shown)


muradsater (approved):

larshagencognite (approved):
🦄

PR #1639, opened 2025-06-17
[CDF-24990] 😵 Location filename must be unique

doctrino:
# Description

## Problem
Locations filter filenames must be unique across modules
image

## Changelog

- [x] Patch
- [ ] Minor
- [ ] Skip

### Fixed

- Location filter filenames must no longer be unique across modules. This bug was introduced in `0.5.24`.

## templates

No changes.


cognite_toolkit/_cdf_tk/builders/_location.py
```diff
@@ -81,24 +79,3 @@ def build(
...
-        It also adds the kind as a suffix.
-
-        """
-        filestem = source_path.stem
-        # Get rid of the local index
-        filestem = INDEX_PATTERN.sub("", filestem)
-
-        # Increment to ensure we do not get duplicate filenames when we flatten the file
-        # structure from the module to the build directory.
-        self.resource_counter = index
```
doctrino, line 96: This was causing the bug. In the `_create_destination_path` method the `self.resource_counter` is the state. It is incremented each time this function is called. Here, it is overwritten to be the the topological order of the locations in the current modules. Thus, when two location filter filenames are the same the last will overwrite the first. In other words, here a local index is used instead of a global one.

The fix is simply to delete this method and replace it with original `_create_destination_path` as this is using a global index. Since this method is called above in the topological order, the dependencies between the filters are still kept.

tests/test_unit/test_cli/test_behavior.py
```diff
@@ -649,7 +683,7 @@ def test_build_deploy_keep_special_characters(
     my_cdf_toml.cdf.file_encoding = encoding
     monkeypatch.setattr(cdf_toml, "_CDF_TOML", my_cdf_toml)
     BuildCommand(silent=True).execute(
-        False, NAUGHTY_PROJECT, build_dir, None, None, False, env_vars_with_client.get_client(), "raise"
+        False, NAUGHTY_PROJECT, build_dir, ["encoding_issue"], None, False, env_vars_with_client.get_client(), "raise"
```
doctrino, line 686: This change is to ensure that this test does not get influenced by other data in the `naughty_project`. I do this by selecting the `encoding_issue` module such that only that is selected.

larshagencognite (approved):
🦄

PR #1640, opened 2025-06-17
[prep CDF-24984] 👷 Refactor migration command

doctrino:
# Description

This is a pure refactoring. 

1. First making the `_migrate.py` module into a package. Motivation is to separate out the data classes from the command.
2. Move the reading csv logic out from the `MigrateTimeseriesCommand` to the `MigrationMappingList` class. This will make it easier to reuse across multiple migration commands.

## Changelog

- [ ] Patch
- [ ] Minor
- [x] Skip

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -310,3 +310,4 @@ my_*/
 !cognite_toolkit/_builtin_modules/custom/my_module/
 function_local_venvs*/
 qa_pending_id/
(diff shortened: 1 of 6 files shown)


muradsater (approved):

larshagencognite (approved):
🦄

PR #1643, opened 2025-06-17
[CDF-24984] 🗺 CogniteMigrate model

doctrino:
# Description

Stacked on #1644, #1645

## Deploys the CogniteMigration
image

This model will be required to run any other migration command. The thinking is that migrating any of the core asset-centric resources `assets`, `events`, `timeseries`, `files` and `sequences` should have an entry in this model. Then, applications such as Charts, Canvas as well as resources such as 3D and annotations can use this mapping in their migration.

## Changelog

- [x] Patch
- [ ] Minor
- [ ] Skip

### Added

- [Alpha] Added the `cdf migrate prepare` command that is used to deploy the `CogniteMigrate` model used to store mappings from asset-centric to data modeling.

## templates

No changes.

### DIFF
--- a/cognite_toolkit/_builtin_modules/inrobot/default.config.yaml
+++ b/cognite_toolkit/_builtin_modules/inrobot/default.config.yaml
@@ -1,3 +1,3 @@
-first_root_asset_external_id: <change_me>
-first_location: <change_me>
+first_root_asset_external_id: my_root_asset
+first_location: my_location
 robot_1_dataset_ext_id: ds_robot_1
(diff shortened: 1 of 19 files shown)


muradsater (dismissed):

doctrino:
> This PR is doing two different things:
> 
> * Adding the CogniteMigrate model and related code to enable it
> * Refactor on deploy.py, changing `execute` to `deploy_build_directory` and splitting `load_resources` into other functions.
> 
> I cannot see the relation between the two changes, they seem pretty isolated / self-contained
> 
> Is it possible for you to split this PR in two?

It is already split, this has just become outdated. Updated now.
 
> Also, regarding this CogniteMigration model. Do you have any internal documentation explaining how this will work when the migration happens? Is this going to create mapping node for every single entry for the asset centric-model resource? Can this impact on the overall instance limit in general?

The plan is to test this out in some pilots, than bring the approach to the architecture forum in the fall.

PR #1647, opened 2025-06-18
[CDF-25019] 🧑‍🔧 Refactor: producer-worker for reuse

doctrino:
# Description

The worker-producer implementation we use for the dump command has hardcoded descriptions that are for downloading, processing and writing to file. I want to reuse this in the migration, were instead of writing to file, I will write back to CDF. This refactoring makes the descriptions of the tasks parameters instead of hardcoded. In addition, I did a few variable renamings as well as providing a doc-string.

## Changelog

- [ ] Patch
- [ ] Minor
- [x] Skip


larshagencognite (approved):
🦄 with a non-blocking suggestion

PR #1648, opened 2025-06-18
⛓Use YAML classes in the validation of assets, events, filemetadata, dataset, and labels

doctrino:
# Description

The YAML classes we use for validation are already made. This PR connects them to the loader class such that they will be used in the `cdf build` command. In addition, I made a set of tests cases for each as well as updated the function that humanize the pydantic errors. 

## Changelog

- [x] Patch
- [ ] Minor
- [ ] Skip

### Improved

- In the `cdf build` command, Toolkit now gives improved warning messages if the syntax in the configuration file is incorrect for `assets`, `events`, `filemetadata`, `dataset`, and `labels`.

## templates

No changes.


larshagencognite (approved):
🦄

PR #1652, opened 2025-06-19
[prep CDF-25031] 🗒Add optional dependency <code>openpyxl</code>

doctrino:
# Description

In the upcoming profiling commands we will support writing the results to a spreadsheet. We need this dependency for that purporse.

## Changelog

- [ ] Patch
- [ ] Minor
- [x] Skip


larshagencognite (approved):
🦄

PR #1655, opened 2025-06-19
[CDF-25032] 😄Extend aggregators

doctrino:
# Description

**REVIEWER** This is a +200 line change, with 350 lines of testing. The test is large due to that I have to setup a full pipeline with RAW tables, transformations to all the resource types such that I can test the aggregations.

Preperation for profiling were we want to aggregate across asset hierarchies, check used datasets and transformations.

image

## Changelog

- [ ] Patch
- [ ] Minor
- [x] Skip


cognite_toolkit/_cdf_tk/utils/aggregators.py
```diff
@@ -33,6 +49,48 @@ def transformation_count(self) -> int:
...
+                ids.append(id_)
+                seen.add(id_)
+            try:
+                int_id = int(id_)  # type: ignore[arg-type]
+            except (ValueError, TypeError):
+                continue
+            if int_id not in seen:
+                ids.append(int_id)
+                seen.add(int_id)
+        return ids
```
ronpal, line 92: Couldn't this be simplified by just appending to the ids list and using set() to make it unique, instead of using 'seen'?

doctrino: I want to maintain the original order. I think that leads to less surprises when using this method.

cognite_toolkit/_cdf_tk/utils/aggregators.py
```diff
@@ -33,6 +49,48 @@ def transformation_count(self) -> int:
...
+                for transformation in chunk:
+                    if SQLParser(transformation.query or "", operation="profiling").is_using_data_set(
+                        data_set_ids, data_set_external_ids
+                    ):
+                        found_transformations.append(transformation)
+        return found_transformations
+
+    @staticmethod
+    def _to_unique_int_list(results: list) -> list[int]:
+        """Converts a list of results to a list of integers, ignoring non-integer values.
```
larshagencognite, line 72: clarify in comment that it also tries to convert to `int`

cognite_toolkit/_cdf_tk/utils/aggregators.py
```diff
@@ -60,8 +118,26 @@ def __init__(self, client: ToolkitClient) -> None:
...
+        return self.client.assets.aggregate_count(filter=self._create_hierarchy_filter(hierarchy, data_set_external_id))
+
+    @classmethod
+    def _create_hierarchy_filter(
+        cls, hierarchy: str | None, data_set_external_id: str | None = None
+    ) -> AssetFilter | None:
+        if hierarchy is None:
+            return None
+        return AssetFilter(
+            asset_subtree_ids=[{"externalId": hierarchy}] if hierarchy else None,
```
larshagencognite, line 131: You have already checked for `None`, so this check is for empty string?

cognite_toolkit/_cdf_tk/utils/sql_parser.py
```diff
@@ -45,6 +45,20 @@ def destination_columns(self) -> list[str]:
...
+        self, data_set_ids: list[int] | None = None, data_set_external_ids: list[str] | None = None
+    ) -> bool:
+        """Check if the SQL query uses any of the provided data set IDs or external IDs."""
+        for data_set_id in data_set_ids or []:
+            if str(data_set_id) in self.query:
+                return True
+        for data_set_external_id in data_set_external_ids or []:
+            if f'dataset_id("{data_set_external_id}")' in self.query:
+                return True
+            elif f"dataset_id('{data_set_external_id}')" in self.query:
```
larshagencognite, line 58: `elif` condition looks like a duplicate

tests/test_integration/test_utils/test_aggregators.py
```diff
@@ -0,0 +1,362 @@
...
+
+        actual_count = aggregator.count(root)
+        assert actual_count == expected_count
+
+        used_data_sets = aggregator.used_data_sets(root)
+        assert used_data_sets == [expected_dataset_external_id]
+
+        transformation_count = aggregator.transformation_count()
+        assert transformation_count >= 1  # We know at least one transformation is writing to the resource type.
+        used_transformations = aggregator.used_transformations(used_data_sets)
```
larshagencognite, line 359: test used_transformations with id as well as external id?

cognite_toolkit/_cdf_tk/utils/aggregators.py
doctrino, line 131:
```suggestion
            asset_subtree_ids=[{"externalId": hierarchy}],
```

cognite_toolkit/_cdf_tk/utils/sql_parser.py
larshagencognite: Right, did not notice

cognite_toolkit/_cdf_tk/utils/aggregators.py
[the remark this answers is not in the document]
doctrino: This is essentially a specialist cleaning method for the output of the aggregator functions, so I prefer to keep it as one method. I extended the docs for it.

cognite_toolkit/_cdf_tk/utils/sql_parser.py
doctrino: I could do a one liner, but that would be even less readable.

tests/test_integration/test_utils/test_aggregators.py
doctrino: I updated the TimeSeries transformation to use the `datasetId` instead of `dataset_id("{{external_id}})`.

ronpal (approved):

cognite_toolkit/_cdf_tk/utils/sql_parser.py
larshagencognite: not suggesting this is great, but could be
```
for quote in ['\'', '\"']:
    if f'dataset_id({quote}{data_set_external_id}{quote})' in self.query:
        return True
```
when it comes to readability, I found it hard to spot the diff, as the current code does not highlight it

cognite_toolkit/_cdf_tk/utils/aggregators.py
```diff
@@ -125,10 +130,10 @@ def count(self, hierarchy: str | None = None, data_set_external_id: str | None =
     def _create_hierarchy_filter(
         cls, hierarchy: str | None, data_set_external_id: str | None = None
     ) -> AssetFilter | None:
-        if hierarchy is None:
+        if hierarchy is None and data_set_external_id is None:
             return None
         return AssetFilter(
-            asset_subtree_ids=[{"externalId": hierarchy}] if hierarchy else None,
+            asset_subtree_ids=[{"externalId": hierarchy}],
```
larshagencognite, line 136: This can be `None` now

cognite_toolkit/_cdf_tk/utils/sql_parser.py
```diff
@@ -50,7 +50,7 @@ def is_using_data_set(
     ) -> bool:
         """Check if the SQL query uses any of the provided data set IDs or external IDs."""
         for data_set_id in data_set_ids or []:
-            if str(data_set_id) in self.query:
```
larshagencognite: how does using a format string help?

cognite_toolkit/_cdf_tk/utils/aggregators.py
```diff
@@ -60,8 +123,26 @@ def __init__(self, client: ToolkitClient) -> None:
     def display_name(self) -> str:
         return "Assets"
 
-    def count(self) -> int:
-        return self.client.assets.aggregate_count()
+    def count(self, hierarchy: str | None = None, data_set_external_id: str | None = None) -> int:
+        return self.client.assets.aggregate_count(filter=self._create_hierarchy_filter(hierarchy, data_set_external_id))
+
+    @classmethod
+    def _create_hierarchy_filter(
```
larshagencognite, line 130: These `_create_hierarchy_filter` functions should be unit tested with the nine options (None, "", "non-empty") x s (None, "", "non-empty")

larshagencognite (approved):
🦄

PR #1658, opened 2025-06-20
[prep CDF-25035] 🕵 Lookup function Ids

doctrino:
# Description

Function schedules does not support `functionExternalId` only `functionId` (it is deprecated). We will continue to support `functionExternalId` in Toolkit as that is necessary to be able to move a schedule seamlessly between CDF projects (dev, staging, prod). This PR is adding functions to the lookup API which is a pattern we use in Toolkit to replace internal IDs with external IDs and vice versa in the `cdf deploy` command.

## Changelog

- [ ] Patch
- [ ] Minor
- [x] Skip


Aashutosh-cognite (approved):

larshagencognite (approved):
🦄

PR #1665, opened 2025-06-23
[CDF-25070] ⛓️‍💥 Unlink timeseries

doctrino:
# Description

This implements the unlink endpoint for timeseries. Currently, this is only deployed in a dev cluster. The unlink endpoint enables you to unlink CogniteTimeSeries from an asset centric timeseries. This is useful in the case when you want to delete the CogniteTimeSeries nodes, but not lose the datapoints stored. 

The expected way this will be used

1. Unlink asset-centric timeseries from CogniteTimeSeries.
2. Delete CogniteTimeSeries.
3. Set pending instance ID on the asset-centric timeseries.
4. Create new CogniteTimeSeries.

This is necessary if you want the asset-centric timeseries to be linked to a different node than what is currently linked to.

## Changelog

- [ ] Patch
- [ ] Minor
- [x] Skip

### DIFF
--- a/cognite_toolkit/_cdf_tk/client/api/extended_timeseries.py
+++ b/cognite_toolkit/_cdf_tk/client/api/extended_timeseries.py
@@ -125,7 +125,7 @@ class ExtendedTimeSeriesAPI(TimeSeriesAPI):
         if id is None and external_id is None:
             return None
         if isinstance(id, int) and isinstance(external_id, str):
-            raise ValueError("Either id or external_id must be provided, or both as a sequence.")
+            raise ValueError("Cannot specify both id and external_id as single values. Use one or the other.")
         is_single = isinstance(id, int) or isinstance(external_id, str)
         identifiers = IdentifierSequence.load(id, external_id)
 
--- a/cognite_toolkit/_cdf_tk/client/api/extended_timeseries.py
+++ b/cognite_toolkit/_cdf_tk/client/api/extended_timeseries.py
@@ -1,6 +1,7 @@
 from collections.abc import Sequence
-from typing import Any, overload
+from typing import Any, cast, overload
 
+from cognite.client import ClientConfig, CogniteClient
 from cognite.client._api.time_series import SortSpec, TimeSeriesAPI
 from cognite.client._constants import DEFAULT_LIMIT_READ
 from cognite.client.data_classes.data_modeling import NodeId
@@ -18,6 +19,11 @@ from cognite_toolkit._cdf_tk.client.data_classes.pending_instances_ids import Pe
 
 class ExtendedTimeSeriesAPI(TimeSeriesAPI):
     """Extended TimeSeriesAPI to include pending ID methods."""
+
+    def __init__(self, config: ClientConfig, api_version: str | None, cognite_client: CogniteClient) -> None:
+        super().__init__(config, api_version, cognite_client)
+        self._PENDING_IDS_LIMIT = 1000
+        self._UNLINK_LIMIT = 1000
 
     @overload
     def set_pending_ids(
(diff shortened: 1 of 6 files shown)


muradsater (dismissed):

cognite_toolkit/_cdf_tk/client/api/extended_timeseries.py
```diff
@@ -90,6 +96,68 @@ def _set_pending_ids(self, identifiers: Sequence[PendingInstanceId]) -> Extended
...
+
+        Args:
+            id (int | Sequence[int] | None): The ID(s) of the time series.
+            external_id (str | SequenceNotStr[str] | None): The external ID(s) of the time series.
+
+        """
+        if id is None and external_id is None:
+            return None
+        if isinstance(id, int) and isinstance(external_id, str):
+            raise ValueError("Either id or external_id must be provided, or both as a sequence.")
```
larshagencognite, line 128: This message is a bit unclear, at least I don't read `either` as strictly not both.

cognite_toolkit/_cdf_tk/client/api/extended_timeseries.py
```diff
@@ -90,6 +96,68 @@ def _set_pending_ids(self, identifiers: Sequence[PendingInstanceId]) -> Extended
...
+        is_single = isinstance(id, int) or isinstance(external_id, str)
```
larshagencognite, line 129: this will say that it is single if id is int and external_id is sequence, no? Is the user restricted to only call the method through one of the overloads?

cognite_toolkit/_cdf_tk/client/api/extended_timeseries.py
```diff
@@ -90,6 +96,68 @@ def _set_pending_ids(self, identifiers: Sequence[PendingInstanceId]) -> Extended
...
+        identifiers = IdentifierSequence.load(id, external_id)
+
+        tasks = [
+            {
+                "url_path": f"{self._RESOURCE_PATH}/unlink-instance-ids",
```
larshagencognite, line 134: tasks should be typed, no?

cognite_toolkit/_cdf_tk/client/api/extended_timeseries.py
doctrino: I don't think I understand your question.

The user is either calls the methods as

```python
# Returning a single timeseries
client.timeseries.unlink(id=123)
# OR
client.timeseries.unlink(external_id="my_id")

# Returning multiple timeseries
client.timeseries.unlink(id=[123, ...], external_id=["my_id", ...])


```

cognite_toolkit/_cdf_tk/client/api/extended_timeseries.py
```diff
@@ -90,6 +96,68 @@ def _set_pending_ids(self, identifiers: Sequence[PendingInstanceId]) -> Extended
...
+        retrieved_items = tasks_summary.joined_results(lambda res: res.json()["items"])
+
+        result = ExtendedTimeSeriesList._load(retrieved_items, cognite_client=self._cognite_client)
+        if is_single:
+            if len(result) == 0:
+                return None
+            if len(result) > 1:
+                raise ValueError("Expected a single time series, but multiple were returned.")
+            return cast(ExtendedTimeSeries, result[0])
+        return cast(ExtendedTimeSeriesList, result)
```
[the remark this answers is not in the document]
doctrino: The `ExtendedTimeSeriesList` is subclassing `TimeSeriesList` which leads the _load method to return `CogniteResourceList[TimeSeries]`. We know that that the return of `_load` is more specific, thus the cast here.

cognite_toolkit/_cdf_tk/client/api/extended_timeseries.py
larshagencognite: Yeah, if the dict is what is expected by SDK it makes sense

ronpal (approved):

larshagencognite (approved):
🦄

PR #1651, opened 2025-06-18
[CDF-24890] Search configuration resource type

Aashutosh-cognite:
# Description

This PR introduces a new resource type `Search Configurations`. Search Configuration do not have APIs in SDK instead these use ADS - a separate set of APIs, same as the Location Filter.

## Changelog

- [ ] Patch
- [ ] Minor
- [x] Skip

### Added

- `Search Configuration` resource type

## templates

No changes.

PR #1666, opened 2025-06-23
Search Configuration API

Aashutosh-cognite:
# Description

Stacked on #1651 

This PR introduces to search configurations API client to utilise list and update APIs for configurations. This will be used by Configurations resources to load search config resources and validate it.

## Changelog

- [ ] Patch
- [ ] Minor
- [x] Skip


cognite_toolkit/_cdf_tk/client/api/search_config.py
```diff
@@ -0,0 +1,60 @@
...
+            chunk_size: The number of configurations to return in each chunk. None will return all configurations.
+
+        Yields:
+            SearchConfig or SearchConfigList
+
+        """
+        return iter(self.list())
+
+    def __iter__(self) -> Iterator[SearchConfig]:
+        return self.__call__()
```
doctrino, line 39: This endpoint does not support pagination, so lets skip this
```suggestion
```

cognite_toolkit/_cdf_tk/client/api/search_config.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def update(self, configuration_update: SearchConfigWrite) -> SearchConfig:
```
doctrino, line 41:
```suggestion
    def upsert(self, configuration_update: SearchConfigWrite) -> SearchConfig:
```

tests/test_integration/test_toolkit_client/test_search_config.py
```diff
@@ -0,0 +1,73 @@
...
+    SearchConfigList,
+    SearchConfigView,
+    SearchConfigWrite,
+)
+
+SEARCH_CONFIG_NAME = "Search Config Name"
+
+
+@pytest.fixture(scope="session")
+def existing_search_config(toolkit_client: ToolkitClient) -> dict:
```
doctrino, line 17: Why not return a `SearchConfig` here?
```suggestion
def existing_search_config(toolkit_client: ToolkitClient) -> SearchConfig:
```

tests/test_integration/test_toolkit_client/test_search_config.py
```diff
@@ -0,0 +1,73 @@
...
+
+    return result
+
+
+class TestSearchConfigAPI:
+    def test_list_search_configurations(self, toolkit_client: ToolkitClient, existing_search_config: dict) -> None:
+        """Test that we can list search configurations."""
+        search_configs = toolkit_client.search_configurations.list()
+        assert isinstance(search_configs, SearchConfigList)
+        assert len(search_configs) == existing_search_config["total_configs"]
```
doctrino, line 49: I think this is already tested in the fixture

doctrino (commented):
A few adjustments. I think we should also ask if there is `delete` endpoint, as supporting this in Toolkit will be difficult without delete.

tests/test_integration/test_toolkit_client/test_search_config.py
Aashutosh-cognite: I also wanted count of already present configs thus was returning in dict. Could use tuple instead. with first element as count and second as SearchConfig. id can be fetched using SearchConfig.id

doctrino (dismissed):

cognite_toolkit/_cdf_tk/client/api/search_config.py
```diff
@@ -0,0 +1,42 @@
...
+from cognite.client._api_client import APIClient
+from cognite.client.config import ClientConfig
+
+from cognite_toolkit._cdf_tk.client.data_classes.search_config import SearchConfig, SearchConfigList, SearchConfigWrite
+
+
+class SearchConfigurationsAPI(APIClient):
+    def __init__(self, config: ClientConfig, api_version: str | None, cognite_client: CogniteClient) -> None:
+        super().__init__(config, api_version, cognite_client)
+        self._api_version = None
```
HMEiding, line 13: Why is this set to `None`?

tests/test_integration/test_toolkit_client/test_search_config.py
```diff
@@ -0,0 +1,55 @@
...
+SEARCH_CONFIG_NAME = "Search Config Name"
+
+
+@pytest.fixture(scope="session")
+def existing_search_config(toolkit_client: ToolkitClient) -> SearchConfig:
+    view = SearchConfigView(external_id="CogniteTimeSeries", space="cdf_cdm")
+    search_config = SearchConfigWrite(view=view, use_as_name=SEARCH_CONFIG_NAME)
+
+    configs = toolkit_client.search_configurations.list()
+    if configs:
```
HMEiding, line 21: Is there a reason to check if this is `None`?

cognite_toolkit/_cdf_tk/client/api/search_config.py
Aashutosh-cognite: for this API we are not using the `APIClient` class `_get_base_url_with_base_path` method, and the API version for this API is hardcoded at line 14.

Aashutosh-cognite: `_api_verison` is also not getting utilised anywhere else, required only for initialising the base class.

HMEiding: It is used in `_get_base_url_with_base_path()` of `APIClient` as you say. If you expand on this class to later, things could break. So isn't it better to use the value provided when instantiating? Is there a particular reason for setting it to `None`?

Aashutosh-cognite: I tried to remove this line and tests were failing, it seems the client is concatinating the _RESOURCE_PATH with the ordinary /api/ path. Same fix is done in location filter API. There it is mentioned that this is a temporary fix.

doctrino (dismissed):

cognite_toolkit/_cdf_tk/client/api/search_config.py
doctrino: A bit more context here. Toolkit builds on the cognite-sdk. However, we often have early support for new features which has not yet gotten a stable API. The location filter and search configuration API are examples of this. We keep close communication with those team so we can keep up if there are changes. 

Currently, these APIs are 'hidden' and have a different URL path than normal endpoints. We do this 'trick' to enable us reusing the cognite-sdk as much as possible, without implementing the handling of requests from scratch. The ideal solution here is to do a larger change in the cognite-sdk, but that is out of the scope for this particular PR.

This is temporary until those APIs are supported in the `cognite-sdk`.

HMEiding: If what you want is a different behavior of `_get_base_url_with_base_path()` isn't it better to override that directly?

doctrino (approved):

cognite_toolkit/_cdf_tk/client/api/search_config.py
Aashutosh-cognite: We have overrriden `_get_base_url_with_base_path()` method in `SearchConfigurationAPI` class.

PR #1684, opened 2025-06-28
[prep CDF-24982] 🚅 Implemented fast delete

doctrino:
# Description

This is a reimplementation of the `client.data_modeling.instances.delete(...)` method from the cognite-sdk overcoming the following issues

1. The `cognite-sdk` method is limited to 1 worker, even though the API support 2. This will be used in the purge command that potentially deletes million of nodes. It will also be used in migration on cleanups. Then, we want the full power of the API.
2. Standardized the signature of the function to simply take instance IDs instead of distinguishing between nodes and edges. 

## Changelog

- [ ] Patch
- [ ] Minor
- [x] Skip


ronpal (dismissed):

ronpal (approved):

rsjr (commented):
Code implementation is OK, but I have some questions regarding the feature itself:

- When the user runs toolkit purge, the default will be the "fast" mode? I am asking you this because maybe for this use case, this should be a deliberate user decision, as this fills up the deletion limits. 
- Did you run some manual tests with a bigger data loads to see how better this, compared to single "thread" mode?
- The aim of this feature is it only migraiton users?
- This seems something that should live in the Python SDK, no?

Also, a very nitpick of mine: "delete_fast" and "apply_fast" seem a bit vague in terms of naming... Should be "concurrent" or something like it.

doctrino:
> Code implementation is OK, but I have some questions regarding the feature itself:
> 
> * When the user runs toolkit purge, the default will be the "fast" mode? I am asking you this because maybe for this use case, this should be a deliberate user decision, as this fills up the deletion limits.

I will do that

> * Did you run some manual tests with a bigger data loads to see how better this, compared to single "thread" mode?

No yet, main motivation for this one is the standardize of the inteferface.

> * The aim of this feature is it only migraiton users? No, all users.
> * This seems something that should live in the Python SDK, no? No, I don't think Erlend want that as it is easy to misuse. 
> 
> Also, a very nitpick of mine: "delete_fast" and "apply_fast" seem a bit vague in terms of naming... Should be "concurrent" or something like it.

rsjr (commented):
risk review ok

PR #1687, opened 2025-06-29
🦾Fix helper

doctrino:
# Description

The `get_toolkit_client` is a helper function not part of the package. It is used to debug. 

This fixes a deviation on this helper function vs how Toolkit generally the token login flow to name the variables.

## Changelog

- [ ] Patch
- [ ] Minor
- [x] Skip


ronpal (approved):
Nice catch

PR #1693, opened 2025-06-30
🐛Unwrap bug

doctrino:
# Description

This helper function is used in the error message if the request fails. The current implementation assumes that all instances are of the same type, which is incorrect, while the fix just use the identifier.

## Changelog

- [ ] Patch
- [ ] Minor
- [x] Skip

### DIFF
--- a/cognite_toolkit/_cdf_tk/client/api/extended_data_modeling.py
+++ b/cognite_toolkit/_cdf_tk/client/api/extended_data_modeling.py
@@ -9,7 +9,7 @@ from cognite.client._api.data_modeling.instances import InstancesAPI
 from cognite.client._api_client import T
 from cognite.client._cognite_client import ClientConfig, CogniteClient
 from cognite.client._http_client import HTTPClient, HTTPClientConfig, get_global_requests_session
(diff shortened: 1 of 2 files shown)


cognite_toolkit/_cdf_tk/client/api/extended_data_modeling.py
```diff
@@ -97,9 +97,15 @@ def apply_fast(
...
+        def unwrap_element(el: T) -> InstanceApply | NodeId | EdgeId | T:
             if isinstance(el, dict):
-                return type(items[0])._load(el, cognite_client=self._cognite_client)
+                instance_type = el.get("instanceType")
+                if instance_type == "node":
+                    return NodeId.load(el)
+                elif instance_type == "edge":
+                    return EdgeId.load(el)
+                else:
+                    return InstanceApply._load(el)
```
doctrino, line 108: This is a fallback in case a new type of instance will be created. Not in practice it is not possible to hit this without a complaint by `mypy`.

larshagencognite (approved):
🦄

PR #1706, opened 2025-07-02
[CDF-25148] Migration model round 2

doctrino:
# Description

## Changelog

- [x] Patch
- [ ] Minor
- [ ] Skip

### Changed

- [alpha] The model deployed with `cdf migrate prepare` has now changed. 

## templates

No changes.

PR #1708, opened 2025-07-02
Temporarily skip test

doctrino:
# Description

I am changing the migration model we are using. This causes the integration test that deploys it to fail as it is now in different versions depending on which branch you run the test from. Temporarily disabling it.

Ref: https://github.com/cognitedata/toolkit/pull/1706

## Changelog

- [ ] Patch
- [ ] Minor
- [x] Skip


thorkildcognite:
risk review ok

PR #1713, opened 2025-07-03
External libraries hardening

ronpal:
# Description

In some cases, the downloaded zip file was stale. The context manager should clean up on exit, but sometimes fails when process is aborted

## Changelog

- [x] Patch
- [ ] Minor
- [ ] Skip

### Fixed

- Stale deployment packs zips removed

## templates

No changes.


thorkildcognite:
risk review ok

PR #1714, opened 2025-07-04
[CDF-25165] ⛓Fix 404 links

doctrino:
# Description

We have a tests that picks up broken links, but that test was broken so it did not run or report back. Fixed it.

## Changelog

- [x] Patch
- [ ] Minor
- [ ] Skip

### Fixed

- Warnings that contains the URL to to auth and config docs no longer gives 404.

## templates

No changes.

### DIFF
--- a/tests/test_unit/test_cdf_tk/tests_constants.py
+++ b/tests/test_unit/test_cdf_tk/tests_constants.py
@@ -6,7 +6,7 @@ from cognite_toolkit._cdf_tk.constants import URL
 
 @pytest.mark.parametrize(
     "url, name",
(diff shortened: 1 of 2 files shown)


tests/test_unit/test_cdf_tk/tests_constants.py
```diff
@@ -6,7 +6,7 @@
 
 @pytest.mark.parametrize(
     "url, name",
-    [pytest.param(url, id=name) for name, url in vars(URL).items() if not name.startswith("_")],
+    [pytest.param(url, name, id=name) for name, url in vars(URL).items() if not name.startswith("_")],
```
doctrino, line 9: This is the bug in the test.

ronpal (approved):

PR #1717, opened 2025-07-04
[CDF-25018] ➡ cdf migrate files

doctrino:
# Description

**Context**: This is part of the alpha migration tooling in Toolkit, for overview [see here](https://cognitedata.atlassian.net/wiki/spaces/~170890702/pages/5455773697/Migration+Tooling+in+Toolkit). This is step 3 of that process, enabling linking of asset-centric files to cognite-files, such that the new CogniteFile can point to an existing asset-centric file and thus not needing to download and reupload the file content.

**CAVEAT** This functionality is only available in the CDF dev clusters. However, in agreement with @thorkildcognite, we do release it now due to staffing availability. If it is run on a prod cluster you will just get `Failed to set pending IDs for Files: The feature set-pending-instance-ids not enabled in this environment. | code: 400` 

This PR adds the command for migrating files, meaning linking them to the existing asset-centric FileMetadata.

## Changelog

- [x] Patch
- [ ] Minor
- [ ] Skip

### Added

- [alpha] New command `cdf migrate files`. 

## templates

No changes.


Aashutosh-cognite (approved):
LGTM

arnels (approved):
🦄

PR #1726, opened 2025-07-05
[CDF-25147]🐛 Bug in row count

doctrino:
# Description

Fixes bug in SQL causing it to ignore the row count. Note this has not yet been released to end-users, thus no need for a release. 

## Changelog

- [ ] Patch
- [ ] Minor
- [x] Skip


Aashutosh-cognite (approved):
LGTM

PR #1736, opened 2025-07-08
[CDF-25200]🧹Cleanup from future import __annotations__

doctrino:
# Description

We have switched our code base a while back from having lower bound on `3.8` to now having lower bound on `3.10`. This makes `__from__ future import annotations` no longer necessary. This PR removes it. A benefit of this is that AI tools gets (hopefully) less confused in the code base as it no longer sees 3.9 and earlier syntax for Python.

## Changelog

- [ ] Patch
- [ ] Minor
- [x] Skip

### DIFF
--- a/cognite_toolkit/_cdf_tk/cdf_toml.py
+++ b/cognite_toolkit/_cdf_tk/cdf_toml.py
@@ -1,5 +1,3 @@
-from __future__ import annotations
-
 import re
 import sys
 import urllib
@@ -20,11 +18,14 @@ from cognite_toolkit._cdf_tk.exceptions import (
 from cognite_toolkit._cdf_tk.tk_warnings import MediumSeverityWarning
 
 if sys.version_info >= (3, 11):
+    from typing import Self
+
     import tomllib
     from tomllib import TOMLDecodeError
 else:
     import tomli as tomllib
     from tomli import TOMLDecodeError
+    from typing_extensions import Self
 
 
 @dataclass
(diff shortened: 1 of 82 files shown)


cognite_toolkit/_cdf_tk/cdf_toml.py
```diff
@@ -114,7 +115,7 @@ class CDFToml:
     is_loaded_from_file: bool = False
 
     @classmethod
-    def load(cls, cwd: Path | None = None, use_singleton: bool = True) -> CDFToml:
+    def load(cls, cwd: Path | None = None, use_singleton: bool = True) -> "CDFToml":
```
Aashutosh-cognite, line 118: Just a thought, in `Library` data class we are specifying `Self` as return type where as here in `CDFToml` `"CDFToml"`. Shouldn't we have a consistent return type style guide? like either `Self` as return type or `"<class name>"`?

Aashutosh-cognite (commented):
LGTM. Have a look at the comment though it is not important for now.

Aashutosh-cognite (approved):

cognite_toolkit/_cdf_tk/cdf_toml.py
doctrino: This return the `_CDF_TOML` global which is fixed to be of `CDFToml` thus it will always return `CDFToml`. The ones that returns `Self` will return self meaning the class itself or any derived class. This is not the case for this `load` method.

psalaberria002:
:unicorn:

PR #1744, opened 2025-07-12
[CDF-25246] 🐛Bug aggregate

doctrino:
# Description

Bug in how we handled empty list in the aggregate method influencing the command below.

## Changelog

- [x] Patch
- [ ] Minor
- [ ] Skip

### Fixed

- The `cdf dump assets/timeseries` no longer fails with `CogniteAPIError: Request had 1 constraint violations` if you use interactive selictions.

## templates

No changes.


rsjr (commented):
risk review ok

PR #1756, opened 2025-07-14
[CDF-25252] 😋 SearchConfigYAML

doctrino:
# Description

**Context** In Toolkit, we implement pydantic for all DTO (data transfer objects) in the API. This is instead of using the DTO classes from the `cognite-sdk`. We do this for two reasons. The Toolkit DTO objects deviates from the API in some small ways, for example, we do not have internal IDs, instead we use external ID such that a resource can be deployed across multiple CDF projects. In addition, `pydantic` classes gives much better feedback to the user if they are not following the syntax of the YAMl.

This PR implements the `SearchConfigYAML` for the search config API.

**Note** I found (or Gemini) a bug/confusing error in `validate_resource_yaml_pydantic` when running tests for `SearchConfigYAMl`. Fixed it and updated 

## Changelog

- [x] Patch
- [ ] Minor
- [ ] Skip

### Improved

- Improved location information for `ResourceFormatWarning`s.

## templates

No changes.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -2,7 +2,7 @@
 repos:
   - repo: https://github.com/astral-sh/uv-pre-commit
     # uv version.
-    rev: 0.7.20
+    rev: 0.7.21
     hooks:
       - id: uv-lock
   - repo: https://github.com/astral-sh/ruff-pre-commit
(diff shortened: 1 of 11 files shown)


psalaberria002 (approved):
:unicorn:

PR #1758, opened 2025-07-14
[CDF-25199] 🐛Unchanged workflow version redeploys

doctrino:
# Description

See below

## Changelog

- [x] Patch
- [ ] Minor
- [ ] Skip

### Fixed

- When running `cdf deployd` workflow versions with a transformation tasks without defaults defined is no longer redeployed.

## templates

No changes.


Aashutosh-cognite (approved):

rsjr (commented):
risk review ok
I know it's a small fix, but it would help next time to have more info in the PR description. I have bit a context on toolkit, but others might not :)

PR #1769, opened 2025-07-19
[CDF-25286] 🦾 Improve producer worker

doctrino:
# Description

**Context**: The `ProducerWorker` is an internal structure that implements a producer/worker pattern for doing tasks concurrently. 

In addition to the entry below, found a bug in the `ProduceWorker` that did not handle if the iter call fails. This does not impact the end-user as the `ProducerWorker` always has a valid iterable when called from the CLI commands. 

## Changelog

- [x] Patch
- [ ] Minor
- [ ] Skip

### Improved

- When running `cdf dump asset/timeseries`, Toolkit have fixed the the process bar descriptions.

## templates

No changes.


cognite_toolkit/_cdf_tk/utils/producer_worker.py
```diff
@@ -97,7 +97,7 @@ def run(self) -> None:
             process_task = progress.add_task(self.process_description.title(), total=self.iteration_count)
             process_thread = threading.Thread(target=self._process_worker, args=(progress, process_task))
 
-            write_task = progress.add_task(self.download_description.title(), total=self.iteration_count)
+            write_task = progress.add_task(self.write_description.title(), total=self.iteration_count)
```
doctrino, line 100: This was the bug, used the download description twice instead of the write description.

Aashutosh-cognite (approved):
LGTM

paco-cognite:
risk review ok

PR #1780, opened 2025-07-23
[CDF-25285] 🦿Convert asset-centric properties to container properties Part 3

doctrino:
# Description

This method will be used to convert the asset-centric properties/metadata to the appropriate container property. 

This is part 1 of 3 parts. I am splitting this up to try to limit the size. 
* Part 1: Conversion of non-list values.
* Part 2: Conversion of list-values
* Part 3 (this PR): Special cases from asset-centric to data modeling. For example, going from timeseries to `CogniteTimeSeries` the property `isString` must convert from a boolean to an enum.

## Changelog

- [ ] Patch
- [ ] Minor
- [x] Skip

### DIFF
--- a/cognite_toolkit/_cdf_tk/utils/dtype_conversion.py
+++ b/cognite_toolkit/_cdf_tk/utils/dtype_conversion.py
@@ -2,8 +2,10 @@ import ctypes
 import json
 from abc import ABC, abstractmethod
 from collections.abc import Mapping
-from typing import ClassVar, cast
+from typing import ClassVar, Literal, cast
 
+from cognite.client.data_classes import Label, LabelDefinition
+from cognite.client.data_classes.data_modeling import ContainerId
 from cognite.client.data_classes.data_modeling.data_types import Enum, ListablePropertyType
 from cognite.client.data_classes.data_modeling.instances import PropertyValueWrite
 from cognite.client.data_classes.data_modeling.views import PropertyType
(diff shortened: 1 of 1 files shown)


cognite_toolkit/_cdf_tk/utils/dtype_conversion.py
```diff
@@ -300,3 +367,20 @@ def _convert(self, value: str | int | float | bool | dict) -> PropertyValueWrite
...
+SPECIAL_CONVERTER_BY_SOURCE_DESTINATION: Mapping[
+    tuple[tuple[AssetCentric, str], tuple[ContainerId, str]],
+    type[_SpecialCaseConverter],
+] = {}
+_to_check = [_SpecialCaseConverter]
+while _to_check:
+    _cls_ = _to_check.pop()
+    for _subclass in _cls_.__subclasses__():
+        if ABC in _subclass.__bases__:
+            _to_check.append(_subclass)
```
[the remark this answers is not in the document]
doctrino: That will fail for the case we have.

ronpal (approved):

Aashutosh-cognite (dismissed):

cognite_toolkit/_cdf_tk/utils/dtype_conversion.py
```diff
@@ -299,3 +366,20 @@ def _convert(self, value: str | int | float | bool | dict) -> PropertyValueWrite
...
+        else:
+            # We know that this is a mutable mapping, but we do not want to expose that outside this module.
+            SPECIAL_CONVERTER_BY_SOURCE_DESTINATION[  # type: ignore[index]
+                (_subclass.source_property, _subclass.destination_container_property)
+            ] = _subclass
+# Cleanup
+del _to_check, _cls_, _subclass
```
rsjr, line 385: This whole block here would benefit with a some comments/docstring on what is doing. Took me a while to understand what is this, so it might be helpful for others. 

Also, I suggest this whole thing to be moved to a function

```
SPECIAL_CONVERTER_BY_SOURCE_DESTINATION = {}

def populate_converter_registry():
    """
    Scans for all concrete subclasses of _SpecialCaseConverter
    and populates the global registry.
    """
    # Clear the registry to ensure a clean slate on each run
    SPECIAL_CONVERTER_BY_SOURCE_DESTINATION.clear()
    
    _to_check = [_SpecialCaseConverter]
    while _to_check:
        _cls_ = _to_check.pop()
        for _subclass in _cls_.__subclasses__():
            if ABC in _subclass.__bases__:
                _to_check.append(_subclass)
            else:
                key = (_subclass.source_property, _subclass.destination_container_property)
                SPECIAL_CONVERTER_BY_SOURCE_DESTINATION[key] = _subclass

# You can call this function once when your application starts
populate_converter_registry()
```

I don't know what you think, but for me looks more self contained and easier to debug/test if you need to.

Also, if you move this to a function, you will be able to write some unit/mocktests for this. I see that it's doing what is supposed to, but without tests is hard to tell. Can you add some :) ?

doctrino: This has actually been in review for so long that I have implemented this type of function already.

rsjr (commented):
risk review ok

ronpal (approved):

PR #1798, opened 2025-07-28
[CDF-25339] 🗜 FileIO compression 

doctrino:
# Description

This PR introduces the compression logic.

## Changelog

- [ ] Patch
- [ ] Minor
- [x] Skip


Aashutosh-cognite (dismissed):
LGTM

cognite_toolkit/_cdf_tk/utils/fileio/_compression.py
```diff
@@ -0,0 +1,78 @@
...
+    # `Mapping[str, type[Compression]]`, while we know they are actually dictionaries.
+    COMPRESSION_BY_NAME[subclass.name] = subclass  # type: ignore[index]
+    if not subclass.file_suffix:
+        continue
+    if subclass.file_suffix in COMPRESSION_BY_SUFFIX:
+        raise TypeError(
+            f"Duplicate compression file suffix {subclass.file_suffix!r} found for classes "
+            f"{COMPRESSION_BY_SUFFIX[subclass.file_suffix].__name__!r} and {subclass.__name__!r}."
+        )
+    COMPRESSION_BY_SUFFIX[subclass.file_suffix] = subclass  # type: ignore[index]
```
sighol, line 78: Do you expect there to be a lot of these Compression classes?
It seems a bit over-engineered to use runtime reflection to find subclasses of `Compression` when the list of possible compression methods is probably small?

`__subclasses__` will only find subclasses that have been loaded into memory, so there is still a chance that some classes will be missed if someone refactors and moves some compression classes to another file.

It seems more robust to just use a list
```suggestion
_all_compression_types: list[type[Compression]] = [NoneCompression, GzipCompression]
COMPRESSION_BY_SUFFIX: Mapping[str, type[Compression]] = {x.file_suffix: x for x in _all_compression_types}
COMPRESSION_BY_NAME: Mapping[str, type[Compression]] = {x.name: x for x in _all_compression_types}
assert len(_all_compression_types) == len(COMPRESSION_BY_SUFFIX), "Duplicated compression suffix"
assert len(_all_compression_types) == len(COMPRESSION_BY_NAME), "Duplicated compression name"
```

What do you think? 🤔

cognite_toolkit/_cdf_tk/utils/fileio/_compression.py
```diff
@@ -0,0 +1,78 @@
...
+        self.filepath = filepath
+
+    @abstractmethod
+    def open(self, mode: Literal["r", "w"]) -> TextIOWrapper:
+        """Open the compressed file and return a file-like object."""
+        raise NotImplementedError("This method should be implemented in subclasses.")
+
+    @classmethod
+    def from_filepath(cls, filepath: Path) -> "Compression":
+        return COMPRESSION_BY_SUFFIX.get(filepath.suffix, NoneCompression)(filepath=filepath)
```
sighol, line 29: I wonder if it would be safer to raise an error if the suffix is unknown instead of falling back to None.
If someone with a new version of toolkit compresses with a new format, and someone with an old toolkit version tries to decrompress it, they will probably get some weird errors.

sighol (commented):
Overall I think this looks good, but I have a couple of questions.

cognite_toolkit/_cdf_tk/utils/fileio/_compression.py
doctrino: I do expect a couple more. I typically will have it all in the same file, or just move the subclass call. However it doesn't hurt with a test

sighol: But is it more readable and robust to hard code a list of compression classes?

cognite_toolkit/_cdf_tk/utils/fileio/_compression.py
doctrino: But how do you deal with non-compressed files? For uncompressed files, you will just get an error if you do what you suggest. 

And you will not get a wired error message, just that the file format x is not supported

cognite_toolkit/_cdf_tk/utils/fileio/_compression.py
sighol: I am not suggesting an implementation without classes, but one without reflection. Like the example above.

doctrino: I like reflection, it is a pattern  we use a lot in Toolkit. I like that we automatically register all classes. I have used the hardcoded variant before and experienced that it gets outdated.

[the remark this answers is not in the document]
doctrino: I do not think that is cleaner.

thorkildcognite:
risk review ok

sighol:
@doctrino I think there is a subtle difference between testing that all registered compression classes are ok (which is what is tested now), and explicitely testing that we support eg. gzip compression.

I think it would be better if your last test actually listed the compression classes instead of using the same reflection method that you used in the code. So if something causes any of the compression classes to no longer be registered, then at least a test will fail.

Up to you.

Aashutosh-cognite (approved):

PR #1799, opened 2025-07-28
[CDF-25329] 🧑‍💻 FileIO Read base class 

doctrino:
# Description

This PR introduces the base class for reading.

## Changelog

- [ ] Patch
- [ ] Minor
- [x] Skip

### DIFF
--- a/cognite_toolkit/_cdf_tk/client/api/extended_files.py
+++ b/cognite_toolkit/_cdf_tk/client/api/extended_files.py
@@ -159,3 +159,69 @@ class ExtendedFileMetadataAPI(FilesAPI):
             ignore_unknown_ids=ignore_unknown_ids,
             api_subversion="alpha",
         )
+
+    @overload
+    def unlink_instance_ids(
+        self,
+        id: int | None = None,
+        external_id: str | None = None,
+    ) -> ExtendedFileMetadata | None: ...
+
+    @overload
+    def unlink_instance_ids(
+        self,
+        id: Sequence[int] | None = None,
+        external_id: SequenceNotStr[str] | None = None,
+    ) -> ExtendedFileMetadataList: ...
+
+    def unlink_instance_ids(
+        self,
+        id: int | Sequence[int] | None = None,
+        external_id: str | SequenceNotStr[str] | None = None,
+    ) -> ExtendedFileMetadata | ExtendedFileMetadataList | None:
+        """Unlink instance IDs from files.
+
+        This allows a CogniteFile node in Data Modeling to be deleted without deleting the underlying file content.
+
+        Args:
+            id (int | Sequence[int] | None): The ID(s) of the files.
+            external_id (str | SequenceNotStr[str] | None): The external ID(s) of the files.
+
+        Returns:
+            ExtendedFileMetadata | ExtendedFileMetadataList | None: The updated file metadata object(s). For single item requests, returns `None` if the file is not found.
+        """
+        if id is None and external_id is None:
+            raise ValueError("At least one of id or external_id must be provided.")
+        if isinstance(id, int) and isinstance(external_id, str):
+            raise ValueError("Cannot specify both id and external_id as single values. Use one or the other.")
+        is_list = isinstance(id, Sequence) or (isinstance(external_id, Sequence) and not isinstance(external_id, str))
+        identifiers = IdentifierSequence.load(id, external_id)
+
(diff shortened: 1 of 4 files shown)


cognite_toolkit/_cdf_tk/utils/fileio/_readers.py
```diff
@@ -0,0 +1,45 @@
...
+
+from ._base import FileIO
+from ._compression import COMPRESSION_BY_SUFFIX, Compression
+
+
+class FileReader(FileIO, ABC):
+    def __init__(self, input_file: Path) -> None:
+        self.input_file = input_file
+
+    def read_chunks(self) -> Iterator[JsonVal]:
```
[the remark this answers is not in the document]
doctrino: This is intentional. It is up to the user of the reader to convert date and dateime object.

cognite_toolkit/_cdf_tk/utils/fileio/_readers.py
```diff
@@ -0,0 +1,55 @@
...
+        with compression.open("r") as file:
+            yield from self._read_chunks_from_file(file)
+
+    @abstractmethod
+    def _read_chunks_from_file(self, file: TextIOWrapper) -> Iterator[JsonVal]:
+        """Read chunks from the file."""
+        raise NotImplementedError("This method should be implemented in subclasses.")
+
+    @classmethod
+    def from_filepath(cls, filepath: Path) -> "FileReader":
```
arnels, line 30: Can you add some tests for this method?

cognite_toolkit/_cdf_tk/utils/fileio/_readers.py
```diff
@@ -0,0 +1,55 @@
...
+
+        if suffix in FILE_READ_CLS_BY_FORMAT:
+            return FILE_READ_CLS_BY_FORMAT[suffix](input_file=filepath)
+        raise ToolkitValueError(
+            f"Unknown file format: {suffix}. Available formats: {humanize_collection(FILE_READ_CLS_BY_FORMAT.keys())}."
+        )
+
+
+FILE_READ_CLS_BY_FORMAT: Mapping[str, type[FileReader]] = {}
+for subclass in get_concrete_subclasses(FileReader):  # type: ignore[type-abstract]
```
arnels, line 48: Can you add some tests for this code path?

doctrino: This is part of PRs to introduce new functionality. As this is just a base class, it does not make sense to introduce tests for this. For example, when I add the records read/write classes I have this tests that covers this part of the code: https://github.com/cognitedata/toolkit/pull/1802/files#diff-8fca510d461329e4b2707e116ae1146f4dcb77a6536acf852abc92c290c71531R226

cognite_toolkit/_cdf_tk/utils/fileio/_readers.py
doctrino: Similarly to the from format, the same test also uses this when I introduce the concrete classes: https://github.com/cognitedata/toolkit/pull/1802/files#diff-8fca510d461329e4b2707e116ae1146f4dcb77a6536acf852abc92c290c71531R189

cognite_toolkit/_cdf_tk/utils/fileio/_readers.py
arnels: I understand your reasoning, but I'd prefer that we add dedicated tests for the base class itself.

While the code will be covered by the subclass tests, a base class should not depend on its children for testing. Relying on subclass tests makes our test suite more brittle. For example, if those subclasses are changed or removed in the future, we could silently lose test coverage for this core logic.

cognite_toolkit/_cdf_tk/utils/fileio/_readers.py
arnels: Same applies here: https://github.com/cognitedata/toolkit/pull/1799#discussion_r2253680779

cognite_toolkit/_cdf_tk/utils/fileio/_readers.py
doctrino: Ok, added a few tests.

Aashutosh-cognite (approved):

Aashutosh-cognite (approved):

PR #1800, opened 2025-07-28
[CDF-25329] ✍ Write base class 

doctrino:
# Description

This PR introduces the base class for writing

## Changelog

- [ ] Patch
- [ ] Minor
- [x] Skip


Aashutosh-cognite (dismissed):

cognite_toolkit/_cdf_tk/utils/fileio/_writers.py
```diff
@@ -0,0 +1,119 @@
...
+        raise NotImplementedError("This method should be implemented in subclasses.")
+
+    @classmethod
+    def create_from_format(
+        cls,
+        format: str,
+        output_dir: Path,
+        kind: str,
+        compression: type[Compression],
+    ) -> "FileWriter":
```
thorkildcognite, line 101: Nit: I tried, and failed, to find a different way to indicate this type (since Self can't be parameterized and the parameterization is runtime, it all gets complex).  The way you did it is a way, but shouldn't the type annotation here be quoted in single quotes, and not double quotes? It is claimed that PEP-484 specifies that, but it doesn't really spell it out.

thorkildcognite:
:unicorn:

cognite_toolkit/_cdf_tk/utils/fileio/_writers.py
doctrino: As far as I understand single and double quotes can be used interchangeably in Python.

PR #1801, opened 2025-07-29
[CDF-25341] 🐛 Fix sanitize sheet title

doctrino:
# Description

 User Error
image

Cognite Toolkit has several profiling commands to give an overview over the data ingestion pipelines (raw -> transformation -> asset-centric).

These commands support writing the output to a spreadsheet, were the sheet name is typically a asset hierarchy ID or similar. This PR fixes the issue if there are characters that are used in the externalId but which are invalid in a spreadsheet by introducing a sanitize function. 

## Changelog

- [x] Patch
- [ ] Minor
- [ ] Skip

### Fixed

- [alpha] When running `cdf profile ... ` with output to a spreadsheet file, Toolkit ensures that the spreadsheet title is sanitized. 

## templates

No changes.


andersfylling:
I'm not familiar with this, nor it's use cases, but I wonder if giving the user a simple error message instead explaining which parts of the input is wrong would avoid surprises?

Then they can alter the title as needed.

PR #1807, opened 2025-07-31
[CDF-25334] 👹 Storage IO interface

doctrino:
# Description

**Context**: This introduces a new interface into Toolkit. The intention of this interface is to use it for downloading and uploading data. This will be used in the `cdf migrate/download/upload/purge` commands. I expect to implement a class for raw, all asset-centric resources, and instances. 

The motivation for having a separate PR for this interface, is to limit the size of the PRs for this new functionality. Note that there is no tests in this PR as this is just an interface and thus there is nothing to test.

## Changelog

- [ ] Patch
- [ ] Minor
- [x] Skip

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -2,7 +2,7 @@
 repos:
   - repo: https://github.com/astral-sh/uv-pre-commit
     # uv version.
-    rev: 0.8.3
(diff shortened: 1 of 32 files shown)


rsjr (commented):
risk review ok

PR #1808, opened 2025-07-31
[CDF-25368] 🗄 StoreageIO raw

doctrino:
# Description

This is the first step in adding support for downloading and uploading RAW tables. This implements the standardized interface for the download and upload functionality. 

## Changelog

- [ ] Patch
- [ ] Minor
- [x] Skip

PR #1802, opened 2025-07-30
[CDF-25338] 🤗Introduce record formats

doctrino:
# Description

This is part of a series of PRs for adding a standardized interface for reading/writing CSV/parquet/NDJson/YAML files with and without compression.

This PR introduces the reader/writers for record formats, YAML and NDJson.

## Changelog

- [ ] Patch
- [ ] Minor
- [x] Skip

PR #1826, opened 2025-08-10
[CDF-25314] ⬇ Add download command

doctrino:
# Description

Stacked on #1800, #1802, #1808

This is part of a series of PRs for introducing download/upload commands into Toolkit. 

This PR introduces the download command itself with testing. Note that this is not exposed to any end-users, thus I do not have a new release of Toolkit included.

## Changelog

- [ ] Patch
- [ ] Minor
- [x] Skip


cognite_toolkit/_cdf_tk/utils/fileio/_writers.py
```diff
@@ -36,6 +36,11 @@ def __init__(
         self._file_count_by_filename: dict[str, int] = Counter()
         self._writer_by_filepath: dict[Path, T_IO] = {}
 
+    @property
+    def file_count(self) -> int:
```
doctrino, line 40: This is just a convenience just in the new download command

cognite_toolkit/_cdf_tk/storageio/_base.py
```diff
@@ -50,6 +59,11 @@ def data_to_json_chunk(self, data_chunk: T_WritableCogniteResourceList) -> list[
     def json_chunk_to_data(self, data_chunk: list[dict[str, JsonVal]]) -> T_CogniteResourceList:
         raise NotImplementedError()
 
+    @abstractmethod
+    def configurations(self, identifier: T_StorageID) -> Iterable[StorageIOConfig]:
+        """Return configurations for the storage item."""
+        raise NotImplementedError()
```
doctrino, line 65: This is a small adjustment that became apparent when writing the download command.

cognite_toolkit/_cdf_tk/commands/_download.py
```diff
@@ -0,0 +1,90 @@
...
+        output_dir: Path,
+        verbose: bool,
+        file_format: str,
+        compression: str,
+        limit: int | None = 100_000,
+    ) -> None:
+        """Downloads data from CDF to the specified output directory.
+
+        Args:
+            identifiers: The identifiers of the resources to download.
```
ronpal, line 32: I struggle with understanding what the **identifier** is. For example if have a Time series with `NodeId(space='springfield_pi_instances', external_id='pi:160922')`

What should the command parameters be?

`cdf download data timeseries  ...`? I think we need some examples or guidance here.

cognite_toolkit/_cdf_tk/storageio/_base.py
```diff
@@ -15,6 +16,13 @@
...
 
 
+@dataclass
+class StorageIOConfig:
+    kind: str
+    folder_name: str
+    value: JsonVal
+
+
 class StorageIO(ABC, Generic[T_StorageID, T_CogniteResourceList, T_WritableCogniteResourceList]):
```
ronpal, line 26: I should have caught this before, but I don't understand the purpose of the StorageIO interface (as in what is the role, is it storage of downloaded data? So is it an abstraction of underlaying storage? Maybe a comment would help. 

Also, what is the relationship between the StorageIO interface (and its family of implementations) vs utils/fileio? Will fileio inherit from storageio at some point?

doctrino: Today we have the loader interface to interact with governed resources. Similar to that I think we need a new interface to interact with data as these have different operations. Instead of CRUD it is essentially streaming up and down. My plan is to implement StorageIO for raw, assets, events, filemetadata, timeseries, and instances.

cognite_toolkit/_cdf_tk/commands/_download.py
doctrino: In the context of StorageIO (see below), we need a way to identify which data to get. Unlike in the loaders we cannot identify individual entries, so we need another way. My suggestion is that an identifier is RAW table for Raw, dataset/asset hierarchy for classic resources, and a view and/or container for instances. They all determine how we are getting the data.

cognite_toolkit/_cdf_tk/storageio/_base.py
ronpal: And fileio is a utility independent of those interface? Got it.

cognite_toolkit/_cdf_tk/commands/_download.py
doctrino: What about calling it selector instead?

cognite_toolkit/_cdf_tk/commands/_download.py
```diff
@@ -0,0 +1,92 @@
...
+            io: The StorageIO instance that defines how to download and process the data.
+            output_dir: The directory where the downloaded files will be saved.
+            verbose: If True, prints detailed information about the download process.
+            file_format: The format of the files to be written (e.g., ".ndjson").
+            compression: The compression method to use for the downloaded files (e.g., "none", "gzip").
+            limit: The maximum number of items to download for each identifier. If None, all items will be downloaded.
+        """
+        target_directory = output_dir / io.folder_name
+        target_directory.mkdir(parents=True, exist_ok=True)
```
dmivankov, line 41: there seems to be some weakness wrt reusing directory or intermediate results stemming from
- allowing destination to have data
- auto-numerating clashes internally
- not using atomic move from temporary clean download location to target resting path

Which may show up as confusing output when running and aborting or running multiple times into same destination dir, it might be not clear what state destination ends up in: partial files, files from previous runs etc

Not clear enough how critical this is, who's the expected caller for downloads (and which precaution steps or assumptions will it have)

ronpal (approved):

cognite_toolkit/_cdf_tk/commands/_download.py
doctrino: You have some good points. I have noted them down in [this task](https://cognitedata.atlassian.net/browse/CDF-25572). 

This is a continuation of the current `cdf dump` task, which have the same behavior, so I prefer we keep it as is for now.

dmivankov:
Noticed the ticket :+1:

dmivankov:
risk review ok given that non-empty output_dir case will be followed-up :heavy_check_mark:

PR #1830, opened 2025-08-11
[CDF-25497] 🐛 Avoid reusing sessions in transformations

doctrino:
# Description

The user is experiencing error like these:
image
I think this is because of the caching of the session, thus removing that.

## Changelog

- [x] Patch
- [ ] Minor
- [ ] Skip

### Fixed

- When deploying transformations with credentials, Toolkit no longer reuses the session for the same credentials. This should remove the `Failed to bind nonce with session...` errors.

## templates

No changes.


ronpal (approved):
🏃

dmivankov:
btw, sdk should probably do the same around https://github.com/cognitedata/cognite-sdk-python/blob/master/cognite/client/data_classes/transformations/__init__.py#L148

PR #1848, opened 2025-08-18
[CDF-25547] ✍ Rename identifier to selector in StorageIO

doctrino:
# Description

A pure renaming. When finding data to download/upload/migrate it make more sense to call it a selector than an identifier. We use identifier to identify individual resources, while a selector to select chunks of data such as RAW tables, asset through dataset/hierarchy, instances through containers/views,

## Changelog

- [ ] Patch
- [x] Skip


muradsater (dismissed):

Aashutosh-cognite (commented):
🌳

einarmo:
:unicorn:

PR #1849, opened 2025-08-18
[CDF-25202] 😛 Support dump functions

doctrino:
# Description

image

## Changelog

- [x] Patch
- [ ] Skip

### Added

- [alpha] Support for dumping function configurations. 

## templates

No changes.


cognite_toolkit/_cdf_tk/commands/dump_resource.py
```diff
@@ -490,6 +499,64 @@ def __iter__(self) -> Iterator[tuple[list[Hashable], CogniteResourceList | None,
...
+            choices=choices,
+        ).ask()
+        if selected_function_ids is None:
+            raise ToolkitValueError("No functions selected for dumping.")
+        return tuple(selected_function_ids)
+
+    def __iter__(self) -> Iterator[tuple[list[Hashable], CogniteResourceList | None, ResourceLoader, None | str]]:
+        self.identifier = self._selected()
+        loader = FunctionLoader.create_loader(self.client)
+        if self.functions:
```
rsjr, line 527: It's Friday and I need more coffee, but here it goes.

Here, the function expects `self.functions` to be True, but this happens only when running this in `_interactive_select` mode. 

The else clause down below yields:

```
else:
    yield list(self.identifier), None, loader, None
```

So, if it's "non-iteractive", it will return None always. Is this the expected behavior?

PR #1851, opened 2025-08-19
[CDF-25554] 😅 Simplify migration mapping

doctrino:
# Description

In the initial testing, I have seen a few cases of confusion. As a result simplifying this. 

## Changelog

- [x] Patch
- [ ] Skip

### Removed

- [alpha] In the `cdf migrate timeseries/files` the mapping file must now be `id,space,externalId` or `id,dataSetId,space,externalId`. You can no longer have `externalId` in the first column to reference the asset-centric timeseries/file. The reason is to simplify the use and make the mapping file only have one valid way of referencing asset-centric resources, and thus cause less confusion. 

## templates

No changes.


Aashutosh-cognite (approved):
✌🏻

rsjr (commented):
risk review ok
good stuff

PR #1852, opened 2025-08-19
[CDF-25531] 📉 Charts data

doctrino:
# Description

This implements data classes for the Chart Data object. This is necessary for writing the migration from asset-centric timeseries to CogniteTimeSeries

## Changelog

- [ ] Patch
- [x] Skip

### DIFF
--- a/cognite_toolkit/_cdf_tk/client/api/charts.py
+++ b/cognite_toolkit/_cdf_tk/client/api/charts.py
@@ -56,7 +56,7 @@ class ChartsAPI(APIClient):
             result.extend([Chart._load(item, cognite_client=self._cognite_client) for item in response.json()["items"]])
 
         if not result:
-            raise ValueError("No charts were upserted. This may indicate an issue with the upsert operation.")
+            raise ValueError("No charts were upserted. This may indicate an issue with the upsert endpoint.")
 
         if isinstance(items, ChartWrite):
             if len(result) != 1:
@@ -65,6 +65,11 @@ class ChartsAPI(APIClient):
                     "This may indicate an issue with the upsert operation."
                 )
             return result[0]
+        elif len(result) != len(items):
+            raise ValueError(
+                "The number of upserted charts does not match the number of input charts. "
+                "This may indicate an issue with the upsert endpoint."
+            )
         return result
 
     @overload
--- a/cognite_toolkit/_cdf_tk/client/data_classes/charts_data.py
+++ b/cognite_toolkit/_cdf_tk/client/data_classes/charts_data.py
@@ -26,8 +26,7 @@ class ChartObject(CogniteObject):
     def _load(cls, resource: dict[str, Any], cognite_client: CogniteClient | None = None) -> Self:
         """Load a ChartObject from a dictionary."""
         instance = super()._load(resource, cognite_client=cognite_client)
-        known_camel_case_props = cls._known_camel_case_props()
-        instance._unknown_fields = {k: v for k, v in resource.items() if k not in known_camel_case_props}
+        instance._unknown_fields = {k: v for k, v in resource.items() if k not in cls._known_camel_case_props()}
         return instance
 
     @classmethod


cognite_toolkit/_cdf_tk/client/data_classes/charts_data.py
```diff
@@ -0,0 +1,300 @@
...
+class SubSetting(ChartObject):
+    auto_align: bool | None = None
+
+
+@dataclass
+class FlowElement(ChartObject):
+    id: str | None = None
+    type: str | None = None
+    position: tuple[float | None, float | None] | None = None
+    data: dict[str, object] | None = None
```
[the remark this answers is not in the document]
doctrino: That is unknown. So it has to be an object.

cognite_toolkit/_cdf_tk/client/data_classes/charts_data.py
```diff
@@ -0,0 +1,300 @@
...
+    # This allows unknown fields to be preserved when loading and dumping ChartObjects
+    # (serialization and deserialization).
+    _unknown_fields: dict[str, object] | None = field(default=None, init=False, repr=False)
+
+    @classmethod
+    def _load(cls, resource: dict[str, Any], cognite_client: CogniteClient | None = None) -> Self:
+        """Load a ChartObject from a dictionary."""
+        instance = super()._load(resource, cognite_client=cognite_client)
+        known_camel_case_props = {to_camel_case(f.name) for f in fields(cls)}
+        instance._unknown_fields = {k: v for k, v in resource.items() if k not in known_camel_case_props}
```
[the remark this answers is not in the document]
doctrino: Moved it to a lru_cached method.

cognite_toolkit/_cdf_tk/client/api/charts.py
```diff
@@ -0,0 +1,116 @@
...
+        filter_: dict[str, str | bool] = {}
+        if visibility is not None:
+            filter_["visibility"] = visibility.upper()
+        if is_owned is not None:
+            filter_["isOwned"] = is_owned
+
+        response = self._post(
+            url_path=self._RESOURCE_PATH + "/list",
+            json={"filter": filter_} if filter_ else {},
+        )
```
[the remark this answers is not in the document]
doctrino: Not supported by the API

doctrino:
> PR is at ~750 lines of code, but I understand the need to not split this added API into more parts - _but_ it would be nice to see a comment justifying it in the description 😉
> 
> After reading through I am also not convinced that the team reviews have been particularly in-depth (looking at you @muradsater 👀 ).

This PR got outdated when the one it was based on got merged, and I forgot to update it. Sorry. It is in-fact within the 500 lines limit.

In defense of @muradsater, your comments are good but are mostly docstrings and minor refactorings, thus mostly low risk. The adjustments to the tests are very good comments from you, but are subtle and easy to miss.

cognite_toolkit/_cdf_tk/client/api/charts.py
```diff
@@ -54,6 +54,10 @@ def upsert(self, items: ChartWrite | Sequence[ChartWrite]) -> Chart | ChartList:
                 json=body,
             )
             result.extend([Chart._load(item, cognite_client=self._cognite_client) for item in response.json()["items"]])
+
+        if not result:
```
rbtcollins, line 58: Shouldn't we expect exactly as many results as inputs? Send 10, get 10?

rbtcollins: I'm not sure, its more a discussion - perhaps in the SDK  / API channels. Clearly for a single item upsert we're enforcing it. So then if we upsert 2, can we rely on 2 coming back.

cognite_toolkit/_cdf_tk/client/api/charts.py
```diff
@@ -42,7 +42,7 @@ def upsert(self, items: ChartWrite | Sequence[ChartWrite]) -> Chart | ChartList:
             items (ChartWrite | Sequence[ChartWrite]): Chart(s) to upsert. Can be a single ChartWrite or a sequence of ChartWrite.
 
         Returns:
-            ChartList: List of upserted charts if multiple items are provided, otherwise a single Chart.
+            Chart | ChartList: List of upserted charts if multiple items are provided, otherwise a single Chart.
         """
         item_sequence = items if isinstance(items, Sequence) else [items]
```
rbtcollins, line 47: Sequence is deprecated and being removed. Use [Iterable](https://docs.python.org/3/library/collections.abc.html#collections.abc.Iterable) (for type signatures). Per the docs on that, you can either use isinstance, or perhaps call `iter(items)` and catching a failure :- but that is an edge case and I suspect just isinstance will be fine.

cognite_toolkit/_cdf_tk/client/data_classes/charts_data.py
```diff
@@ -0,0 +1,303 @@
...
+    @classmethod
+    def _load(cls, resource: dict[str, Any], cognite_client: CogniteClient | None = None) -> Self:
+        """Load a ChartObject from a dictionary."""
+        instance = super()._load(resource, cognite_client=cognite_client)
+        known_camel_case_props = cls._known_camel_case_props()
+        instance._unknown_fields = {k: v for k, v in resource.items() if k not in known_camel_case_props}
+        return instance
+
+    @classmethod
+    @lru_cache(maxsize=1)
```
rbtcollins, line 33: This deserves a docstring or comment explaining why it is safe to cache on the class [because fields are presumed immutable for a class], and we're not going to support dynamic programming that violates that expectation.

cognite_toolkit/_cdf_tk/client/data_classes/charts_data.py
```diff
@@ -0,0 +1,303 @@
...
+class SubSetting(ChartObject):
+    auto_align: bool | None = None
+
+
+@dataclass
+class FlowElement(ChartObject):
+    id: str | None = None
+    type: str | None = None
+    position: tuple[float | None, float | None] | None = None
+    data: dict[str, object] | None = None
```
rbtcollins, line 85: Do we know it is definitely an object? Would Any be more accurate?

cognite_toolkit/_cdf_tk/client/data_classes/charts_data.py
```diff
@@ -0,0 +1,303 @@
...
+@dataclass
+class Flow(ChartObject):
+    zoom: float | None = None
+    elements: list[FlowElement] | None = None
+    position: tuple[float | None, float | None] | None = None
+
+    def dump(self, camel_case: bool = True) -> dict[str, Any]:
+        data = super().dump(camel_case=camel_case)
+        if self.elements:
+            data["elements"] = [el.dump(camel_case=camel_case) for el in self.elements]
```
rbtcollins, line 97: Doesn't super.dump propogate the camel_case parameter down to FlowElement?

cognite_toolkit/_cdf_tk/client/data_classes/charts_data.py
```diff
@@ -0,0 +1,303 @@
+import sys
+from dataclasses import dataclass, field, fields
```
rbtcollins, line 2: pydantic can be a lot nicer in my experience - and faster too - have you considered using it here? it might resolve the need to manually set camel_case everywhere...

tests/test_unit/test_cdf_tk/test_client/test_charts.py
```diff
@@ -0,0 +1,124 @@
...
+    def test_chart_data_changed(self) -> None:
+        """The ChartData is frontend of the Chart object, and it is not enforced in any way by the backend API.
+        Thus, it can change completely without any notice. This tests ensures that whatever the ChartData is,
+        the serialization and deserialization works correctly.
+        """
+        chart_data = {
+            "this": "is",
+            "completely": {
+                "changed": ["compared", "to", "the", "previous", "version"],
+            },
```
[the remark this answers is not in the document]
rbtcollins: I would add - spend a little time with hypothesis, and teach it how to do that, Then never worry about it again.

cognite_toolkit/_cdf_tk/client/api/charts.py
rbtcollins: I need to tweak this comment, because reality is messy. I misspoke on Sequence itself being deprecated (it stays, but it doesn't model what you are asking for in your code).

The deprecation which is one of those carefully worded things is here...
https://typing.python.org/en/latest/spec/historical.html#generics-in-standard-collections
-> collections.abc.Sequence and typing.Sequence are deprecated aliases

I do recommend switching to Iterable[] instead, because you do not demand more of the input than iterability, AFAICT.

cognite_toolkit/_cdf_tk/client/data_classes/charts_data.py
doctrino: This follows the PySDK which we do not use pydantic as it is too big as I understand it to have as a dependency.

cognite_toolkit/_cdf_tk/client/data_classes/charts_data.py
doctrino: No, unfortunately not Pydantic would be stronger.

cognite_toolkit/_cdf_tk/client/api/charts.py
rbtcollins: That is true. And we do depend on `__len__` from my suggestion that we check the length of the return value.

However I like to think about the code I would like to write. And, for instance, 

`api.upsert(make_chart(**params) for params in use_case)` would be convenient.

There's no good reason we shouldn't support that. But supporting that means not depending on indexed access. And we don't! Dropping the len requirement is trivial too - change chunker to return its offsets as it goes, or just capture it inline:

```
  count = 0
  for (pos, chunk) in chunker ....
    ...
    count += len(body["items"])

....
    elif len(result != count:
```

Further, there are many many places in the stdlib where infinite iterators are not useful but are accepted and the caller needs to choose what they want (for instance `itertools,.chain` - putting an infinite iterator in the first parameter makes calling chain meaningless(; I see no unique burden placed on callers here.

cognite_toolkit/_cdf_tk/client/data_classes/charts_data.py
rbtcollins: We hate falsely narrow types more, because then the correct handling code doesn't get written at all...

cognite_toolkit/_cdf_tk/client/api/charts.py
haakonvt: Good points, we typically have a separate method for iteration-based fetching, but no such method for create/update/etc. so I like the idea of happily accepting iterators of any kind! Would be good to start this out in the SDK though for completeness

PR #1854, opened 2025-08-20
[prep CDF-25295] 🖌 Introduce ingestion/consumer view

doctrino:
# Description

This extends the MigrationMapping class that is used when asset-centric resources are converted to instances with ingestionView and consumerView, which will be used for custom ingestion view as well as preferred consumer View. The consumer view will be used in step 4 of the process in the Canvas migration. 

## Changelog

- [ ] Patch
- [x] Skip


doctrino:
Moved to #1860

PR #1859, opened 2025-08-22
[prep CDF-25295] 🙃 Read csv unprocessed.

doctrino:
# Description

We have a standardized interface for reading data from .parquet, .csv, .ndsjon, .yaml. To comply with that the csv reader parser the data.

This PR introduces a special `read_chunks_unprocessed` for csv to enable simply reading a csv file as is without any processing. 

## Changelog

- [ ] Patch
- [x] Skip

### DIFF
--- a/cdf.toml
+++ b/cdf.toml
@@ -15,7 +15,7 @@ infield = true
 dump-data = true
 migrate = true
 purge_instances = true
-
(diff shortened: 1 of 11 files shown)


erlendvollset (commented):
🦄

PR #1872, opened 2025-08-28
[CDF-25609]🧵Thread safe writing

doctrino:
# Description

**Reviewer** This is a 29 lines change with the rest being testing.

This is a preparation to use the FileWriter with the ProducerWorker which ultimately will be used in migration. The plan is to use the `NDJsonWriter` to capture issues and writing them to fail as they happing in downloading asset-centric resources, conversion to data model instances, and writing to data model instances.

## Changelog

- [ ] Patch
- [x] Skip


Aashutosh-cognite (approved):

cognite_toolkit/_cdf_tk/utils/fileio/_writers.py
```diff
@@ -44,18 +45,22 @@ def __init__(
         self.max_file_size_bytes = max_file_size_bytes
         self._file_count_by_filename: dict[str, int] = Counter()
         self._writer_by_filepath: dict[Path, T_IO] = {}
+        self._lock = threading.RLock()  # Use RLock to allow recursive locking
```
haakonvt, line 48: I don't understand why you need a reentrant lock here, could you elaborate?
```suggestion
        self._lock = threading.Lock()
```

doctrino: I guess it is overkill as the lock is only used internally the FileWriter.

dbrattli (approved):
Looks solid! Risk review mentoring done. @haakonvt feel free to add a unicorn to this PR.

cognite_toolkit/_cdf_tk/utils/fileio/_writers.py
```diff
@@ -77,18 +83,20 @@ def _get_writer(self, filepath: Path, filename_base: str) -> T_IO:
         return self._writer_by_filepath[filepath]
 
     def __enter__(self) -> Self:
-        self._file_count_by_filename.clear()
-        self._writer_by_filepath.clear()
-        return self
+        with self._lock:
+            self._file_count_by_filename.clear()
+            self._writer_by_filepath.clear()
```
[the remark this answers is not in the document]
dbrattli: Returning `self` seems the right approach since the context manager will be used once and the constructor already initializes them as empty. If a defensive approach is needed:

```py
def __enter__(self) -> Self:
    # Defensive check - should never trigger in normal usage
    if self._writer_by_filepath or self._file_count_by_filename:
        raise RuntimeError("FileWriter context manager should not be reused")
    return self
```

PR #1878, opened 2025-08-28
[prep CDF-25553] 👷‍♂️ Validate file access

doctrino:
# Description

Extends the validator class for giving a good error message to the user if the lack access to the FilesAcl This will be run at the beginning of commands to ensure that the user gets a nice error message. 

## Changelog

- [ ] Patch
- [x] Skip


Aashutosh-cognite (approved):

tests/test_unit/test_cdf_tk/test_utils/test_validate_access.py
```diff
@@ -3,6 +3,7 @@
     Capability,
     DataModelInstancesAcl,
     DataModelsAcl,
+    FilesAcl,  # Added import for FilesAcl
```
andreavs, line 6: super-nit: I guess this inline comment was added by cursor? Feels out of place :grin:

andreavs (commented):
:unicorn:

PR #1880, opened 2025-08-29
[CDF-25626]  Robust stopping

doctrino:
# Description

In addition to the change described below. I did some cleanup and simplification of the WorkerProducer. 

image

## Changelog

- [x] Patch
- [ ] Skip

### Improved

- [alpha] When running the commands `cdf purge instances/download/upload` the user now use the esc-key to abort the operation. This is to avoid accidentally pressing `enter`, which is often used to select what to purge/download/upload.

### Fixed

- [alpha} When downloading a raw table `cdf download raw`, the CLI now correctly exits when the operation is interrupted by the user. 

## templates

No changes.


cognite_toolkit/_cdf_tk/storageio/_raw.py
```diff
@@ -32,7 +32,9 @@ def download_iterable(self, selector: RawTable, limit: int | None = None) -> Ite
             db_name=selector.db_name,
             table_name=selector.table_name,
             limit=limit,
-            partitions=8,
+            # We cannot use partitions here as it is not thread safe. This spawn multiple threads
+            # that are not shut down until all data is downloaded. We need to be able to abort.
+            partitions=None,
```
doctrino, line 37: This is the bug.

cognite_toolkit/_cdf_tk/utils/producer_worker.py
```diff
@@ -22,6 +25,10 @@
 T_Download = TypeVar("T_Download", bound=Sized)
 T_Processed = TypeVar("T_Processed", bound=Sized)
 
+# Sentinels for signaling finish
+PROCESS_FINISH_SENTINEL = object()
+WRITE_FINISH_SENTINEL = object()
```
doctrino, line 30: I am replacing the non-tread safe flags with these SENTINEL to shut down the process and write workers.

cognite_toolkit/_cdf_tk/utils/producer_worker.py
```diff
@@ -91,20 +98,23 @@ def __init__(
...
-        self.stopped_by_user = False
+    @property
+    def error_occurred(self) -> bool:
+        """Indicates if an error occurred during execution."""
+        return self._error_event.is_set()
+
+    @property
+    def stopped_by_user(self) -> bool:
+        """Indicates if the execution was stopped by the user."""
+        return self._stop_event.is_set()
```
doctrino, line 117: Replacing this with thread safe events.

cognite_toolkit/_cdf_tk/utils/producer_worker.py
```diff
@@ -148,103 +167,144 @@ def user_input_listener() -> None:
...
+
+    def _put_with_error_check(self, items: T_Item, target_queue: queue.Queue[T_Item]) -> bool:
+        """Helper to put items into a queue with error checking."""
+        while not self._error_event.is_set():
+            try:
+                target_queue.put(items, timeout=0.5)
+                return True
+            except queue.Full:
+                continue
+        return False
```
doctrino, line 226: Instead of using `queue.put(items)` which is a blocking call, this ensure that we check if any error has occurred. The previous version of WorkerProducerExecutor could freeze if an error occurred while whiting to put an item on the queue.

cognite_toolkit/_cdf_tk/loaders/_resource_loaders/agent_loaders.py
```diff
@@ -22,7 +22,7 @@ class AgentLoader(ResourceLoader[str, AgentUpsert, Agent, AgentUpsertList, Agent
     list_write_cls = AgentUpsertList
     kind = "Agent"
     _doc_base_url = ""
-    _doc_url = "https://api-docs.cognite.com/20230101-alpha/tag/Agents/operation/main_api_v1_projects__projectName__ai_agents_post"
+    _doc_url = "https://api-docs.cognite.com/20230101-beta/tag/Agents/operation/main_ai_agents_post/"
```
doctrino, line 25: This is unrelated - but the tests starts failing immediately when the URL is no longer available - so it needs to be fixed to get the tests to pass.

Aashutosh-cognite (approved):

cognite_toolkit/_cdf_tk/utils/producer_worker.py
```diff
@@ -148,103 +167,152 @@ def user_input_listener() -> None:
...
-                    self.stopped_by_user = True
                     self._stop_event.set()
                     break
 
-            self._stop_event.set()
+            # After a possible interrupt, we must wait for all threads to finish their
+            # graceful shutdown. This is important to prevent data loss.
+            for t in [download_thread, process_thread, write_thread, input_thread]:
+                if t.is_alive():
+                    t.join()
```
gaelenm, line 185: This may be a bit of an edge case, but what happens if another keyboard interrupt occurs during this join?

If the `download_thread` is interrupted in the above loop, could the `process_thread` or the `write_thread` be interrupted here?  
Or would it matter? 
I guess a better question might be are there any side effects of not gracefully shutting down the `process_thread` or `write_thread` if an exception is not caught here?

gaelenm (approved):

cognite_toolkit/_cdf_tk/utils/producer_worker.py
doctrino: @gaelenm Given the explanation above, I assume this is safe to leave as is?

gaelenm (approved):

Aashutosh-cognite (approved):

sighol (approved):
:unicorn:

PR #1885, opened 2025-08-30
[CDF-25639] 💹 Added chart interactive select

doctrino:
# Description

This PR introduces an interactive select for for the user to select which charts to migrate (will also be used for download and upload)

## Changelog

- [ ] Patch
- [x] Skip

### DIFF
--- a/cognite_toolkit/_builtin_modules/common/cdf_apm_base/data_models/views/ApmConfig.view.yaml
+++ b/cognite_toolkit/_builtin_modules/common/cdf_apm_base/data_models/views/ApmConfig.view.yaml
@@ -1,7 +1,7 @@
 externalId: APM_Config
 name: APM Config
 description: "Config data for APM applications"
-version: 1 # must be this exact version
+version: '1' # must be this exact version
 space: APM_Config # Must be this exact space
 properties:
   name:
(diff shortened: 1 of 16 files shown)


Aashutosh-cognite (dismissed):

cognite_toolkit/_cdf_tk/utils/interactive_select.py
```diff
@@ -366,6 +369,82 @@ def _select_external_ids(self, select_filter: CanvasFilter) -> list[str]:
...
+        ).ask()
+        if user_response is None:
+            raise ToolkitValueError("No Chart selection made. Aborting.")
+        return user_response
+
+    def _select_external_ids(self, select_filter: ChartFilter) -> list[str]:
+        available_charts = self.client.charts.list(visibility=(select_filter.visibility or "PUBLIC"))
+        if select_filter.select_all and select_filter.owned_by is None:
+            return [chart.external_id for chart in available_charts]
+        users = self.client.iam.user_profiles.list(limit=-1)
```
amorken, line 408: How does this degrade without a user profile read capability?

cognite_toolkit/_cdf_tk/utils/interactive_select.py
```diff
@@ -366,6 +369,82 @@ def _select_external_ids(self, select_filter: CanvasFilter) -> list[str]:
...
+        display_name_by_user_identifier = {user.user_identifier: user.display_name or "missing" for user in users}
```
amorken, line 409: Seeing how this is used (dict.get() with a default value equal to the key) wouldn't it degrade better if the user_identifier without a display name was simply not included as a key in the map rather than mapped to "missing"?

amorken (commented):
Overall looks good, a couple of questions that may well have answers that do not need code changes.

ronpal (approved):

amorken (approved):
🦄 
Sorry for the delay 😞

PR #1909, opened 2025-09-02
[CDF-24802] 🥟 Added ExtractionPipelineYAML

doctrino:
# Description

This PR introduces the ExtractionPipelineYAML class to validate extraction pipeline. Note this is not yet exposed, so no change to the end user.

## Changelog

- [ ] Patch
- [x] Skip

### DIFF
--- a/cognite_toolkit/_cdf_tk/resource_classes/extraction_pipeline.py
+++ b/cognite_toolkit/_cdf_tk/resource_classes/extraction_pipeline.py
@@ -1,0 +1,55 @@
+from typing import Literal
+
+from pydantic import Field
+
+from .base import BaseModelResource, ToolkitResource
+
+
+class RawTable(BaseModelResource):
+    db_name: str = Field(description="Database name", min_length=1)
+    table_name: str = Field(description="Table name", min_length=1)
+
+
+class Contact(BaseModelResource):
+    name: str | None = Field(None, description="Contact name")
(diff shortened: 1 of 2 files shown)


cognite_toolkit/_cdf_tk/resource_classes/extraction_pipeline.py
```diff
@@ -0,0 +1,54 @@
+from typing import Literal
+
+from pydantic import Field
+
+from .base import BaseModelResource, ToolkitResource
+
+
+class RawTable(BaseModelResource):
```
Aashutosh-cognite, line 8: already a RawTableYAML class is there, why not use that? Is it because of the base class `ToolkitResource`

Aashutosh-cognite (approved):
LGTM. Just one doubt or adjustment.

andreavs (commented):
:unicorn: 

Wondering if it would be better to try to autogenerate these classes from the service contract, but that is outside the scope of this particular PR.

doctrino:
> 🦄
> 
> Wondering if it would be better to try to autogenerate these classes from the service contract, but that is outside the scope of this particular PR.

We have small deviation here and there. For example, we do not allow internal IDs, and often require externalIds even though they are optional. Note I have also experimented with this type of generation for PySDK, and our service contracts are highly inconsistent which makes it very hard and likely easier to handcraft like this.

PR #1911, opened 2025-09-03
[CDF-25666] 🗺 Data mapper interface

doctrino:
# Description

**Context**: This is part of the `alpha` migration tooling in Toolkit, for overview [see here](https://cognitedata.atlassian.net/wiki/spaces/~170890702/pages/5455773697/Migration+Tooling+in+Toolkit). This is step 3 and 4 of that process. 

This introduces a generalized interface that we will use to go from asset-centric assets/event/timeseries/files/annotations -> to DM equivalent. It will also be used for charts/canvas were the mapping operation is to update the asset-centric references to data modeling references. 

## Changelog

- [ ] Patch
- [x] Skip


rsjr (approved):
risk review ok

PR #1919, opened 2025-09-06
[CDF-25698] 🤓 HTTP Client

doctrino:
# Description

Adds an HTTPClient that does not use any queues to make it compatible with a WorkerProducer. 

## Changelog

- [ ] Patch
- [x] Skip


doctrino:
Moved to #1922 and #1923

PR #1922, opened 2025-09-07
[CDF-25698] 🤓 HTTP Client Part 1

doctrino:
# Description

Adds an HTTPClient that does not use any queues or threading to make it compatible with a WorkerProducer.

## Changelog

- [ ] Patch
- [x] Skip

### DIFF
--- a/tests/test_unit/test_cdf_tk/test_utils/test_http_client.py
+++ b/tests/test_unit/test_cdf_tk/test_utils/test_http_client.py
@@ -178,6 +178,12 @@ class TestHTTPClient:
         assert response.error == "service unavailable"
         assert len(rsps.calls) == 1
 
+    def test_raise_if_already_retied(self, http_client_one_retry: HTTPClient) -> None:
+        http_client = http_client_one_retry
+        bad_request = ParamRequest(endpoint_url="https://example.com/api/resource", method="GET", status_attempt=3)
+        with pytest.raises(RuntimeError, match="RequestMessage has already been attempted 3 times."):
+            http_client.request_with_retries(bad_request)
+
     def test_error_text(self, http_client: HTTPClient, rsps: responses.RequestsMock) -> None:
         rsps.get("https://example.com/api/resource", json={"message": "plain_text"}, status=401)
         results = http_client.request(ParamRequest(endpoint_url="https://example.com/api/resource", method="GET"))


cognite_toolkit/_cdf_tk/utils/http_client/_client.py
```diff
@@ -0,0 +1,274 @@
...
+            body = response.json()
+        except ValueError as e:
+            return request.create_responses(response, error_message=f"Invalid JSON response: {e!s}")
+
+        if 200 <= response.status_code < 300:
+            return request.create_responses(response, body)
+        elif request.status_attempt < self._max_retries and response.status_code in self._retry_status_codes:
+            request.status_attempt += 1
+            time.sleep(self._backoff_time(request.total_attempts))
+            return [request]
```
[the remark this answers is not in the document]
doctrino: This is desirable. We want to count status attempts differently from timeout and connection.

cognite_toolkit/_cdf_tk/utils/http_client/_client.py
```diff
@@ -0,0 +1,275 @@
...
+        return final_responses
+
+    def _create_thread_safe_session(self) -> requests.Session:
+        session = requests.Session()
+        adapter = HTTPAdapter(
+            pool_connections=self._pool_connections,
+            pool_maxsize=self._pool_maxsize,
+            max_retries=Retry(total=0),  # We handle retries manually
+        )
+        session.mount("http://", adapter)
```
ronpal, line 135: Do we ever allow `http`?

ronpal (dismissed):

ronpal (approved):

cognite_toolkit/_cdf_tk/utils/http_client/_client.py
```diff
@@ -0,0 +1,273 @@
+import gzip
```
polomani, line 1: I don't get it, is this going to be used instead of pysdk? why not just set `max_workers=1` to limit concurrency?

doctrino: Multiple reasons.
1. We need to handle items differently than the SDK, see next PR https://github.com/cognitedata/toolkit/pull/1923
2. We support APIendpoints that are not (yet) supported by the SDK, then relying on the internals of the SDK is a risk as earlier risk-review has pointed out.
3. We need to be able to get the json directly, not necessarily use the same dataclass as the SDK.

Note that this is version 2 of an HTTPClient, I made one earlier this summer that I will remove once this has replaced it.

cognite_toolkit/_cdf_tk/utils/http_client/_client.py
```diff
@@ -0,0 +1,273 @@
...
+import random
+import socket
+import sys
+import time
+from collections.abc import MutableMapping, Sequence, Set
+from typing import Literal
+
+import requests
+import urllib3
```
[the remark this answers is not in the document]
doctrino: That is another dependency. We will move with the PySDK on that.

cognite_toolkit/_cdf_tk/utils/http_client/_client.py
doctrino: 1. In the regular SDK, if you have 1000 items and one of them have an issue, all fails. In part to referenced above, we split ItemsRequest to ensure that each item is tried. We will use this in upload and migration commands. 
3. That may be, but changing the PySDK is a slower process than Toolkit and I need this yesterday. 

Note that Toolkit already have its own HTTPClient, this is just a refactoring of that (easier to write a new and replace, than to do an actual refacotring).

ronpal (approved):

PR #1927, opened 2025-09-08
🐛Fix documentation

doctrino:
# Description

You cannot give `externalId` for the asset-centric resources any more. Just internal ID

## Changelog

- [ ] Patch
- [x] Skip


rsjr (approved):
risk review ok

PR #1932, opened 2025-09-08
[CDF-24817] 🎏 StreamlitYAML

doctrino:
# Description

We are in the progress of adding pydantic classes to match the YAML
format Toolkit expects for configurations. The goal is to give better
error message to the user on syntax errors.

This PR introduces the StreamlitYAML class to validate streamlit applications. Note this is not yet exposed, so no change
to the end user.

## Changelog

- [ ] Patch
- [x] Skip


haakonvt (commented):
🦄

PR #1935, opened 2025-09-08
[CDF-24808] 🎈 Relationship YAML

doctrino:
# Description

This PR introduces the RelatoinshipYAMl to validate relationships. Note this is not yet exposed, so no change to the end user.

## Changelog

- [ ] Patch
- [x] Skip


Aashutosh-cognite (dismissed):
MIght be good to remove `| list` in type hint.
https://github.com/cognitedata/toolkit/pull/1935/files#r2331414563

PR #1937, opened 2025-09-09
[CDF-25713] ➕ Add as_id to StorageIO interface

doctrino:
# Description

Add a `.as_id` to the StroageIO interface. This interface is used in the download/upload/migrate/purge commands. This new as_id method will be used to log each item. For example, in the migration command we need to log the outcome of the migration of each Asset to CogniteAsset.

## Changelog

- [ ] Patch
- [x] Skip


Aashutosh-cognite (approved):
LGTM

PR #1942, opened 2025-09-10
[CDF-25717] 😅Remove auth check on extraction pipeline config

doctrino:
# Description

## User error
The user failed to deploy an extraction pipeline configuration as they only had `extractionPipelineAcl` scoped to  a dataset:
image

## Solution
Temporary solution remove this check.

## Risk
This means we do not give a good error message when the user lacks extraction pipeline acl, but instead they will get an API failure on the create requests instead. This is a temporarily solution to unblock user, as a permanent solution requires a redesign of how we do this check in Toolkit to account for dependencies between resources. I captured this in the following task: https://cognitedata.atlassian.net/browse/CDF-25720?atlOrigin=eyJpIjoiNjBhN2EwNDEyOWFjNDAwMjk2NDhhYjk3NmRjZThhNzQiLCJwIjoiaiJ9

The risk is worth it as this is currently a blocker for our users.

## Changelog

- [x] Patch
- [ ] Skip

### Fixed

- When deploying extraction pipeline configurations, Toolkit no longer require `extractionPipelineAcl` that is all scoped. 

## templates

No changes.


cognite_toolkit/_cdf_tk/loaders/_resource_loaders/extraction_pipeline_loaders.py
```diff
@@ -27,7 +27,6 @@
 )
 from cognite.client.data_classes.capabilities import (
     Capability,
-    ExtractionConfigsAcl,
```
doctrino, line 30: Pending, asking the extractor team whether this is used (or deprecated) as it is no longer in the API docs:
image

PR #1944, opened 2025-09-10
😁Use new YAML classes

doctrino:
# Description

These classes are already implemented and well tested. Setting the `yaml_cls` in the Loader (CRUD interface) will start using this in the build command to validate the YAML syntax of the user.

## Details:
The testing of the YAML resource classes we use to validate the syntax of the user configurations: https://github.com/cognitedata/toolkit/tree/main/tests/test_unit/test_cdf_tk/test_resource_classes

Then, by setting yaml_cls in the resource loader (will be renamed ResourceCRUD), it will be used in the cdf build command to give the user a warning about whether they follow the syntax here: https://github.com/cognitedata/toolkit/blob/main/cognite_toolkit/_cdf_tk/commands/build_cmd.py#L603

## Changelog

- [x] Patch
- [ ] Skip

### Improved

- Warning on YAML syntax errors when running `cdf build` for resources types hosted extractor mapping and destinations, as well as streamlit applications.

## templates

No changes.


Aashutosh-cognite (approved):
🚀

PR #1945, opened 2025-09-10
[CDF-25693] Fixing default config path

ronpal:
# Description

When loading modules, the path to default configuration was hardcoded to builtin modules, so the generated config.<env>.yaml files were populated with the wrong content. This is now fixed.

## Changelog

- [x] Patch
- [ ] Skip

### Fixed

- Fixed creation of environment-specific config files to follow the relevant modules. 

## templates

No changes.


doctrino (approved):

polomani (commented):
🦄

PR #1946, opened 2025-09-10
[CDF-25696]  🤗 CI: updated secret reference

doctrino:
# Description

Removes the use of environments for secrets used in the PR builds

This should take care of the this type of noise:
image

## Changelog

- [ ] Patch
- [x] Skip


.github/workflows/build.yml
```diff
@@ -87,13 +87,12 @@ jobs:
   coverage:
     name: Create coverage report
     runs-on: ubuntu-latest
-    environment: dev
```
doctrino, line 90: I have updated the secrets for the repo such that they are available directly.

PR #1955, opened 2025-09-11
[CDF-24806] 😀 Hosted Extractor Job YAML

doctrino:
# Description

This PR introduces the HostedExtractorJobYAMLclass to validate hosted extractor job. Note this is not yet exposed, so no change to the end user.
## Changelog

- [ ] Patch
- [x] Skip


ikezedev (approved):
🦄

PR #1965, opened 2025-09-14
[CDF-25742] α Support setting alpha flag

doctrino:
# Description

In the HTTPCLient in Toolkit, we need to be able to do calls against the alpha version of the CDF API.  I will need to use this  in the migration, were I need it to set-pending-instance-ids on files and timeseries for example. 

## Changelog

- [ ] Patch
- [x] Skip

### DIFF
--- a/cognite_toolkit/_cdf_tk/utils/http_client/_client.py
+++ b/cognite_toolkit/_cdf_tk/utils/http_client/_client.py
@@ -143,7 +143,7 @@ class HTTPClient:
         session.mount("https://", adapter)
         return session
 
-    def _create_headers(self) -> MutableMapping[str, str]:
+    def _create_headers(self, api_version: str | None = None) -> MutableMapping[str, str]:
         headers: MutableMapping[str, str] = CaseInsensitiveDict()
         headers.update(requests.utils.default_headers())
         auth_name, auth_value = self._config.credentials.authorization_header()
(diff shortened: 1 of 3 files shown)


Aashutosh-cognite (approved):
🚀

cognite_toolkit/_cdf_tk/utils/http_client/_client.py
```diff
@@ -152,7 +152,7 @@ def _create_headers(self) -> MutableMapping[str, str]:
         headers["accept"] = "application/json"
         headers["x-cdp-sdk"] = f"CogniteToolkit:{get_current_toolkit_version()}"
         headers["x-cdp-app"] = self._config.client_name
-        headers["cdf-version"] = self._config.api_subversion
+        headers["cdf-version"] = api_version or self._config.api_subversion
```
thorkildcognite, line 155: As requests ignore any header values that are None, this shouldn't be an actual problem.

tests/test_unit/test_cdf_tk/test_utils/test_http_client.py
```diff
@@ -204,6 +204,22 @@ def test_error_text(self, http_client: HTTPClient, rsps: responses.RequestsMock)
         assert response.status_code == 401
         assert response.error == '{"message": "plain_text"}'
 
+    def test_request_alpha(self, http_client: HTTPClient, rsps: responses.RequestsMock) -> None:
```
thorkildcognite, line 207: You could have had a test that you are not always sending alpha by mistake, but I think that is something you can pick up as part of a possible future fix. Nothing to block this PR for.

thorkildcognite (commented):
🦄

PR #1974, opened 2025-09-15
[prep CDF-25015] 🧵 Thread safe dict

doctrino:
# Description

We need a threadsafe dictionary to track identifier in the migration command.

## Changelog

- [ ] Patch
- [x] Skip


cognite_toolkit/_cdf_tk/utils/safe_dict.py
```diff
@@ -0,0 +1,96 @@
+import threading
+from collections import UserDict
+from collections.abc import Iterator
+from typing import Any, Generic
+
+from cognite_toolkit._cdf_tk.utils.useful_types import T_ID, T_Value
+
+
+class ThreadSafeDict(UserDict, Generic[T_ID, T_Value]):
```
msusag, line 9: Should the python file be called thread_safe_dict.py, to reflect that the "safe" is related to thread-safety.

cognite_toolkit/_cdf_tk/utils/safe_dict.py
```diff
@@ -0,0 +1,96 @@
...
+    """A thread-safe dictionary implementation.
+
+    This class extends `UserDict` to provide a dictionary that is safe for concurrent access
```
msusag, line 12: The getter methods in this thread safe dictionary class is using the same lock as the writers.
So you will not achieve true multi-threading for the operations in this class. This is fine though as long as callers do not expect it to be really multi-threaded (for reads at least).

A further improvement of this class would be to create a read-write lock, and use that accordingly in the various methods. This can of course be done in later iterations. It can get tricky to ensure fairness among writers and readers with this, especially write starvation would need to be handled.

I suggest for now just to document in the docstring that all operations, reads and writes, are exclusive access (Pessimistic locking). And add a TODO about support for adding support for a shared exclusive lock, to allow multiple readers for more performance read handling.

tests/test_unit/test_cdf_tk/test_utils/test_safe_dict.py
```diff
@@ -0,0 +1,283 @@
...
+        d.update(c=3, d=4)
+        assert d["c"] == 3
+        assert d["d"] == 4
+
+        # Update with iterable of pairs
+        d.update([("e", 5), ("f", 6)])
+        assert d["e"] == 5
+        assert d["f"] == 6
+
+    def test_performance_under_contention(self):
```
msusag, line 257: Will this test or other tests run for many seconds or more?  
That could slow down CI? 
Maybe some of them should be skipped by default, only enabled for manual runs?

cognite_toolkit/_cdf_tk/utils/safe_dict.py
msusag: I see you already say 
"It uses a threading lock to ensure that operations on the dictionary are atomic and do not lead to race conditions."
This does document the fact that it is single-threaded via the term "atomic", but could be made even more clear by saying that only one thread may enter either a read or write function at any point in time.

msusag:
@doctrino , please see my comments.

tests/test_unit/test_cdf_tk/test_utils/test_safe_dict.py
```diff
@@ -0,0 +1,283 @@
...
+
+        # Test clear
+        d.clear()
+        assert len(d) == 0
+        assert list(d.keys()) == []
+
+    def test_initialization_with_data(self):
+        """Test ThreadSafeDict can be initialized with data."""
+        # Test with dict
+        d1 = ThreadSafeDict({"a": 1, "b": 2})
```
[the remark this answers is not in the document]
doctrino: CICD pass so then it is good I think.

tests/test_unit/test_cdf_tk/test_utils/test_safe_dict.py
```diff
@@ -0,0 +1,283 @@
...
+import time
+from concurrent.futures import ThreadPoolExecutor, as_completed
+
+import pytest
+
+from cognite_toolkit._cdf_tk.utils.safe_dict import ThreadSafeDict
+
+
+class TestThreadSafeDict:
+    def test_basic_operations(self):
```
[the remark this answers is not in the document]
doctrino: Added. We do not (yet) have mypy checks for tests. We will at some point.

cognite_toolkit/_cdf_tk/utils/safe_dict.py
doctrino: done

cognite_toolkit/_cdf_tk/utils/safe_dict.py
doctrino: It is a good point, extended the docs. For now I think this is sufficent

ronpal (approved):

tests/test_unit/test_cdf_tk/test_utils/test_safe_dict.py
msusag: 4 ms is an acceptable runtime for the test :)

PR #1977, opened 2025-09-16
[CDF-25753] 👷‍♂️ Rework purge spaces part 1

doctrino:
# Description

This is part 1 of reimplementing the `cdf purge space` commands. It only includes the `dry-run` mode and is not connected to any command. The motivation for only doing the dry-run is to keep this PR small and fast to review.

The motivation for this reimplementation is as follows:
* The `/data-modeling/statistics` endpoint makes dry-run a lot faster to implement, but requires a new approach.
* We have new components such as HTTPClient and WorkerProducer that will be very useful for running the purge code.
* The current implementation lacks proper testing, a reimplementation will ensure it is well tested.

## Changelog

- [ ] Patch
- [x] Skip


cognite_toolkit/_cdf_tk/commands/_purge.py
```diff
@@ -111,6 +114,57 @@ def space(
...
+        if stats is None:
+            raise ToolkitMissingResourceError(f"Space {selected_space!r} does not exist")
+
+        # ValidateAuth
+        validator = ValidateAccess(client, "purge")
+        if include_space or (stats.containers + stats.views + stats.data_models) > 0:
+            validator.data_model(["read", "write"], spaces={selected_space})
+        if (stats.nodes + stats.edges) > 0:
+            validator.instances(["read", "write"], spaces={selected_space})
+
```
ronpal, line 148: Really like this utility/pattern

ronpal (approved):

haakonvt (commented):
🦄

PR #1981, opened 2025-09-18
🐛 API spec for extraction pipeline

doctrino:
# Description

Small bug in the YAML spec for extraction pipelines

## Changelog

- [x] Patch
- [ ] Skip

### Fixed

- When running `cdf build` Toolkit no longer give warnings on extraction pipelines for unused field sendNotification. For example: ` In contacts[1] unused field: 'sendNotification'`

## templates

No changes.


Aashutosh-cognite (approved):
🚀

PR #1978, opened 2025-09-16
[CDF-25753] 👷‍♂️ Rework purge spaces part 2

doctrino:
# Description

This is part 2 of reimplementing the `cdf purge space` commands.

The motivation for this reimplementation is as follows:
* The `/data-modeling/statistics` endpoint makes dry-run a lot faster to implement, but requires a new approach.
* We have new components such as HTTPClient and WorkerProducer that will be very useful for running the purge code.
* The current implementation lacks proper testing, a reimplementation will ensure it is well tested.

## Changelog

- [ ] Patch
- [x] Skip

PR #1984, opened 2025-09-18
🚨Warning message on purge

doctrino:
# Description

We had a user purging the space instead of instances, leading to data loss. This is a stop-gap solution to try to avoid this happening again. Once #1977 and #1978 are merged, I will add a better solution for it in the reimplementation of this command.

## Changelog

- [x] Patch
- [ ] Skip

### Improved

- When running the `cdf purge space` command, the user will now get an extra warning.

## templates

No changes.


Aashutosh-cognite (dismissed):
🏃🏻

Aashutosh-cognite (approved):

haakonvt (commented):
🦄

PR #1986, opened 2025-09-18
[prep CDF-25782] 😃 Prepare migration model change

doctrino:
# Description

This intgeration test (test running against CDF) needs to be skipped until the change of the model is done. Change is implemented in #1987 and will be activated again after that is merged. 

## Changelog

- [ ] Patch
- [x] Skip


rbtcollins (approved):
:unicorn:

PR #1991, opened 2025-09-19
[CDF-25770] 🪡Use Thread safe internal caching. 

doctrino:
# Description

The `AssetCentricMigrationIOAdapter` will be used in a multi-threading so need a thread-safe internal caching. 

## Changelog

- [ ] Patch
- [x] Skip


cognite_toolkit/_cdf_tk/commands/_migrate/adapter.py
```diff
@@ -113,47 +115,53 @@ def __init__(
...
+            item.get("space"),
+            item.get("externalId"),
+            item.get("instanceType"),
+        )
+        if isinstance(space, str) and isinstance(external_id, str) and isinstance(instance_type, str):
+            if instance_type == "node":
+                return NodeId(space=space, external_id=external_id)
+            elif instance_type == "edge":
+                return EdgeId(space=space, external_id=external_id)
+        return None
```
doctrino, line 141: Refactored out for clarity

cognite_toolkit/_cdf_tk/commands/_migrate/adapter.py
```diff
@@ -113,47 +115,53 @@ def __init__(
         self.instance = instance
         # Used to cache the mapping between instance IDs and asset-centric IDs.
         # This is used in the as_id method to track the same object as it is converted to an instance.
-        self._id_by_instance_id: dict[InstanceId, int] = {}
+        self._id_by_instance_id = ThreadSafeDict[InstanceId, int]()
```
doctrino, line 118: This is the change.

Aashutosh-cognite (dismissed):
🚀

cognite_toolkit/_cdf_tk/commands/_migrate/adapter.py
```diff
@@ -113,47 +115,52 @@ def __init__(
...
     def as_id(self, item: dict[str, JsonVal] | object) -> int:
         if isinstance(item, AssetCentricMapping):
             instance_id = item.mapping.instance_id
-            id_ = item.mapping.id
             if instance_id not in self._id_by_instance_id:
-                self._id_by_instance_id[instance_id] = id_
-            return id_
-        elif isinstance(item, Event | Asset | TimeSeries | FileMetadata):
+                self._id_by_instance_id.setdefault(instance_id, item.mapping.id)
+            return item.mapping.id
```
sighol, line 148: These operations are not thread safe together.
Let's say that the `as_id` function is called simultaneously by two threads with a conflicting mapping:
- Thread A: `{"id": 1, "instanceId": {"space": "a", "externalId:"a","instanceType":"node"}}`
- Thread B: `{"id": 2, "instanceId": {"space": "a", "externalId:"a","instanceType":"node"}}`

Thread A will get `1` returned while Thread B will get `2`.

You could instead write it as
```python
instance_id = item.mapping.instance_id
self._id_by_instance_id.setdefault(instance_id, item.mapping.id)
return self._id_by_instance_id[instance_id]
```
Then both threads will get the same value back, which then would be a bit unexpected for one of the threads.

The reason I am curious about this, is that most of the operations that you do on dictionaries here are already thread safe in CPython because of GIL. That is `key in dict` and `value = dict[key]`. I am not sure if `dict[key] = value` and `dict.setdefault(key, value)` are atomic when the dictionary is growing.

But anyway, the confusing part of the code is that you are using a thread safe dictionary, but still using it in a way that is not thread safe. Do you expect that two threads would try to write conflicting instance_id->id mappings? Does that need to be handled in some way and reported on? Or are you using a thread safe dict only to be safe when the dictionary grows? Just a sentence about why it needs to be thread safe would suffice.

Very sorry for the wall of text.

doctrino: I appreciate the wall :). I do not expect multiple write for the same ID. In principle, this code would probably be fine without any thread safety data structure. What motivated me to add it is that this function will be called from multiple threads, and my thinking that I should use a thread safe data structure. 

The use case is that the item first writes to this dict, then later in the pipeline, a different thread will look it up.

sighol: @doctrino Maybe also add a comment that you don't expect multiple threads to write to the same keys, but that you added a thread safe dictionary just in case?

sighol (commented):
🦄

PR #1997, opened 2025-09-20
[CDF-25733]🏗Refactor MigrationMappingList

doctrino:
# Description

This PR refactors the `MigrationMappingList` to use the `ModelList` base class.

## Changelog

- [ ] Patch
- [x] Skip


cognite_toolkit/_cdf_tk/commands/_migrate/data_classes.py
```diff
@@ -59,7 +48,24 @@ def get_ingestion_view(self) -> str:
     def as_asset_centric_id(self) -> AssetCentricId:
         return AssetCentricId(resource_type=self.resource_type, id_=self.id)  # type: ignore[arg-type]
 
-    @field_validator("data_set_id", "ingestion_view", mode="before")
+    @model_validator(mode="before")
```
doctrino, line 51: We now need to move this to the pydantic class used to validate each row, instead of having it on the list class.

kornelione:
🦄

PR #2000, opened 2025-09-21
[CDF-25801] GraphQL YAML

doctrino:
# Description

This PR introduces the `GraphQLDataModelYAML` to validate graphql data model configurations. Note this is not yet exposed, so no change to the end user. Note this is a hidden alpha feature only used by two customers, but to complete the transformation to use pydantic for validation of configurations we also need to cover this.

## Changelog

- [ ] Patch
- [x] Skip


rsjr (approved):
risk review ok

PR #2002, opened 2025-09-22
[CDF-25805] 😅 Migration create properties

doctrino:
# Description

This introduces a new improved version of creating properties for an instance from an asset-centric resource. Instead of using a nested structure for the mapping from asset-centric property to instance property,  I have replaced it with a simple `dict[str, str]` that uses json path on the key.

REVIEWER **This is 62 lines of code with one 28 line helper function. The rest is tests**

## Changelog

- [ ] Patch
- [x] Skip


Aashutosh-cognite (approved):
LGTM

PR #2008, opened 2025-09-24
[CDF-25825] Support keyvault directives in multiline YAML strings

ronpal:
# Description

- Updated regex pattern in ExtractionPipelineConfigCRUD to handle !keyvault
  directives anywhere in the config string, not just after colons
- Changed from r':\s+!(\w+)' to r'!(\w+)' to support multiline connection
  strings with keyvault references like 'UID=!keyvault SECRET-ID'
- Added test case for multiline connection string scenario
- Fixes issue where keyvault directives in folded YAML blocks were not
  being processed correctly
## Changelog

- [x] Patch
- [ ] Skip

### Fixed

- KeyVault directive resolution in multiline config

## templates

No changes.


ronpal:
The pattern is not supported by extractors. Hence, closing this PR as it it would just push the problem downstream.

PR #2009, opened 2025-09-24
[CDF-25826] Dump proper error message on no selection

Aashutosh-cognite:
# Description

User encountered an error while dumping transformations.
*Before* 
In case of multi-select, user need to press `space` key to do the selection. But if user directly presses the `enter` key, it shows a success message with no dumped file.

Screenshot 2025-09-24 at 3 47 06 PM

Screenshot 2025-09-24 at 3 45 45 PM

*After*
Now if user directly presses the `enter` key, it will fail with error message along with navigation and selection helper.

Screenshot 2025-09-24 at 3 51 45 PM

This PR also contains similar check for other resources as well.

## Changelog

- [x] Patch
- [ ] Skip

### Improved

- Improved handling and error message on `cdf dump transformation` (other resources as well) incase of no selection.

## templates

No changes.

### DIFF
--- a/cognite_toolkit/_cdf_tk/commands/dump_resource.py
+++ b/cognite_toolkit/_cdf_tk/commands/dump_resource.py
@@ -300,8 +300,10 @@ class TransformationFinder(ResourceFinder[tuple[str, ...]]):
             "Which transformation(s) would you like to dump?",
             choices=choices,
         ).ask()
-        if selected_transformation_ids is None:
-            raise ToolkitValueError("No transformations selected for dumping.")
+        if not selected_transformation_ids:
+            raise ToolkitValueError(
+                "No transformations selected for dumping. Use arrow keys to navigate and space key to select. Press enter to confirm."
+            )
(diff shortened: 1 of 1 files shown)


cognite_toolkit/_cdf_tk/commands/dump_resource.py
```diff
@@ -300,8 +300,10 @@ def _interactive_select(self) -> tuple[str, ...]:
             "Which transformation(s) would you like to dump?",
             choices=choices,
         ).ask()
-        if selected_transformation_ids is None:
-            raise ToolkitValueError("No transformations selected for dumping.")
+        if not selected_transformation_ids:
```
Aashutosh-cognite, line 303: this is the bug fix

doctrino (approved):
Great. You can optionally do what Gemini suggest.

doctrino (approved):

rsjr (approved):
risk review ok

PR #2010, opened 2025-09-24
[prep CDF-25832] 🚩Look-up DM only flag

doctrino:
# Description

Adds 

## Changelog

- [ ] Patch
- [x] Skip

### DIFF
--- a/cognite_toolkit/_cdf_tk/client/_toolkit_client.py
+++ b/cognite_toolkit/_cdf_tk/client/_toolkit_client.py
@@ -14,6 +14,7 @@ from .api.extended_raw import ExtendedRawAPI
 from .api.extended_timeseries import ExtendedTimeSeriesAPI
 from .api.lookup import LookUpGroup
 from .api.migration import MigrationAPI
+from .api.project import ProjectAPI
 from .api.robotics import RoboticsAPI
 from .api.search import SearchAPI
 from .api.token import TokenAPI
@@ -133,6 +134,8 @@ class ToolkitClient(CogniteClient):
         self.migration = MigrationAPI(self.data_modeling.instances)
         self.token = TokenAPI(self)
         self.charts = ChartsAPI(self._config, self._API_VERSION, self)
+        if config is not None:
+            self.project = ProjectAPI(config=config)
 
     @property
     def config(self) -> ToolkitClientConfig:
--- a/cognite_toolkit/_cdf_tk/client/api/project.py
+++ b/cognite_toolkit/_cdf_tk/client/api/project.py
@@ -1,0 +1,23 @@
+from typing import TYPE_CHECKING
+
+from cognite_toolkit._cdf_tk.utils.http_client import HTTPClient, ParamRequest
+
+if TYPE_CHECKING:
+    from cognite_toolkit._cdf_tk.client import ToolkitClientConfig
+
+
+class ProjectAPI:
+    def __init__(self, config: ToolkitClientConfig) -> None:
+        self._config = config
+        self._http_client = HTTPClient(config, split_items_status_codes=set())
+
+    def info(self) -> dict[str, str]:
+        """Retrieve information about the current project."""
+        response = self._http_client.request_with_retries(
+            ParamRequest(
+                endpoint_url=self._config.create_api_url(""),
+                parameters={"withDataModelingStatus": "true"},
+                method="GET",
+            )
+        )
+        return response.json()


Aashutosh-cognite (dismissed):
🚀

cognite_toolkit/_cdf_tk/client/api/project.py
```diff
@@ -0,0 +1,25 @@
...
+    def status(self) -> ProjectStatus:
+        """Retrieve information about the current project."""
+        response = self._http_client.request_with_retries(
+            ParamRequest(
+                endpoint_url=f"{self._config.base_url}/api/v1/projects?withDataModelingStatus=true", method="GET"
+            )
+        )
+        response.raise_for_status()
+        body = response.get_first_body()
+        # We expect the API will always return exactly one project when querying by project name
```
thorkildcognite, line 24: This isn't the case. You are calling /api/v1/projects (the list operation), and not the get on the specific project (which I couldn't get to work right now). So, you are not guaranteed that there is 1 hit, and not what is first.

Example query I just ran using one of my service principals (submit-query is just an auth wrapper, but I think you see what it is doing):


``` 
$ ./submit-query.py "/api/v1/projects?withDataModelingStatus=true"
{
    "items": [
        {
            "urlName": "prodverification",
            "dataModelingStatus": "HYBRID"
        },
        {
            "urlName": "thorkild-demo",
            "dataModelingStatus": "HYBRID"
        }
    ]
}
```

So I think you will have to filter the output, or find some way of using /api/v1/projects/<project name> , but I couldn't get that to work with the withDataModelingStatus flag.

cognite_toolkit/_cdf_tk/client/_toolkit_client.py
```diff
@@ -133,6 +134,8 @@ def __init__(self, config: ToolkitClientConfig | None = None, enable_set_pending
         self.migration = MigrationAPI(self.data_modeling.instances)
         self.token = TokenAPI(self)
         self.charts = ChartsAPI(self._config, self._API_VERSION, self)
+        if config is not None:
+            self.project = ProjectAPI(config=config, cognite_client=self)
```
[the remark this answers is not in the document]
doctrino: I will also require the Toolkit client to be set in the future, so I have captured that in another task: https://cognitedata.atlassian.net/browse/CDF-25853

cognite_toolkit/_cdf_tk/client/api/project.py
```diff
@@ -0,0 +1,27 @@
...
+    def status(self) -> ProjectStatusList:
+        """Retrieve information about the current project."""
+        response = self._http_client.request_with_retries(
+            ParamRequest(
+                endpoint_url=f"{self._config.base_url}/api/v1/projects?withDataModelingStatus=true", method="GET"
+            )
+        )
+        response.raise_for_status()
+        body = response.get_first_body()
+        return ProjectStatusList._load(body["items"], cognite_client=self._cognite_client)  # type: ignore[arg-type]
```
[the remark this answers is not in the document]
doctrino: This is following the style of the CogniteSDK and it is the expected response of the API, so we will keep it as is.

cognite_toolkit/_cdf_tk/client/api/project.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class ProjectAPI:
+    def __init__(self, config: "ToolkitClientConfig", cognite_client: CogniteClient) -> None:
+        self._config = config
+        self._cognite_client = cognite_client
+        self._http_client = HTTPClient(config, split_items_status_codes=set())
+
+    def status(self) -> ProjectStatusList:
+        """Retrieve information about the current project."""
```
thorkildcognite, line 19: Nit (non-blocking): This is inaccurate. It now returns list of states of the projects the user has access to, and then the user calls this_project to find the particular one to get the correct project. It is tempting to do the filtering automatically already here, but then there wouldn't be a way to access all the other projects you have access to, which can be useful, so I would instead change the docstring. That can be done in a follow-up, though.

thorkildcognite (approved):
🦄

cognite_toolkit/_cdf_tk/client/api/project.py
doctrino: It is good point, will do it in a follow up, as this is internal and not visible to the und-user.

ronpal (approved):

PR #2016, opened 2025-09-25
Reactivate tests

doctrino:
# Description

After changing the migration model in #1987, we had to skip a few tests. This reactivates the tests again as the new migration model can be deployed to our test project.

## Changelog

- [ ] Patch
- [x] Skip


thorkildcognite (commented):
🦄

PR #2020, opened 2025-09-25
[CDF-25782] 🛁 Annotation mappings

doctrino:
# Description

This PR adds the mapping from file and asset annotations to CogniteDiagramAnnotation.

## Changelog

- [ ] Patch
- [x] Skip


olacognite (commented):
It is not very clear to me how these mappings will be used. Some observations though: 

I believe the annotationType actually is the same as the edge.type.externalId, at least for diagrams.AssetLink and diagrams.FileLink. (I think these two are the ones you are targetting now?)

How to use the source fields in datamodelling is not quite clear yet. My team has so far been using sourceId in datamodelling as a free string field to indicate that CogniteDiagramAnnotations have been created by our service. This might be used differently by others. 

data.assetRef.id and externalId are of course the classic ids, they somehow correspond to the nodeId of the endNode. How does the mapping deal with this? 

I see that you are flipping the direction of the edge for file annotations. I have not seen anyone do this before. A file annotation is like an asset annotation, but instead of referencing an asset in a file, it references a file in a file. I believe the convention also going forward that the file that mentions another resource whether it is asset or file will be the start node. 

The idea behind the endNode region is that two P&IDs that have line continuation boxes that point to each other could in theory use a single annotation, but it is a bit clunky to make this work, and I am not aware that it is in use.

olacognite:
> > How to use the source fields in datamodelling is not quite clear yet. My team has so far been using sourceId in datamodelling as a free string field to indicate that CogniteDiagramAnnotations have been created by our service. This might be used differently by others.
> 
> This is a suggestion by me, for migration, we have to make a decision here. My thinking is that an application is a type of SourceSystem and that the tooling could automatically create one instance. However, I would like to do it as you are doing the annotation team, so I have updated the PR.

For the classical annotations, it is assumed that either creating_app or creating_user is set, but if there is a way to reference either through an actual source node, it would have been ideal I think. What you suggest should work, but might be worth discussing how source information is supposed to be provided in general in DM. 

> > The idea behind the endNode region is that two P&IDs that have line continuation boxes that point to each other could in theory use a single annotation, but it is a bit clunky to make this work, and I am not aware that it is in use.
> 
> Got it, so we just ignore it for now?

Yep

erlendvollset (commented):
🦄

PR #2022, opened 2025-09-26
[CDF-25769] 🤩 Added validation to migration command.

doctrino:
# Description

**Context**: This is part of the `alpha` migration tooling in Toolkit, for overview [see here](https://cognitedata.atlassian.net/wiki/spaces/~170890702/pages/5455773697/Migration+Tooling+in+Toolkit). This is step 3 and4 of that process. 

In Toolkit, a command class contains the logic that is executed when a CLI command is called.

This PR is a copy-paste of the validation logic from the `BaseMigrateCommand` to `MigrationCommand`. Currently, there is a one migration command for each resource type the Toolkit supports. We are in the process of replacing all of these with a single command, `MigrationCommand`. that contains the logic for migrating any resource and uses a implementations of a shared interface for each resource type. Once the `MigrationCommand` is finished, I will replace all the other migration commands.

## Changelog

- [ ] Patch
- [x] Skip

### DIFF
--- a/cognite_toolkit/_cdf_tk/commands/_migrate/command.py
+++ b/cognite_toolkit/_cdf_tk/commands/_migrate/command.py
@@ -5,16 +5,25 @@ from pathlib import Path
 from rich.console import Console
 from rich.table import Table
 
+from cognite_toolkit._cdf_tk.client import ToolkitClient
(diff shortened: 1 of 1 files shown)


ronpal (dismissed):

haakonvt:
🦄

PR #2023, opened 2025-09-26
[CDF-25848] allow modules prefix in packages.toml

ronpal:
# Description

Giving package authors some leeway if they add "modules/" to their package toml manifests because it can be confusing whether to include or not. 

allowed: 

[packages.quickstart]
title = "Quickstart"
description = "Get started with Cognite Data Fusion in minutes."
canCherryPick = false
modules = [
    "cdf_ingestion",
    "sourcesystem/cdf_pi",
    "sourcesystem/cdf_sap_assets",
    "sourcesystem/cdf_sap_events",
    "sourcesystem/cdf_sharepoint",
    "models/cdf_process_industry_extension",
    "contextualization/cdf_connection_sql",
    "contextualization/cdf_p_and_id_parser",
    "industrial_tools/cdf_search"
]

also allowed:

[packages.quickstart]
title = "Quickstart"
description = "Get started with Cognite Data Fusion in minutes."
canCherryPick = false
modules = [
    "modules/cdf_ingestion",
    "modules/sourcesystem/cdf_pi",
    "modules/sourcesystem/cdf_sap_assets",
    "modules/sourcesystem/cdf_sap_events",
    "modules/sourcesystem/cdf_sharepoint",
    "modules/models/cdf_process_industry_extension",
    "modules/contextualization/cdf_connection_sql",
    "modules/contextualization/cdf_p_and_id_parser",
    "modules/industrial_tools/cdf_search"
]

## Changelog

- [x] Patch
- [ ] Skip

### Improved

- leading "modules/" segment is now optional in packages.toml

## templates

No changes.

### DIFF
--- a/cognite_toolkit/_cdf_tk/data_classes/_packages.py
+++ b/cognite_toolkit/_cdf_tk/data_classes/_packages.py
@@ -90,10 +90,11 @@ class Packages(dict, MutableMapping[str, Package]):
                         module_path_obj = Path(module_path)
 
                         # Remove "modules/" prefix if present to normalize user input
-                        if module_path_obj.parts and module_path_obj.parts[0] == "modules":
-                            normalized_path = Path(*module_path_obj.parts[1:])
-                        else:
-                            normalized_path = module_path_obj
+                        normalized_path = (
+                            Path(*module_path_obj.parts[1:])
+                            if module_path_obj.parts and module_path_obj.parts[0] == "modules"
+                            else module_path_obj
+                        )
 
                         # Try to find module by normalized path, then by adding "modules/" prefix
                         if (module := module_by_relative_path.get(normalized_path)) is None:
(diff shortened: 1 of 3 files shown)


cognite_toolkit/_cdf_tk/data_classes/_packages.py
```diff
@@ -87,7 +87,23 @@ def load(
...
+
+                        # Try to find the module with the path as is.
+                        module = module_by_relative_path.get(module_path_obj)
+
+                        if not module:
+                            # If not found, try normalizing the 'modules/' prefix.
+                            if module_path_obj.parts and module_path_obj.parts[0] == "modules":
+                                # The path has 'modules/' prefix, so try without it.
+                                path_without_prefix = Path(*module_path_obj.parts[1:])
+                                module = module_by_relative_path.get(path_without_prefix)
```
finnag, line 100: I'd prefer if you still failde here, but printed a warning about what could possibly work instead.

ronpal:
> Can you explain why this is necessary? Allowing multiple forms of the source code seems like bad idea.
> 
> In particular, would it not be better to warn about it, and suggest "did you mean " instead?

@finnag important point here is that the user of the Toolkit cannot influence the content of the package, since it is imported from an external source. So if the package.toml is "malformed", there's no use warning the user about it as they are not able to do anything about it. 

I agree it is unfortunate to have this alternative form in the code, I would prefer not to. But we are optimizing for the user, not the coder.

finnag:
@ronpal book a meeting with me or contact me on slack if you want a faster discussion here

PR #2027, opened 2025-09-28
[CDF-25832] 🧑‍🚀 Support CogniteFile in Function deployment

doctrino:
# Description

Starting to change the behavior of Toolkit depending on project type. 

Note that Toolkit is currently depending on a private method in the `cognite-sdk`, so I needed to refactor that part before introducing the new functionality.

## Changelog

- [x] Patch
- [ ] Skip

### Added

- You can now use `CogniteFile` instead of asset-centric file when deploying a Function. You do this by specifying a `space` in the function configuration

### Fixed

- Data-modeling only projects, require you to use `CogniteFile` instead of asset-centric file when deploying a Function. 

## templates

No changes.


einarmo (commented):
:unicorn: - looks fine as a change, but the PR title is pretty misleading. You say in the description that this is a preliminary, but the title should probably have reflected that too.

PR #2031, opened 2025-09-29
[CDF-25832] 🧑‍🚀 Support CogniteFile in Function deployment

doctrino:
# Description

TODO

## Changelog

- [x] Patch
- [ ] Skip

### Added

- You can now use `CogniteFile` instead of asset-centric file when deploying a Function. You do this by specifying a `space` in the function configuration

### Fixed

- Data-modeling only projects, require you to use `CogniteFile` instead of asset-centric file when deploying a Function. 

## templates

No changes.


Aashutosh-cognite (approved):
LGTM

PR #2033, opened 2025-09-29
[CDF-25858] Enable statistics on deployment packs

ronpal:
# Description

Maintainers of external module (a.k.a. deployment pack) libraries want to track the usage through the Toolkit in Mixpanel. They will want to know when deployment packs are _installed_, _built_ and _deployed_.

This PR

- extends module data-classes so that it is possible to tag them with identifiers (package and module)
- extends tracking in`ToolkitCommand` so that its subclasses can have additional values added to the emitted Mixpanel event.
- uses this approach in the BuildCommand to include package and module ids to the Mixpanel event

Screenshot 2025-10-08 at 10 27 28

Screenshot 2025-10-08 at 16 06 36

## Changelog

- [x] Patch
- [ ] Skip

### Added

- Statistics on Deployment Pack installation

## templates

No changes.


cognite_toolkit/_cdf_tk/commands/build_cmd.py
```diff
@@ -323,8 +326,20 @@ def build_modules(
                     warning_count=module_warnings,
                     status=built_status,
                     iteration=iteration,
+                    module_id=module.module_id,
+                    package_id=module.package_id,
                 )
                 build.append(built_module)
+
+                if module.package_id:
+                    package_ids = self._additional_tracking_info.setdefault("package_ids", [])
```
doctrino, line 335: I would camelCase for consistency with other mixpanel attributes.

doctrino (commented):
I struggle to find those events in our mixpanel, so makes it hard to test the follwing

* Can you split on list values in Mixpanel? Or is it simply treated as a string?  `modules_ids: ["agent","data_model"]`.
* What happens if there are multiple `package_id` in the same build? Can you build the dashboards you need to build using these events?

ronpal:
> I struggle to find those events in our mixpanel, so makes it hard to test the follwing
> 
> * Can you split on list values in Mixpanel? Or is it simply treated as a string?  `modules_ids: ["agent","data_model"]`.
> * What happens if there are multiple `package_id` in the same build? Can you build the dashboards you need to build using these events?

Resolved in call. Screenshot added as proof.

PR #2034, opened 2025-09-29
[CDF-25198] 😁 Improved purge space part 4 (final)

doctrino:
# Description

This is part 4 of reimplementing the `cdf purge space` commands. This flips the purge application to use the new command.

The motivation for this reimplementation is as follows:
* The `/data-modeling/statistics` endpoint makes dry-run a lot faster to implement, but requires a new approach.
* We have new components such as HTTPClient and WorkerProducer that will be very useful for running the purge code.
* The current implementation lacks proper testing, a reimplementation will ensure it is well tested.

**NOTE** This is technically a breaking change, we no longer allow the user to delete nodes with datapoins/file content, however, we have had incidents with the current version, thus I suggest we release this immediately instead of waiting for the next minor.

## Changelog

- [x] Patch
- [ ] Skip

### Improved

- The `cdf purge space` command has been reworked. The command now has an improved user experience for selecting space. It is a log faster when running in `--dry-run`, and it does not delete timeseries datapoints and files by default anymore.
 
## templates

No changes.


Aashutosh-cognite (approved):
🚀

PR #1994, opened 2025-09-20
😁Use new YAML classes

doctrino:
# Description

These classes are already implemented and well tested. Setting the yaml_cls in the CRUD interfac will start using this in the build command to validate the YAML syntax of the user. The motivation for doing this in a separate PR is to avoid merge conflicts. 

This is the same type of PR as for example #1944 

## Details:
The testing of the YAML resource classes we use to validate the syntax of the user configurations: https://github.com/cognitedata/toolkit/tree/main/tests/test_unit/test_cdf_tk/test_resource_classes

Then, by setting yaml_cls in the resource CRUDit will be used in the cdf build command to give the user a warning about whether they follow the syntax here: https://github.com/cognitedata/toolkit/blob/main/cognite_toolkit/_cdf_tk/commands/build_cmd.py#L603

## Changelog

- [x] Patch
- [ ] Skip

### Improved

- Warning on YAML syntax errors when running cdf build for resources types hosted extractors source and jobs, relationships, all robotic resources, sequences, data models, nodes and edges.

## templates

No changes.

PR #2045, opened 2025-10-02
[CDF-25781] 😁 Use YAML classes

doctrino:
# Description

These classes are already implemented and well tested. Setting the yaml_cls in the CRUD interfac will start using this in the build command to validate the YAML syntax of the user. The motivation for doing this in a separate PR is to avoid merge conflicts. 

This is the same type of PR as for example #1944, #1994

## Details:
The testing of the YAML resource classes we use to validate the syntax of the user configurations: https://github.com/cognitedata/toolkit/tree/main/tests/test_unit/test_cdf_tk/test_resource_classes

Then, by setting yaml_cls in the resource CRUDit will be used in the cdf build command to give the user a warning about whether they follow the syntax here: https://github.com/cognitedata/toolkit/blob/main/cognite_toolkit/_cdf_tk/commands/build_cmd.py#L603

## Changelog

- [x] Patch
- [ ] Skip

### Improved

- Warning on YAML syntax errors when running cdf build for resources types sequence rows, graphql model (alpha feature), cognite file and resource mapping (alpha feature)

## templates

No changes.


sighol (commented):
🦄

PR #2050, opened 2025-10-06
[CDF-25800] 🤖 AgentYAML class

doctrino:
# Description

This PR introduces the `AgentYAML` to validate agents configurations. Note this is not yet exposed, so no change to the end user. 

## Changelog

- [ ] Patch
- [x] Skip


Aashutosh-cognite (approved):
🚀

PR #2051, opened 2025-10-07
[CDF-25802] 😃 Infield v1 YAML

doctrino:
# Description

This PR introduces the `InfieldV1YAML` to validate infield configurations. Note this is not yet exposed, so no change to the end user. Also note that the InfieldV1 resource is an alpha feature. 

## Changelog

- [ ] Patch
- [x] Skip


Aashutosh-cognite (approved):
🚀

PR #2053, opened 2025-10-08
[CDF-25907] 🐛 Iterate instances

doctrino:
# Description

We have a small bug when iterating over all instances. The builtin index is on space + externalId and not just `externalId`.

## Changelog

- [x] Patch
- [ ] Skip

### Improved

- Faster downloading of instances when you run `cdf purge space`. 

## templates

No changes.


Aashutosh-cognite (approved):
🚀

PR #2456, opened 2026-02-03
[CDF-26977] 🐌Unit conversion in migration

doctrino:
# Description

Turns out this was not a bug - Toolkit did it correctly. There is however room for better error messages.

## Bump

- [ ] Patch
- [x] Skip


doctrino:
Not a real issue

PR #2831, opened 2026-04-09
[CDF-27646] 🩸 Support upload raw table deploy v2

doctrino:
# Description

Support for uploading RAW tables in deploy v2. 

I did this in two steps:
1. Added identifier, required a small trick see comment in code.
2. Find all raw tables in the linage yaml and call the UploadCommand

image

## Bump

- [ ] Patch
- [x] Skip

### DIFF
--- a/cognite_toolkit/_cdf_tk/commands/build_v2/data_classes/_lineage.py
+++ b/cognite_toolkit/_cdf_tk/commands/build_v2/data_classes/_lineage.py
@@ -58,6 +58,10 @@ class ResourceLineageItem(_BaseLineageModel):
         resource_type: ResourceType = info.data["type"]
         return resource_type.load_identifier(value)
 
+    @field_serializer("identifier", when_used="always")
+    def serialize_identifier(self, value: Identifier) -> dict[str, Any]:
+        return value.dump()
+
 
 class ModuleLineageItem(_BaseLineageModel):
     """Tracks a module through the build process."""
--- a/cognite_toolkit/_cdf_tk/commands/build_v2/data_classes/_lineage.py
+++ b/cognite_toolkit/_cdf_tk/commands/build_v2/data_classes/_lineage.py
@@ -15,6 +15,7 @@ from pydantic import (
 )
 from pydantic.alias_generators import to_camel
 
+from cognite_toolkit._cdf_tk.client._resource_base import Identifier
 from cognite_toolkit._cdf_tk.commands.build_v2.data_classes._build import BuildFolder, BuiltModule
 from cognite_toolkit._cdf_tk.commands.build_v2.data_classes._insights import (
     ConsistencyError,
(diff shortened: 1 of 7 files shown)


cognite_toolkit/_cdf_tk/commands/build_v2/data_classes/_lineage.py
```diff
@@ -42,6 +45,22 @@ class ResourceLineageItem(_BaseLineageModel):
...
+        if not isinstance(value, dict):
+            return value
+        if "type" not in info.data:
+            raise ValueError("Resource type must be provided to load identifier")
+        resource_type: ResourceType = info.data["type"]
+        return resource_type.load_identifier(value)
+
+    @field_serializer("identifier", when_used="always")
+    def serialize_identifier(self, value: Identifier) -> dict[str, Any]:
+        return value.dump()
```
doctrino, line 63: Pydantic does not deal well with subclasses. Thus, using the resource type to get the crud to load the identifier correctly. This ensures that we for example handle ExternalId and TaskId correctly, even though they both are stored as 

```
identifier:
  externalId: my_id
```

cognite_toolkit/_cdf_tk/commands/deploy_v2/command.py
```diff
@@ -197,13 +201,12 @@ def deploy(
         options: DeployOptions | None = None,
     ) -> Sequence[DeploymentResult]:
         options = options or DeployOptions(environment_variables=env_vars.dump())
-        build_dir = self.read_build_directory(user_build_dir, options.include)
+        build_lineage = self.read_build_lineage(user_build_dir)
```
doctrino, line 204: Loading linage at the top level.

cognite_toolkit/_cdf_tk/commands/deploy_v2/command.py
```diff
@@ -883,3 +896,40 @@ def _display_results(
                 for skip in total.skipped
             ]
             console.print(Panel("\n".join(skipped_str), title="Skipped resources", expand=False))
+
+    @classmethod
+    def _find_raw_tables(cls, build_lineage: BuildLineage) -> Mapping[RawTableSelector, list[Path]]:
+        selections: dict[RawTableSelector, list[Path]] = defaultdict(list)
```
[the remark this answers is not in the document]
doctrino: I don't like this solution as it hides the fact that this will always be a RAWTableSelector.

cognite_toolkit/_cdf_tk/commands/deploy_v2/command.py
```diff
@@ -883,3 +896,40 @@ def _display_results(
                 for skip in total.skipped
             ]
             console.print(Panel("\n".join(skipped_str), title="Skipped resources", expand=False))
+
+    @classmethod
+    def _find_raw_tables(cls, build_lineage: BuildLineage) -> Mapping[RawTableSelector, list[Path]]:
```
ronpal, line 901: Maybe a comment about why raw gets this special treatment in an otherwise clean code, and that this method should be removed once the feature has been sunset?

ronpal (approved):

PR #2838, opened 2026-04-10
[CDF-27591]🔧 Quality-of-life Deploy v2

doctrino:
# Description

Fixes the following two issues

## Issue 1: Warning about files that are OK

Today we warn about file in functions/streamlit folders - even though this is OK.
image

## Issue 2: All zeros in result table

We have two CRUD classes for Groups, so need to ensure both are included. Currently, we only include groups with scoped resources.

image

### FIX

We now deploy all scoped groups correctly.

image

## Bump

- [ ] Patch
- [x] Skip

### DIFF
--- a/tests/test_unit/test_cli/test_build_deploy_snapshots/complete_org_v2.yaml
+++ b/tests/test_unit/test_cli/test_build_deploy_snapshots/complete_org_v2.yaml
@@ -362,6 +362,17 @@ FunctionScheduleResponse:
   nonce: dummy-nonce
 GroupResponse:
 - capabilities:
+  - projectsAcl:
+      actions:
+      - LIST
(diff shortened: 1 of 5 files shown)


tests/test_unit/test_cli/test_build_deploy_snapshots/complete_org_v2.yaml
```diff
@@ -361,6 +361,17 @@ FunctionScheduleResponse:
   name: daily-8am-utc
   nonce: dummy-nonce
 GroupResponse:
+- capabilities:
```
doctrino, line 364: This is the bugfix, we now deploy all scoped groups.

ronpal (approved):

PR #2847, opened 2026-04-12
[CDF-24986] 😮‍💨Charts migration with backend services.

doctrino:
# Description

## Bump

- [x] Patch
- [ ] Skip

### Added

- [alpha] Support for migrating Charts with monitoring jobs and scheduled calculations.


cognite_toolkit/_cdf_tk/client/resource_classes/charts_data.py
```diff
@@ -130,21 +133,8 @@ class ChartThreshold(ChartElement):
     calls: list[ChartCall] | None = None
 
 
-class ChartScheduledCalculationUIElement(ChartElement):
-    color: str | None = None
-    created_at: int | None = None
+class ChartScheduledCalculationUIElement(ChartCalculation):
```
doctrino, line 136: The ChartWorkflowUIElement and ChartScheduledCalculationUIElement are the same elements in the UI, only difference is description/calls. Thus combining these to the same base class.

cognite_toolkit/_cdf_tk/commands/_migrate/data_mapper.py
```diff
@@ -306,6 +312,19 @@ def _populate_cache(self, source: Sequence[ChartResponse]) -> dict[str, set[int]
...
+                    timeseries_ids.add(model.timeseries_id)
+                if model.timeseries_external_id:
+                    timeseries_external_ids.add(model.timeseries_external_id)
+            for calculation in chart.scheduled_calculations or []:
+                if calculation.target_timeseries_external_id:
+                    timeseries_external_ids.add(calculation.target_timeseries_external_id)
+                for step in calculation.graph.steps:
+                    for inp in step.inputs:
+                        if inp.type == "ts" and isinstance(inp.value, str):
+                            timeseries_external_ids.add(inp.value)
```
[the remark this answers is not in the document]
doctrino: This cannot be an internal ID

cognite_toolkit/_cdf_tk/commands/_migrate/data_mapper.py
```diff
@@ -336,13 +352,7 @@ def map(self, source: Sequence[ChartResponse]) -> Sequence[ChartRequest | None]:
...
         for item in source:
             identifier = item.external_id
-            if item.data.scheduled_calculation_collection or item.data.monitoring_jobs:
-                # These are not yet supported.
-                output.append(None)
-                issues.append(self._create_missing_support_issue(item))
-                self.logger.tracker.finalize_item(identifier, "failure")
-                continue
-
+            # Todo: Check owner of MonitoringJobs (probably not for scheduled calculations)
```
Magssch: Should this be addressed before merging or is it for later?

cognite_toolkit/_cdf_tk/commands/_migrate/data_mapper.py
Magssch:
```suggestion
            ids: The charts Ids (InternalId or ExternalId) to look up.
```

cognite_toolkit/_cdf_tk/commands/_migrate/data_mapper.py
doctrino: Needed to do some manual testing to confirm behavior. Interesting, this is only for monitoring jobs, not for scheduled calculations. Done now

cognite_toolkit/_cdf_tk/commands/_migrate/data_mapper.py
```diff
@@ -330,16 +347,23 @@ def prepare(self, source_selector: ChartSelector) -> None:
...
+
     def map(self, source: Sequence[ChartResponse]) -> Sequence[ChartRequest | None]:
         event_ids_by_chart_external_id = self._populate_cache(source)
         output: list[ChartRequest | None] = []
         issues: list[ChartMigrationIssue] = []
         for item in source:
             identifier = item.external_id
-            if item.data.scheduled_calculation_collection or item.data.monitoring_jobs:
-                # These are not yet supported.
-                output.append(None)
```
Magssch, line 341: Should we keep this maybe? I think removing this causes a desync since this is used in a zip() method later

tests_smoke/test_commands/test_migrate_chart/test_migrate_data.yml
```diff
@@ -15,12 +15,217 @@ chart:
...
+    externalId: smoke_test_chart_migrration_monitoring_job
+    interval: 300000
+    metadata: null
+    model:
+      externalId: double_threshold
+      granularity: 1m
+      lowerThreshold: null
+      timeseriesExternalId: null
+      timeseriesId: 1356054795741013
+      timeseriesInstanceId: null
```
Magssch, line 183: Could this being `null` indicate an issue?

doctrino: It is a good observation. It turns out this is what the server returns even when you set `timeseriesInstanceId`, so it is correct as it is.

cognite_toolkit/_cdf_tk/commands/_migrate/data_mapper.py
doctrino: Which zip method? I cannot see that.

Magssch: I was thinking of the one in MigrationCommand._convert in command.py on line 304: 
```python
    @staticmethod
    def _convert(
        mapper: DataMapper[T_Selector, T_DataResponse, T_DataRequest],
    ) -> Callable[[Page[T_DataResponse]], Page[T_DataRequest]]:
        def track_mapping(source: Page[T_DataResponse]) -> Page[T_DataRequest]:
            raw_items = [di.item for di in source.items]
            mapped = mapper.map(raw_items)
            return Page(
                worker_id=source.worker_id,
                items=[
                    DataItem(tracking_id=item.tracking_id, item=target)
                    for target, item in zip(mapped, source.items)
                    if target is not None
                ],
                bookmark=source.bookmark,
            )

        return track_mapping
``` 

Since zip() is not used with strict mode, it seems this would allow lists of uneven lengths and then terminate with whatever is shortest. It seems at least that this if causes the one list to be potentially shorter (and ever mismatching ChartResponses with their ChartRequests

Magssch (approved):

PR #2858, opened 2026-04-15
[CDF-27673] 🐛 Infield users

doctrino:
# Description

See https://github.com/cognitedata/toolkit/pull/2858/changes#diff-dea81a2f4648a92b7eb9adfc619cf0fd1e7bf913e7c743480a26b21db4c0886c

## Bump

- [x] Patch
- [ ] Skip

### Fixed

- [alpha] When running `cdf migrate infiel-data`, Toolkit now maps to `sourceCreatedUser` and `sourceUpdatedUser` for all views that has these properties.

### DIFF
--- a/cognite_toolkit/_cdf_tk/commands/_migrate/conversion.py
+++ b/cognite_toolkit/_cdf_tk/commands/_migrate/conversion.py
@@ -974,24 +974,21 @@ class InFieldUserMapping(CustomContainerPropertiesMapping):
     ) -> ConversionResult:
         created_properties: dict[str, JsonValue | NodeId | list[NodeId]] = {}
         issues: list[str] = []
-        if created_user := source_properties.get("createdBy"):
-            if isinstance(created_user, dict) and "externalId" in created_user:
-                created_properties["sourceCreatedUser"] = created_user["externalId"]
-            elif isinstance(created_user, NodeId):
-                created_properties["sourceCreatedUser"] = created_user.external_id
-            else:
-                issues.append(
-                    f"Invalid createdBy value {created_user!r} for view {context.source_view_id!s}: expected a dict with an externalId field or a NodeId."
-                )
-        if updated_user := source_properties.get("updatedBy"):
-            if isinstance(updated_user, dict) and "externalId" in updated_user:
-                created_properties["sourceUpdatedUser"] = updated_user["externalId"]
-            elif isinstance(updated_user, NodeId):
-                created_properties["sourceUpdatedUser"] = updated_user.external_id
-            else:
-                issues.append(
-                    f"Invalid updatedBy value {updated_user!r} for view {context.source_view_id!s}: expected a dict with an externalId field or a NodeId."
-                )
+
+        def _map_user(source_prop: str, target_prop: str) -> None:
+            if user := source_properties.get(source_prop):
+                if isinstance(user, dict) and "externalId" in user:
+                    created_properties[target_prop] = user["externalId"]
+                elif isinstance(user, NodeId):
+                    created_properties[target_prop] = user.external_id
+                else:
+                    issues.append(
+                        f"Invalid {source_prop} value {user!r} for view {context.source_view_id!s}: "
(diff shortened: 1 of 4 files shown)


cognite_toolkit/_cdf_tk/commands/_migrate/data_mapper.py
```diff
@@ -1091,9 +1092,10 @@ def __init__(
         self._mappings_by_source_view: dict[ViewId, ViewToViewMapping] = {
             mapping.source_view: mapping for mapping in mappings
         }
-        self._custom_properties_mapping: dict[ViewId, CustomContainerPropertiesMapping] = {
-            view_id: mapping for mapping in (custom_properties_mappings or []) for view_id in mapping.VIEW_IDS
-        }
+        self._custom_properties_mapping: dict[ViewId, list[CustomContainerPropertiesMapping]] = defaultdict(list)
+        for mapping in custom_properties_mappings or []:
+            for view_id in mapping.VIEW_IDS:
+                self._custom_properties_mapping[view_id].append(mapping)
```
doctrino, line 1098: Added to handle multiple CustmoContainerPropertiesMapping for the same view

tests_smoke/test_commands/test_migrate_infield/test_migrate_data.yml
doctrino, line 1: The proof that the extra mapping works.

cognite_toolkit/_cdf_tk/commands/_migrate/conversion.py
```diff
@@ -941,6 +941,57 @@ def convert(
...
+            ViewId(space="cdf_apm", external_id="ConditionalAction", version="v1"),
+            ViewId(space="cdf_apm", external_id="MeasurementReading", version="v4"),
+            ViewId(space="cdf_apm", external_id="Observation", version="v5"),
+            ViewId(space="cdf_apm", external_id="Schedule", version="v4"),
+            ViewId(space="cdf_apm", external_id="Template", version="v9"),
+            ViewId(space="cdf_apm", external_id="TemplateItem", version="v7"),
+        }
+    )
+
+    def convert(
```
Magssch, line 972: Could we make this a class / static method? Such that we don't need to instantiate the InFieldUserMapping for every conversion.

cognite_toolkit/_cdf_tk/commands/_migrate/conversion.py
```diff
@@ -941,6 +941,57 @@ def convert(
...
+    VIEW_IDS: ClassVar[Set[ViewId]] = frozenset(
+        {
+            ViewId(space="cdf_apm", external_id="Action", version="v1"),
+            ViewId(space="cdf_apm", external_id="Checklist", version="v7"),
+            ViewId(space="cdf_apm", external_id="ChecklistItem", version="v7"),
+            ViewId(space="cdf_apm", external_id="Condition", version="v1"),
+            ViewId(space="cdf_apm", external_id="ConditionalAction", version="v1"),
+            ViewId(space="cdf_apm", external_id="MeasurementReading", version="v4"),
+            ViewId(space="cdf_apm", external_id="Observation", version="v5"),
+            ViewId(space="cdf_apm", external_id="Schedule", version="v4"),
```
Magssch, line 966: Very optional, but could it be nice here with a comment that Schedule is not handled by this class and instead uses a custom mapping?

doctrino: It is handled by this class :)

cognite_toolkit/_cdf_tk/commands/_migrate/conversion.py
doctrino: No, because the method is not static in the base class. And the other custom conversion depends on parameters being sent in.

Magssch: Missed the other class, looks good then

Magssch (approved):

PR #2867, opened 2026-04-15
🐛 Fix: Show user-friendly errors for invalid function parameters

nikokaoja:
# Description

Function deployment now catches HTTP 400 validation errors and converts them to user-friendly messages specifying which parameter is invalid and its valid range (e.g., "memory must lie in the range [0.1, 1.5]").

## Bump

- [ ] Patch
- [x] Skip

### Fixed

- Function deployment error handling for invalid parameter values (memory, cpu, etc.) now shows clear, actionable error messages instead of crashing with unhandled `ToolkitAPIError`

### DIFF
--- a/cognite_toolkit/_cdf_tk/resource_ios/_resource_ios/function.py
+++ b/cognite_toolkit/_cdf_tk/resource_ios/_resource_ios/function.py
@@ -487,10 +487,11 @@ class FunctionIO(ResourceIO[ExternalId, FunctionRequest, FunctionResponse]):
             try:
                 result = self.client.tool.functions.create([item_to_create])
             except ToolkitAPIError as e:
-                if e.code == 400:
-                    # Handle validation errors with user-friendly message
-                    raise ResourceCreationError(self._parse_validation_error(e, external_id)) from e
-                raise
+                # Handle any HTTP error with user-friendly message
+                raise ResourceCreationError(self._parse_validation_error(e, external_id)) from e
+            except Exception as e:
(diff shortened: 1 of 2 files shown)


cognite_toolkit/_cdf_tk/resource_ios/_resource_ios/function.py
```diff
@@ -428,6 +429,32 @@ def _is_activated(self, action: str) -> bool:
             self.client.functions.activate()
         return False
 
+    @staticmethod
+    def _parse_validation_error(api_error: ToolkitAPIError, external_id: str) -> str:
```
nikokaoja, line 433: opted for this option since alternative would be to detect cloud provider and raise app. validation error:

https://docs.cognite.com/cdf/functions/technical

doctrino (commented):
Two issues here
- You are capturing the issue in deploy and not in build. So it is late. 
- You are essentially overriding the API error with a new generic message, not sure if this is worth it. 

Suggest to introduce this as a rule.

nikokaoja:
> Two issues here
> 
> * You are capturing the issue in deploy and not in build. So it is late.
> * You are essentially overriding the API error with a new generic message, not sure if this is worth it.
> 
> Suggest to introduce this as a rule.

Opted for overriding the API error as suggested by JIRA ticket:

https://cognitedata.atlassian.net/browse/CDF-27636?actionerId=5e8208727dd96b0c156a87bd&sourceType=assign

PR #2871, opened 2026-04-15
[CDF-27690] Add data model as deployment dependency for agents

ronpal:
# Description

Agents were being deployed before data model resources (spaces, containers, views, data models), which could leave non-functional agents deployed when data model deployment failed. This adds `DataModelIO` to `AgentIO.dependencies` so the topological sorter always places data model resources before agents.

Also tracks per-agent data model dependencies from `queryKnowledgeGraph` tools in `get_dependent_items` and `get_dependencies`.

## Bump

- [x] Patch
- [ ] Skip

### Fixed

- Agents are now deployed after data model resources, preventing non-functional agents from being deployed when data model deployment fails


doctrino (approved):
Great stuff

PR #2877, opened 2026-04-16
[CDF-27698] 📈Charts with monitoring jobs

doctrino:
# Description

Manually tested. Note it is hard to have an automated test that one user creates a chart that another user have craeted.

image

## Bump

- [x] Patch
- [ ] Skip

### Added

- [alpah] Migrating charts with monitoring jobs is supported.


aneeide (approved):
lgtm

PR #2897, opened 2026-04-17
[prep CDF-26363, CDF-26928] 🦈Refactor DownloadCommand.

doctrino:
# Description

Refactor DownloadCommand in preparation for downloading asset-centric table format.

* Split the `.download` into smaller method for easy overview. Focused on splitting functionality, such as creating a plan, displaying the plan as well as simple execution of plan. 
* Removed the stop/resume logic that is not yet in use. 

The entire download now fits well within a screenshot
image

## Bump

- [ ] Patch
- [x] Skip


ronpal (approved):
Very clean 👏

PR #2899, opened 2026-04-19
[CDF-27734] 🫛Replace annoying warning

doctrino:
# Description

## Annoying warning

image

## Replaced by only if an environment variable is actually overridden

image

Furthermore, cleaned up the logic in the function that deals with loading the `.env` path.

## Bump

- [x] Patch
- [ ] Skip

### Improved

- When running any Toolkit command, the warning `Overriding environment variables with values from .env file...` is only printed if you actually override and environmental variable with your `.env` file. Furthermore, the warning now says exactly which environmental variable(s) are overridden.


cognite_toolkit/_cdf_tk/apps/_core_app.py
```diff
@@ -71,7 +74,7 @@ def common(
             ),
         ] = True,
         env_path: Annotated[
-            str | None,
+            Path | None,
```
doctrino, line 77: Not a breaking change, as if the user did not pass a path this would have been raised as an exception in the previous code as well as the current one.

ronpal (approved):

PR #2908, opened 2026-04-21
[CDF-27700] Replace container update warning with structured diff and remediation hints

Magssch:
# Description

Replace the potentially harmful "You might need to delete and recreate it" container warning with a structured diff that lists specific properties, constraints, and indexes that differ between a local container YAML and CDF. Adds remediation hints pointing users to `cdf modules pull` and the container changes documentation.

## Bump

- [x] Patch
- [ ] Skip

### Improved

- When running `cdf deploy` with a container that has properties, constraints, or indexes that have been removed locally in your YAML file, toolkit will now list the diff against the deployed container clearly with actionable remediation steps instead of suggesting full container recreation.

### DIFF
--- a/cognite_toolkit/_cdf_tk/resource_ios/_resource_ios/datamodel.py
+++ b/cognite_toolkit/_cdf_tk/resource_ios/_resource_ios/datamodel.py
@@ -438,7 +438,7 @@ class ContainerCRUD(ResourceContainerIO[ContainerId, ContainerRequest, Container
 
         lines = [
             f"Container {item_id} has differing config in your local YAML as opposed to CDF.",
-            "This may occur if you have tried to remove previously deployed properties, constraints or indexes from a container in your local container YAML file.",
+            "This may occur if you have tried to remove previously deployed properties, constraints or indexes from a container in your local YAML file.",
             "Please note that CDF containers do not support removing existing property definitions and CDF toolkit currently does not support removing constraints or indexes.",
             "This warning will persist until the discrepancies are resolved.",
         ]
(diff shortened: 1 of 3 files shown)


cognite_toolkit/_cdf_tk/resource_ios/_resource_ios/datamodel.py
```diff
@@ -409,33 +411,65 @@ def retrieve(self, ids: Sequence[ContainerId]) -> list[ContainerResponse]:
...
-        # The API might silently fail to update a container.
+        # The API might silently fail to apply some changes.
         updated_by_id = {item.as_id(): item for item in updated}
         for local in items:
             item_id = local.as_id()
             local_dict = local.dump()
             if item_id not in updated_by_id:
                 raise ToolkitAPIError(
-                    f"The container {item_id} was not updated. You might need to delete and recreate it.",
+                    f"The container {item_id} was not updated due to an unknown error.",
```
Magssch, line 421: I'm not sure if this is a real API edge case anymore or not, so I've kept it in, but removed the potentially harmful advice.

cognite_toolkit/_cdf_tk/resource_ios/_resource_ios/datamodel.py
```diff
@@ -409,33 +411,65 @@ def retrieve(self, ids: Sequence[ContainerId]) -> list[ContainerResponse]:
...
+        lines = [
+            f"Container {item_id} has differing config in your local YAML as opposed to CDF.",
+            "This may occur if you have tried to remove previously deployed properties, constraints or indexes from a container in your local YAML file.",
+            "Please note that CDF containers do not support removing existing property definitions and CDF toolkit currently does not support removing constraints or indexes.",
+            "This warning will persist until the discrepancies are resolved.",
+        ]
+        if only_in_cdf_props:
+            lines.append(f"• Properties missing in your local YAML: [bold]{', '.join(only_in_cdf_props)}[/bold]")
+        if only_in_cdf_constraints:
+            lines.append(f"• Constraints missing in your local YAML: [bold]{', '.join(only_in_cdf_constraints)}[/bold]")
```
Magssch, line 448: I want us to handle constraints and indexes gracefully in the future, but for the time being, these would trigger the same warning so they need to be handled the same way we do for properties for now

doctrino (approved):

PR #2910, opened 2026-04-21
[CDF-27752] 🦐Selection of config.yaml from defaults

doctrino:
# Description

## No default set
User have to select
image

## Default config yaml is set <code>default_config_yaml = "my_sim/config.dev.yaml"</code> in <code>cdf.toml</code>
Runs without issues
image

## Default env set <code>env=dev</code>, but not default config yaml set in <code>cdf.toml</code>
Runs with a deprecation warning

image

## Bump

- [x] Patch
- [ ] Skip

### Fixed

- When running `cdf build`, Toolkit now correctly selects the `config.[name].yaml` file based on the defaults in the `cdf.toml`.

### DIFF
--- a/cognite_toolkit/_cdf_tk/apps/_core_app.py
+++ b/cognite_toolkit/_cdf_tk/apps/_core_app.py
@@ -341,8 +341,9 @@ class CoreApp(typer.Typer):
 
         if build_env_name is not None and config_yaml is None:
             ToolkitDeprecationWarning(
-                feature="the --env / -e option in cdf build",
-                alternative="--config-yaml / -c with the path to your config file (for example <organization-dir>/config.<env>.yaml)",
+                feature="the --env / -e option in cdf build or default_env in cdf.toml",
+                alternative="--config-yaml / -c or config_yaml in cdf.toml with the path to your config file "
+                "(for example <organization-dir>/config.<env>.yaml)",
             ).print_warning()
             if config_yaml is None:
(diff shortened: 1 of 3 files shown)


cognite_toolkit/_cdf_tk/cdf_toml.py
```diff
@@ -46,7 +46,8 @@ def load(cls, raw: dict[str, Any], cwd: Path) -> Self:
 
         return cls(
             default_organization_dir=default_organization_dir,
-            default_env=raw.get("default_env", "dev"),
+            default_env=raw.get("default_env"),
```
doctrino, line 49: we should no longer assume `dev` default here.

cognite_toolkit/_cdf_tk/apps/_core_app.py
```diff
@@ -312,7 +312,7 @@ def build_v2(
                 "-e",
                 help="Deprecated. Prefer --config-yaml. If set and --config-yaml is omitted, uses <organization-dir>/config.<env>.yaml.",
             ),
-        ] = None,
+        ] = CDF_TOML.cdf.default_env,
```
Magssch, line 315: Now you get a slightly confusing deprecation error if you run `cdf build` without specifying `--env`, it will still say that this is deprecated. Can we modify to say something like "the --env / -e option in cdf build and "default_env" field in cdf.toml is deprecated" so it's explicitly stated? I don't think it's necessarily clear to users that the toml field auto-sets this parameter

Image

Magssch (approved):

PR #2912, opened 2026-04-21
[CDF-27699] Block space purge when containers are referenced by external views

Magssch:
# Description

Adds a pre-purge safety precaution that calls `models/containers/inspect` before deleting containers in a space. Uses `involvedViewCount` (the authoritative total, including inaccessible views) to decide whether to block the purge, and surfaces the visible referencing views (using `involvedView`) in a structured table before raising an error.

## Bump

- [x] Patch
- [ ] Skip

### Added

- Added a safety precaution to `cdf data purge space` that blocks deletion of a space if any container is referenced by views in other spaces, showing a table of affected containers and views.


doctrino (commented):
Good stuff :)

doctrino (approved):

PR #2914, opened 2026-04-21
[🪼]  Fix no access

doctrino:
# Description

Found when setting up a new CDF porject

## Bump

- [x] Patch
- [ ] Skip

### Fixed

- When calling `cdf auth verify`, Toolkit no longer fails with `You are not allowed to access this resource endpoint | code: 403` if you lack access.

### DIFF
--- a/cognite_toolkit/_cdf_tk/commands/auth.py
+++ b/cognite_toolkit/_cdf_tk/commands/auth.py
@@ -430,7 +430,12 @@ class AuthCommand(ToolkitCommand):
                 continue
 
             crud = crud_cls.create_loader(client)
-            if crud.prerequisite_warning() is not None:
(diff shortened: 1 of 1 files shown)


cognite_toolkit/_cdf_tk/commands/auth.py
```diff
@@ -430,7 +430,12 @@ def _get_required_acls(
                 continue
 
             crud = crud_cls.create_loader(client)
-            if crud.prerequisite_warning() is not None:
+            try:
+                warning_str = crud.prerequisite_warning()
```
doctrino, line 434: Checking prerequisite (such migration model) requires access we may not yet have.

ronpal (approved):

PR #2917, opened 2026-04-21
[CDF-27760] 👌Function requirements.txt

doctrino:
# Description

When moving from v0.7 to v0.8 these checks was lost (they were placed inside the old builder classes which we do not use any more)

Adding them back in

image

## Bump

- [x] Patch
- [ ] Skip

### Fixed

- When running `cdf build` with Cognite Functions, Toolkit now validates the `requirement.txt`.

### DIFF
--- a/cognite_toolkit/_cdf_tk/commands/build_v2/data_classes/_build.py
+++ b/cognite_toolkit/_cdf_tk/commands/build_v2/data_classes/_build.py
@@ -79,6 +79,7 @@ class BuiltResource(BaseModel):
     crud_cls: builtins.type[ResourceIO]
     dependencies: set[tuple[builtins.type[ResourceIO], Identifier]] = Field(default_factory=set)
     extra_files: list[ReadExtra] = Field(default_factory=list)
+    has_syntax_errors: bool
 
 
 class BuiltModule(BaseModel):
--- a/cognite_toolkit/_cdf_tk/commands/build_v2/data_classes/_build.py
+++ b/cognite_toolkit/_cdf_tk/commands/build_v2/data_classes/_build.py
@@ -7,7 +7,7 @@ from pydantic import BaseModel, ConfigDict, Field, JsonValue
 
 from cognite_toolkit._cdf_tk.client._resource_base import Identifier
 from cognite_toolkit._cdf_tk.constants import MODULES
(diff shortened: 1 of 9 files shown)


cognite_toolkit/_cdf_tk/commands/build_v2/data_classes/_build.py
```diff
@@ -78,6 +78,8 @@ class BuiltResource(BaseModel):
     build_path: AbsoluteFilePath
     crud_cls: builtins.type[ResourceIO]
     dependencies: set[tuple[builtins.type[ResourceIO], Identifier]] = Field(default_factory=set)
+    extra_files: list[ReadExtra] = Field(default_factory=list)
+    has_syntax_errors: bool
```
doctrino, line 82: Information that we should persist

cognite_toolkit/_cdf_tk/rules/_functions.py
```diff
@@ -54,43 +58,51 @@ def _validate_function(self, function_file: Path) -> Iterable[ConsistencyError]:
             ConsistencyError for any violations of function limits.
         """
         # Parse function_file (YAML) to dict/list, then create FunctionsYAML objects to validate and extract definitions
-        raw_data = read_yaml_file(function_file, expected_output="dict")
+        raw_data = read_yaml_file(resource.build_path, expected_output="dict")
 
         # Ensure we always work with a list of function definitions
-        function_defs = [raw_data] if isinstance(raw_data, dict) else (raw_data or [])
```
doctrino, line 60: We know that the build folder always have one file pre resource.

cognite_toolkit/_cdf_tk/rules/_base.py
```diff
@@ -40,7 +40,7 @@ def _get_validated_resources_with_file(self) -> Iterable[tuple[ToolkitResource,
 
 
 class RuleSetStatus(BaseModel):
-    code: Literal["ready", "skip", "unavailable"]
+    code: Literal["ready", "reduced", "skip", "unavailable"]
```
doctrino, line 43: `reduced` is meant to say that you can do some checks but not others. An alternative is `limited`. Any other suggestion?

ronpal (approved):

PR #2924, opened 2026-04-22
[CDF-27767] Add RUN action to FunctionsAcl

ronpal:
# Description

CDF's `functionsAcl` now supports a `RUN` action in addition to `READ` and `WRITE`.
This PR adds `RUN` to the allowed actions in both the client ACL model and the YAML
capability class.

Also adds `py313` to the allowed function runtimes — CDF recently bumped its default
runtime to Python 3.13.

## Bump

- [x] Patch
- [ ] Skip

### Added

- `FunctionsAcl` now accepts `RUN` as a valid action in addition to `READ` and `WRITE`.

### Fixed

- `py313` added to the list of valid function runtimes to match CDF's updated default.

### DIFF
--- a/cognite_toolkit/_cdf_tk/client/resource_classes/function.py
+++ b/cognite_toolkit/_cdf_tk/client/resource_classes/function.py
@@ -23,7 +23,7 @@ class FunctionBase(BaseModelObject):
     env_vars: dict[str, str] | None = None
     cpu: float | None = None
     memory: float | None = None
-    runtime: FunctionRuntime | None = None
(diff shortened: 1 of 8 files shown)


cognite_toolkit/_cdf_tk/client/resource_classes/function.py
doctrino: I think this should be forward compatible and not break

```suggestion
    runtime: FunctionRuntime | str | None = None
```

doctrino (commented):
One suggestion

doctrino (approved):

PR #2926, opened 2026-04-22
[CDF-27773] Fix custom library being ignored when organization_dir is a subdirectory

Magssch:
# Description

The existence check for `cdf.toml` in `modules init` was scoped to `organization_dir` instead of CWD, so passing a subdirectory (e.g. `cdf modules init my_organization`) caused the user's custom library configuration in their repo root-level `cdf.toml` to be silently replaced by the bundled default. This occurs whenever `organization_dir` is a subdirectory of the directory containing `cdf.toml`.

The default library was further printed misleadingly as `Adding library userdefined` even though it wasn't custom. The fallback logic is moved into `_get_available_packages` so the bundled default keeps its real name (`cognite`) and `init` no longer needs to know about resource paths.

This PR also adds a `LowSeverityWarning` when `cdf.toml` exists but defines no libraries (previously this would crash), falling back to the bundled default in that case.

## Bump

- [x] Patch
- [ ] Skip

### Fixed

- Fix `cdf modules init <organization_dir>` ignoring the custom library configured in `cdf.toml`, and showing the default library as `userdefined` in the output.


cognite_toolkit/_cdf_tk/commands/modules.py
```diff
@@ -294,17 +294,7 @@ def init(
 
         modules_root_dir = organization_dir / MODULES
 
-        # Determine which library to use (if any)
-        library: Library | None = None
-        if library_url:
-            library = Library(url=library_url)
-        elif not (organization_dir / CDFToml.file_name).exists():
```
Magssch, line 301: organization_dir here seemed to be wrong, I don't think we intend to support `cdf.toml` within a nested folder?

cognite_toolkit/_cdf_tk/commands/modules.py
```diff
@@ -865,7 +855,15 @@ def _get_available_packages(self, user_library: Library | None = None) -> tuple[
         cdf_toml = CDFToml.load()
 
         if self._module_source_dir is None:
-            libraries = {"userdefined": user_library} if user_library else cdf_toml.libraries
+            if user_library:
+                libraries = {"userdefined": user_library}
```
Magssch, line 858: Previously "userdefined" would also be used for the default library if `cdf.toml` didn't exist

cognite_toolkit/_cdf_tk/commands/modules.py
doctrino: correct

cognite_toolkit/_cdf_tk/commands/modules.py
```diff
@@ -865,7 +855,16 @@ def _get_available_packages(self, user_library: Library | None = None) -> tuple[
...
+            elif cdf_toml.libraries:
+                libraries = cdf_toml.libraries
+            else:
+                self.warn(LowSeverityWarning("No libraries defined in cdf.toml. Using default library."))
+                default_cdf_toml = CDFToml.load(cwd=RESOURCES_PATH, use_singleton=False)
```
doctrino: Optional: Make a new classmethod to CDFToml.load_default() that wraps this calls. It makes it easier to read here.

cognite_toolkit/_cdf_tk/commands/modules.py
```diff
@@ -865,7 +855,16 @@ def _get_available_packages(self, user_library: Library | None = None) -> tuple[
...
+                libraries = default_cdf_toml.libraries
+                if not libraries:
+                    raise ToolkitError("Default cdf.toml in resources is missing Library configuration.")
```
doctrino: This is a bug in Toolkit
```suggestion
                    raise ToolkitError("Toolkit Bug: Default cdf.toml in resources is missing Library configuration.")
```

doctrino (commented):
One optional and one change in wording.

doctrino (approved):

PR #2935, opened 2026-04-23
[CDF-27670] 	🐛InField config migration

doctrino:
# Description

## Before
image

## After
image

## Bump

- [x] Patch
- [ ] Skip

### Fixed

- When running `cdf migrate infield-configs`, Toolkit now produces one template configuration for each location.

### DIFF
--- a/cognite_toolkit/_cdf_tk/commands/_migrate/creators.py
+++ b/cognite_toolkit/_cdf_tk/commands/_migrate/creators.py
@@ -440,10 +440,13 @@ class InfieldV2ConfigCreator(MigrationCreator):
         ]:
             view_mappings[key] = default_view.dump()
 
+        original_external_id = config.external_id or config.asset_external_id or str(uuid.uuid4())
+        external_id = f"infield_config_{original_external_id}"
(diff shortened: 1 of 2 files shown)


cognite_toolkit/_cdf_tk/commands/_migrate/creators.py
```diff
@@ -440,10 +440,13 @@ def _create_location_config(
...
             view_mappings[key] = default_view.dump()
 
+        original_external_id = config.external_id or config.asset_external_id or str(uuid.uuid4())
+        external_id = f"infield_config_{original_external_id}"
+        name = config.display_name or config.asset_external_id or external_id
         return InFieldCDMLocationConfigRequest(
             space=config.source_data_instance_space or "<Please fill in the source data instance space>",
             external_id=external_id,
-            name="InField Location Config",
+            name=name,
```
doctrino, line 447: Name is used as part of the file name for the resulting template, thus must be different for each location to avoid overwriting each other.

Magssch (approved):

PR #2937, opened 2026-04-23
Fix upserts for records in immutable streams

Magssch:
# Description

Records uploads (via both `cdf upload records` and the `events-to-records` migration) always POSTed to the create-only endpoint, which caused 409 errors when writing to existing records in mutable streams. This PR ensures we first look up the target stream and route all uploads to `/streams/{streamId}/records/upsert` for mutable streams, and `/streams/{streamId}/records` for immutable ones (which doesn't support the upsert endpoint)

## Bump

- [x] Patch
- [ ] Skip

### Fixed

- Uploading records to a mutable stream no longer fails with 409 conflicts when records already exist


cognite_toolkit/_cdf_tk/apps/_migrate_app.py
```diff
@@ -701,13 +715,18 @@ def events_to_records(
         elif data_set_id is not None:
             selected = MigrateDataSetSelector(data_set_external_id=data_set_id, kind="Events")
 
+        if is_interactive:
```
Magssch, line 718: This was missing compared to other migration commands

cognite_toolkit/_cdf_tk/dataio/_records.py
```diff
@@ -26,23 +28,37 @@ class RecordIO(
...
     # TODO: Replace with adaptive limit that targets ~3MB uncompressed response size
     CHUNK_SIZE = 500
     MAX_TOTAL_RECORDS = 1_000_000
     BASE_SELECTOR = RecordContainerSelector
 
+    def __init__(self, client: ToolkitClient) -> None:
+        super().__init__(client)
+        self._stream_by_external_id: dict[str, StreamResponse] = {}
+
+    def _get_stream(self, stream_external_id: str) -> StreamResponse:
```
Magssch, line 42: Not entirely satisfied with this shoehorned solution, but it's an effect of having to switch upload endpoints based on the stream and avoiding doing two calls to fetch the stream... 

Alternately we could fetch the stream in the constructor of RecordIO, but then you would also have selectors in the count() and upload_items() methods that define a Stream. Which is a bit awkward as well, since that technically then _could_ be a different stream than what was set and fetched in the constructor (consistency only guaranteed by external users of the class).

ronpal (approved):

PR #2942, opened 2026-04-24
[CDF-27790] 🪟 InField data migration creatable views

doctrino:
# Description

SHOULD NOT BE MERGED. This is a bug in the InField app and should not be fixed by the migration tooling.

## Bump

- [x] Patch
- [ ] Skip

### Added

- [alpha] When running `cdf migrate infield-data`, Toolkit now sets `visibility=PUBLIC` and `isArchived=false` for all creatable views that has these properties set to `None`.

### DIFF
--- a/cognite_toolkit/_cdf_tk/commands/_migrate/conversion.py
+++ b/cognite_toolkit/_cdf_tk/commands/_migrate/conversion.py
@@ -1021,11 +1021,11 @@ class InFieldCreatableViews(CustomContainerPropertiesMapping):
         issues: list[str] = []
 
         is_archived = source_properties.get("isArchived", None)
-        visibility = source_properties.get("visibility", None)
         if is_archived is None:
             created_properties["isArchived"] = True
+        visibility = source_properties.get("visibility", None)
         if visibility is None:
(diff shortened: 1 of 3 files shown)


cognite_toolkit/_cdf_tk/commands/_migrate/conversion.py
```diff
@@ -992,6 +992,44 @@ def _map_user(source_prop: str, target_prop: str) -> None:
...
+
+    def convert(
+        self, source_properties: dict[str, JsonValue | NodeId | list[NodeId]], context: ConversionContext
+    ) -> ConversionResult:
+        created_properties: dict[str, JsonValue | NodeId | list[NodeId]] = {}
+        issues: list[str] = []
+
+        is_archived = source_properties.get("isArchived", None)
+        if is_archived is None:
+            created_properties["isArchived"] = True
```
atussak:
```suggestion
            created_properties["isArchived"] = False
```
?

doctrino:
```suggestion
            created_properties["isArchived"] = False
```

PR #2952, opened 2026-04-24
Surface validator crashes as visible build insights instead of silently dropping them

Magssch:
# Description

`FailedValidation` (yielded from `except Exception` blocks in `FunctionRules` and `NeatRuleSet`) was collected into `ValidationResult.failed` but never included in `BuildFolder.all_insights`, so any crash during validation was silently swallowed and the user saw a green build even though the validation didn't run. This PR promotes `FailedValidation` to a proper `InsightDefinition` subclass, adds it to the `Insight` type alias, and surfaces it directly in the build summary with a dedicated "Validation incomplete" warning path to indicate potential issues.

## Bump

- [x] Patch
- [ ] Skip

### Fixed

- Fix validator crashes being silently ignored during `cdf build` — they now appear as visible errors with a "Validation incomplete" warning in the build summary.

### DIFF
--- a/cognite_toolkit/_cdf_tk/commands/build_v2/build_v2.py
+++ b/cognite_toolkit/_cdf_tk/commands/build_v2/build_v2.py
@@ -38,7 +38,11 @@ from cognite_toolkit._cdf_tk.commands.build_v2.data_classes import (
     ValidationType,
 )
 from cognite_toolkit._cdf_tk.commands.build_v2.data_classes._build import BuiltResource, ValidationResult
-from cognite_toolkit._cdf_tk.commands.build_v2.data_classes._insights import Insight, ModelSyntaxWarning
+from cognite_toolkit._cdf_tk.commands.build_v2.data_classes._insights import (
+    FailedValidation,
+    Insight,
+    ModelSyntaxWarning,
+)
 from cognite_toolkit._cdf_tk.commands.build_v2.data_classes._module import (
     SUPPORTS_VARIABLE_REPLACEMENT,
     BuildSource,
@@ -63,7 +67,7 @@ from cognite_toolkit._cdf_tk.resource_ios import (
 )
 from cognite_toolkit._cdf_tk.resource_ios._base_ios import FailedReadExtra, ReadExtra, SuccessExtra
 from cognite_toolkit._cdf_tk.rules import LocalRulesOrchestrator, ToolkitGlobalRuleSet, get_global_rules_registry
-from cognite_toolkit._cdf_tk.rules._base import FailedValidation, RuleSetStatus
+from cognite_toolkit._cdf_tk.rules._base import RuleSetStatus
 from cognite_toolkit._cdf_tk.utils import calculate_hash, humanize_collection, safe_write
 from cognite_toolkit._cdf_tk.utils.file import (
     read_yaml_content,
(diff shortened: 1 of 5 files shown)


cognite_toolkit/_cdf_tk/commands/build_v2/data_classes/_insights.py
```diff
@@ -38,6 +38,18 @@ class ConsistencyError(InsightDefinition):
...
 
 
+class FailedValidation(InsightDefinition):
+    """A validator threw an unexpected exception and could not complete.
+
+    This should never happen in normal operation — it indicates a bug in the validator itself.
+    Severity is 0 so it does not inflate max_severity in the build summary; it is checked separately.
+    """
+
+    severity = 0
```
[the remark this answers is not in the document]
Magssch: I chose to set it to zero and handle it as a special case to avoid having to edit the warning that occurs when you have YAML / syntax errors. Additionally, I am of the opinion that we shouldn't say "Stop" necessarily if this happens, since it might still be fine, we just can't guarantee it.

cognite_toolkit/_cdf_tk/commands/build_v2/data_classes/_build.py
```diff
@@ -190,6 +185,7 @@ def all_insights(self) -> InsightList:
             insights.extend(module.all_insights)
         for result in self.validation_results:
             insights.extend(result.insights)
+            insights.extend(result.failed)
```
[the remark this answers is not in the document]
Magssch: Doing a cleanup on this

cognite_toolkit/_cdf_tk/commands/build_v2/data_classes/_insights.py
doctrino: Shouldn't this have a very high severity?

cognite_toolkit/_cdf_tk/commands/build_v2/build_v2.py
```diff
@@ -865,21 +864,29 @@ def _display_build_summary(
...
-                case _:
-                    insight_style = "[red]✗[/]"
+            if insight_type == "FailedValidation":
+                insight_style = "[red]✗[/]"
+            elif severity <= 15:
+                insight_style = "[green]✓[/]"
+            elif severity <= 35:
+                insight_style = "[yellow]![/]"
+            else:
+                insight_style = "[red]✗[/]"
```
doctrino: Not sure I see the value of this. Why not just set the severity? This complicates matters.

Magssch: Fair, I was thinking we do a special case in order to avoid saying "Do not deploy" if this happens, e.g. there is an error in the functions limit API, but then again it isn't really "wrong" to say that either. It might have been more relevant if we had more external validators / dependencies that potentially could sporadically fail. I'll change it back to a higher severity instead

cognite_toolkit/_cdf_tk/commands/build_v2/build_v2.py
```diff
@@ -892,7 +887,7 @@ def _display_build_summary(
             case _:
                 recommendation = (
                     "[red]✗[/] [bold]Do not proceed to deploy.[/bold]\n"
-                    "There are YAML parsing errors that must be fixed before deployment."
+                    "There are critical errors that must be fixed before deployment."
```
Magssch, line 890: This was tweaked now since this path now triggers due to the higher severity of FailedValidation

doctrino: Very good

doctrino (approved):

PR #2956, opened 2026-04-26
😅Fix failing smoke test on greenfield

doctrino:
# Description

## Bump

- [ ] Patch
- [x] Skip

### DIFF
--- a/tests_smoke/test_commands/test_migrate_infield.py
+++ b/tests_smoke/test_commands/test_migrate_infield.py
@@ -200,14 +200,14 @@ def infield_legacy(
     for view_id in sorted_views:
         instance_batch = to_create_by_view_id[view_id]
         try:
-            _ = client.tool.instances.create(instance_batch)
+            _ = client.tool.instances.create(instance_batch, replace=True)
         except ToolkitAPIError as e:
             raise AssertionError(
                 f"Failed to create instance batch for view {view_id!s}. Error: {e}. Batch: {[item.as_id() for item in instance_batch]}"
             ) from e
     if edges:
         try:
-            _ = client.tool.instances.create(edges)
+            _ = client.tool.instances.create(edges, replace=True)
         except ToolkitAPIError as e:
             raise AssertionError(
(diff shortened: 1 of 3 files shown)


tests_smoke/test_commands/test_migrate_infield/test_migrate_data.yml
```diff
@@ -1,6 +1,5 @@
 instances:
-- createdTime: 1772711569310
```
doctrino: createdTime now change for each run.

tests_smoke/test_commands/test_migrate_infield.py
```diff
@@ -352,6 +355,10 @@ def test_migrate_data(
                 "InField migration failed. Found source space identifier in destination data, indicating that some instances were not migrated correctly."
             )
 
+        # Cleanup for next run.
+        destination_node_ids = [node_id for node_ids in destination_by_view_id.values() for node_id in node_ids]
+        toolkit_client.tool.instances.delete(destination_node_ids)
```
doctrino, line 360: The smoke tests currently failed because, [this PR](https://github.com/cognitedata/toolkit/pull/2942) changed the destination data. Since the destination data is not cleaned up, only added to, then, the test started failing.

tests_smoke/test_commands/test_migrate_infield.py
```diff
@@ -200,14 +200,14 @@ def infield_legacy(
...
-            _ = client.tool.instances.create(instance_batch)
+            _ = client.tool.instances.create(instance_batch, replace=True)
         except ToolkitAPIError as e:
             raise AssertionError(
                 f"Failed to create instance batch for view {view_id!s}. Error: {e}. Batch: {[item.as_id() for item in instance_batch]}"
             ) from e
     if edges:
         try:
-            _ = client.tool.instances.create(edges)
+            _ = client.tool.instances.create(edges, replace=True)
```
doctrino, line 210: This is strictly not necessary, but I added it as it make the set up more robust. Legacy InField is set up exactly the same way each time, it it pre-exist it will be overwritten.

ronpal (approved):

PR #2957, opened 2026-04-26
feat(infield): add validation for DataExplorationConfig and ViewMappings in InFieldCDMLocationConfig

laylafy:
# Description

DataExplorationConfig will now be used for configurations for asset cards in infield. It is currently not being used for anything else, so I removed all deprecated fields, and added validation for how it should look like.

Another is to change the view mappings validation for observations list. For now, it should only allow one item (there was also a typo.

Tasks: 
- https://cognitedata.atlassian.net/browse/FO-3487
- https://cognitedata.atlassian.net/browse/FO-3513

## Bump

- [x] Patch
- [ ] Skip

### Improved

- [alpha] When running `cdf build` with InfieldCDMLocationConfig resources, Toolkit will now more strictly validate these.

### DIFF
--- a/cognite_toolkit/_cdf_tk/yaml_classes/infield_cdm_location_config.py
+++ b/cognite_toolkit/_cdf_tk/yaml_classes/infield_cdm_location_config.py
@@ -85,6 +85,8 @@ class ViewMappings(BaseModelResource):
     activity: ViewMapping | None = None
 
     file: ViewMapping | None = None
+    # As of 27/04-26, observation only supported for one view,
+    # but we keep the list for future flexibility.
     observation: list[ViewMapping] | None = Field(None, min_length=1, max_length=1)
 
 
--- a/cognite_toolkit/_cdf_tk/yaml_classes/infield_cdm_location_config.py
+++ b/cognite_toolkit/_cdf_tk/yaml_classes/infield_cdm_location_config.py
@@ -1,4 +1,4 @@
-from typing import Any, Literal
+from typing import Literal
 
 from pydantic import Field
 
@@ -88,6 +88,12 @@ class ViewMappings(BaseModelResource):
     observation: list[ViewMapping] | None = None
 
 
+class DataExplorationConfig(BaseModelResource):
+    """Data exploration configuration."""
+
+    asset_properties_card: ViewMapping | None = None
+
+
 class Discipline(BaseModelResource):
     """Discipline configuration."""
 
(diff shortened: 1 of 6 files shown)


cognite_toolkit/_cdf_tk/yaml_classes/infield_cdm_location_config.py
```diff
@@ -85,7 +85,13 @@ class ViewMappings(BaseModelResource):
     activity: ViewMapping | None = None
 
     file: ViewMapping | None = None
-    observation: list[ViewMapping] | None = None
+    observation: list[ViewMapping] | None = Field(None, min_length=1, max_length=1)
```
DiemasMichiels, line 90: Not a data model expert, but is there a reason why this is still a list if only a length of 1 item is allowed?
If it is due to reasons maybe add a note to this.

laylafy: Yes, we only support one observation view for now in app. I added it as a comment

DiemasMichiels: Ok, I forgot it.
Could we add it as a comment to the code? It is easier to read next to the line where it is initiated.
If it is a list, would `observations` (with **s**) not be a better name?

Magssch:
Looks good, I made a slight modification to the changelog (it's written from an end-user endpoint).

cognite_toolkit/_cdf_tk/yaml_classes/infield_cdm_location_config.py
Magssch: My two cents: I guess the plural (s) would make more sense if the field was called `observationView` or similar. `observations` could, in contrast, also give the impression that it's referring to individual **instances** of an Observation view. So `observation` isn't necessarily more clear in that sense, but at least it is consistent with the other fields. I would suggest to keep it for now

Magssch (approved):

tests/data/complete_org_alpha_flags/modules/my_example_module/cdf_applications/my_location.InFieldCDMLocationConfig.yaml
```diff
@@ -75,15 +79,7 @@ disciplines:
...
-  searchEnabled: true
-  filters:
-  - type: status
-    values:
-    - active
-    - inactive
-  - type: location
-    values:
-    - plant_a
-    - plant_b
```
atussak, line 89: Out of curiosity, what is this stuff you removed? 🤔

laylafy: This is an example of how `dataExplorationConfig` should be configured. All of these are deprecated since we don't use `dataExplorationConfig` for anything a the moment, so I removed them from the example yaml after making the validation.

atussak: I see, thank you!

PR #2959, opened 2026-04-28
Improve build module loading output

ronpal:
# Description

First round of improvements in the build v2 output by grouping summary information, selection context, and module issue details in Toolkit-styled Rich panels. Also centralizes Toolkit Rich helpers so panels, embedded sections, tables, and questionary styling use shared defaults. Not exactly a design system, but an attempt at unifying. 

without --verbose
Screenshot 2026-04-28 at 09 33 00

with --verbose
Screenshot 2026-04-28 at 09 33 26

## Bump

- [ ] Patch
- [x] Skip

### Added/Changed/Improved/Removed/Fixed

- Improved build module loading output with clearer selection, loaded-module, and issue-detail sections.


dev_build.py
```diff
@@ -0,0 +1,41 @@
+"""Quick dev script: run a local build of complete_org without CDF auth.
```
Magssch: Remember to delete

cognite_toolkit/_cdf_tk/commands/build_v2/build_v2.py
```diff
@@ -244,97 +263,116 @@ def _display_module_sources(self, build_source: BuildSource, console: Console, v
...
-        border_color = 0
-        errors: list[str] = []
-        if read_variables:
-            summary_lines.append(f"[green]✓[/] [bold]{read_variables}[/] variables used in the build")
-        if ambiguous_selected_count:
-            summary_lines.append(
-                f"[red]✗[/] [bold]{ambiguous_selected_count}[/] user-selected modules had an ambiguous match with multiple module directories."
-            )
-            border_color = 2
-            errors.append("ambiguous selected")
```
Magssch, line 261: This error and the one below is no longer getting appended to the list of errors. The errors list seems unused now even though it is still instantiated and has a code path checking for it further down.

From testing locally, I can see that we are now giving a separate warning earlier in the output, but I think this gets a bit hidden the way it is setup right now, the final message seems very "successful" to me if that makes sense (see image 1) due to the wording and coloring used now. I think the previous warning was more clear in that there was an actual failure (see image 2).

After:

Image

Before:

Image

cognite_toolkit/_cdf_tk/commands/build_v2/build_v2.py
```diff
@@ -244,97 +263,116 @@ def _display_module_sources(self, build_source: BuildSource, console: Console, v
...
+
+        if verbose and issue_details_section_content:
+            summary_sections.append(ToolkitPanelSection(title="Issue details", content=issue_details_section_content))
+
+        border_style = {0: "green", 1: "yellow", 2: "red"}[border_color]
+        console.print(
+            ToolkitPanel(Group(*summary_sections), title="[bold]Loading modules[/]", border_style=border_style)
+        )
 
         if errors:
```
Magssch, line 379: (see comment above) This seems like a dead code path now

cognite_toolkit/_cdf_tk/commands/build_v2/build_v2.py
```diff
@@ -244,97 +263,116 @@ def _display_module_sources(self, build_source: BuildSource, console: Console, v
...
                     paths_str = ", ".join(p.as_posix() for p in selection.module_paths)
                     table.add_row(selection.name, paths_str)
-            console.print(table)
+            issue_details_section_content.append(table.as_panel_detail())
+            border_color = max(border_color, 2)
 
         if misplaced_modules_count:
-            table = Table(title="Misplaced Modules", expand=False, show_edge=False)
+            issue_summary_section_content.append(
+                f"[yellow]![/] [bold]{misplaced_modules_count}[/] modules are located directly under the another module (misplaced modules)."
```
Magssch: Preexisisting typo, but let's fix it 
```suggestion
                f"[yellow]![/] [bold]{misplaced_modules_count}[/] modules are located directly under another module (misplaced modules)."
```

cognite_toolkit/_cdf_tk/ui.py
```diff
@@ -0,0 +1,149 @@
...
+from rich.padding import Padding, PaddingDimensions
+from rich.panel import Panel
+from rich.style import StyleType
+from rich.table import Column, Table
+from rich.text import Text
+
+__all__ = ["QUESTIONARY_STYLE", "ToolkitPanel", "ToolkitPanelSection", "ToolkitTable"]
+
+
+class ToolkitPanel(Panel):
```
Magssch, line 15: Would it be an option to pass through kwargs instead, only explicitly declaring those we want to override defaults for? E.g.
```
class ToolkitPanel(Panel):
    def __init__(self, renderable, box=rich_box.ROUNDED, *, title_align="left", padding=(1, 2), **kwargs):
        super().__init__(renderable, box, title_align=title_align, padding=padding, **kwargs)
```

Magssch: Primarily because we don't override any defaults for the other params currently, so IDE/Intellisense support etc. wouldn't be an issue and it's quite a lot of passthrough currently. But it's a very optional comment.

cognite_toolkit/_cdf_tk/ui.py
```diff
@@ -0,0 +1,149 @@
...
+        renderables: list[RenderableType] = []
+        if header:
+            renderables.append(header)
+        renderables.extend(content or [])
+        renderables.append("")
+
+        super().__init__(*renderables)
+
+
+class ToolkitTable(Table):
```
Magssch, line 55: Same comment as for ToolkitPanel

cognite_toolkit/_cdf_tk/ui.py
```diff
@@ -0,0 +1,149 @@
...
+    ) -> None:
+        header = f"[bold]{title}:[/]" if title else ""
+        if description:
+            header = f"{header} {description}".strip()
+
+        renderables: list[RenderableType] = []
+        if header:
+            renderables.append(header)
+        renderables.extend(content or [])
+        renderables.append("")
```
Magssch: This newline seems to cause the bottom to always be bigger, we can probably remove it?

cognite_toolkit/_cdf_tk/commands/build_v2/build_v2.py
Magssch: I notice that it's much more clear when using --verbose, but still I am a bit concerned with the messaging in the boxes below (and that this would be much more visible than the previous message about module lookup) I think maybe continuing to throw an error like we did before is best until we potentially refactor that further?

Image

tests/test_unit/test_cdf_tk/test_tk_rich.py
```diff
@@ -0,0 +1,18 @@
+from rich import box
+
+from cognite_toolkit._cdf_tk.ui import ToolkitPanel
+
+
+def test_toolkit_panel_defaults_title_align_left_and_expand_true() -> None:
```
Magssch: These tests are testing passing through and setting variables in the underlying Panel object. I would suggest we remove them for now or test some edge cases perhaps?

cognite_toolkit/_cdf_tk/commands/build_v2/build_v2.py
ronpal: Nice catch, I got a bit too eager there. Restoring it.

cognite_toolkit/_cdf_tk/commands/build_v2/build_v2.py
ronpal: Fixed

cognite_toolkit/_cdf_tk/ui.py
ronpal: Good call, much cleaner now.

tests/test_unit/test_cdf_tk/test_tk_rich.py
ronpal: Agreed. Added more useful tests.

Magssch (approved):

PR #2960, opened 2026-04-28
Fix instance space validation in dataset-based migration path

Magssch:
# Description

`cdf migrate assets/events/timeseries/files --data-set <ds>` runs a pre-flight check to verify the target instance space exists before starting the migration. Due to a wrong attribute reference (`space_source.space` instead of `space_source.instance_space`), the check was always validating `"cognite_migration"` (the node in `SpaceSource`'s own space) rather than the actual target instance space of the mapping (which is on the `instanceSpace` property in `SpaceSource`. The migration would then proceed, and if the target instance space was missing, users would get a failure mid-migration instead of an early, actionable error.

## Bump

- [x] Patch
- [ ] Skip

### Fixed

- [alpha] When running the `cdf migrate assets/events/timeseries/files` commands, toolkit will now validate ahead of time whether the target instance space is missing, instead of failing mid-migration.


ronpal (approved):
Nais

PR #2965, opened 2026-04-29
[CDF-27847] ⬆ Cognite SDK V8

doctrino:
# Description

Changes to go from v7 to v8 of SDK.

* The `ToolkitAPI` which is used for the `LookUpAPI` and `VerifyAPI` (both utility APIs), no longer subclass `APIClient` from the `CogniteClient`, instead they just use a `ToolkitClient` directly. The `APIClient` was private and is now async so cannot be used by us.
* `ToolkitClientConfig` is now required in the init of the `ToolkitClient` as a result of the above change.
* Updated all test that uses `responses` to test the `request` call made by the `cognite-sdk` updated to use `respx` as the `cognite-sdk` now uses `httpx` instead of `requests`. 
* The responses classes we mock in the tests now have `createdTime`, `lastUpdatedTime` and other server set properties required. We still have several tests that use the `cognite-sdk` classes.

## Bump

- [x] Patch
- [ ] Skip

### Changed

- Toolkit now requires `cognite-sdk>v8`.


tests/test_unit/test_cli/test_build_deploy_snapshots/complete_org.yaml
```diff
@@ -514,21 +514,8 @@ Node:
   instanceType: node
   sources:
   - properties:
-      aliases: null
```
doctrino, line 517: Fix in v8 of SDK. No longer writing null values. 

Note this is irrelevant in v0.8 of Toolkit as this is only used for deploy v1.

cognite_toolkit/_cdf_tk/client/api/function_schedules.py
```diff
@@ -130,7 +132,9 @@ def iterate(
         """
         return self._iterate(limit=limit, body=self._create_list_request_body(function_id=function_id))
 
-    def list(self, function_id: int | None = None, limit: int | None = None) -> list[FunctionScheduleResponse]:
+    def list(
+        self, function_id: int | None = None, name: str | None = None, limit: int | None = None
```
doctrino, line 136: Needed name for a test. There is a bug in v8 when converting FunctionSchedules from response to request type.

cognite_toolkit/_cdf_tk/dataio/_datapoints.py
```diff
@@ -107,12 +107,9 @@ def stream_data(
         for timeseries in self.client.time_series(
             data_set_external_ids=[selector.data_set_external_id],
             chunk_size=self.DOWNLOAD_CHUNK_SIZE,
-            is_string=True if selector.data_type == "string" else False,
+            is_string=True if selector.data_type.lower() == "string" else False,
             advanced_filter=Exists(TimeSeriesProperty.external_id),
             limit=limit,
-            # We cannot use partitions here as it is not thread safe. This spawn multiple threads
-            # that are not shut down until all data is downloaded. We need to be able to abort.
-            partitions=None,
```
doctrino, line 115: No longer possible to set `partition` in this context.

ronpal (approved):

PR #2968, opened 2026-04-30
[CDF-27841] Fix incorrect key name in default_env deprecation warning

ronpal:
# Description

The deprecation warning for `--env / -e` / `default_env` told users to set `config_yaml` in `cdf.toml`, but the correct key is `default_config_yaml`. Following the warning verbatim caused the config to be silently ignored, falling back to the `dev` environment with no error.

## Bump

- [x] Patch
- [ ] Skip

### Fixed

- Deprecation warning for `default_env` now correctly refers to `default_config_yaml` (not `config_yaml`) in `cdf.toml`

### DIFF
--- a/cognite_toolkit/_cdf_tk/utils/fileio/_writers.py
+++ b/cognite_toolkit/_cdf_tk/utils/fileio/_writers.py
@@ -259,7 +259,7 @@ class CSVWriter(TableWriter[TextIOWrapper]):
         return csv_writer
 
 
-class ParquetWriter(TableWriter["pq.ParquetWriter"]):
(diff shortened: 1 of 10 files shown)


cognite_toolkit/_cdf_tk/utils/fileio/_writers.py
```diff
@@ -259,7 +259,7 @@ def _create_dict_writer(self, writer: IOBase) -> csv.DictWriter:
         return csv_writer
 
 
-class ParquetWriter(TableWriter["pq.ParquetWriter"]):  # type: ignore[type-var]
```
ronpal, line 262: Unrelated (boy scout principle)

Magssch (approved):

PR #2969, opened 2026-04-30
[CDF-27845] Improve deploy command output styling

ronpal:
## Description

Third round of CLI output polish, following the build v2 output improvements in #2959 (tidy-build-output) and the in-progress tidy-build-output-part-II branch.

Replaces bare `rich.panel.Panel` usage in `DeployCommand` with `ToolkitPanel` from the shared `ui.py` design-component module. Also:
- Adds a `title=` to every panel (consistent with build output)
- Consolidates the three separate drop-mode `Panel` calls in `clean_all_resources` into a single panel with dynamic flag text
- Removes the now-unused `from rich.panel import Panel` import

No behaviour changes. Will address the resource diffs in a separate PR. 

https://github.com/user-attachments/assets/4f6a4927-e9fc-4707-99cc-2c947d620c35

## Bump

- [x] Patch
- [ ] Skip

### Improved

- Cleaner, more readable output of the deploy command


cognite_toolkit/_cdf_tk/commands/deploy_v2/command.py
```diff
@@ -701,16 +730,15 @@ def _categorize_resources(
...
-                            diff_str,
-                            title=f"{crud.display_name}: {identifier}",
-                            expand=False,
-                        )
+                        old_lines = [line.replace(sensitive, "********") for line in old_lines]
+                        new_lines = [line.replace(sensitive, "********") for line in new_lines]
+                    diff = "".join(
+                        difflib.unified_diff(old_lines, new_lines, fromfile="CDF (current)", tofile="Local (desired)")
                     )
+                    console.print(ToolkitPanel(diff, title=f"{crud.display_name}: {identifier}"))
```
Magssch: This hunk replaces to_diff(cdf_dict, resource.raw_dict) with an inline difflib.unified_diff over yaml_safe_dump(...). The original to_diff 1 passes sort_keys=True, but the new code uses the default yaml_safe_dump where sort_keys=False. The result is that any key-order difference between what CDF returns and what's defined locally now shows up in the diff, even when the content is semantically identical — so the rendered diff is now exploding in size for unchanged resources

Was this change meant to be included in this PR or left for later? (ref. the PR description)

I would suggest to either revert to to_diff(...) for this PR or pass sort_keys=True to both yaml_safe_dump calls.

ronpal: You're absolutely right, will put this in a separate PR. Reverting.

Magssch (approved):

PR #2980, opened 2026-05-06
[CDF-27866] Fix duplicate-400 on DataProductVersion redeploy by diffing views before update

ronpal:
## Description

When redeploying a `DataProductVersion` in replace mode, the toolkit was sending all desired views in a single `views.add` payload — including views already present in CDF. Because `views.add` is **append-only** per the API spec, CDF rejects the request with a **400 duplicate error**.

A previous fix attempt (0.8.60) tried switching to `views.set`, but the `DataProductVersion` update API does not support `set` for views — `DataProductVersionPatch` only defines `views.add`.

### Root cause

The API contract (confirmed from the OpenAPI spec) is:
- View refs are **immutable once added** — they cannot be modified or removed.
- The only supported update operation is `views.add`.

### Fix

Before calling the update endpoint in replace mode, the toolkit now retrieves the current version from CDF and diffs the desired views against the existing ones (keyed by `space + externalId + version`). Only **net-new views** are sent in `views.add`. If the local file drops a view that is already in CDF, it is silently ignored — the API does not support removal.

Patch mode behaviour is unchanged (views are only included if explicitly set, which already prevents spurious sends).

### Files changed

- `cognite_toolkit/_cdf_tk/client/resource_classes/data_product_version.py` — `as_update` accepts an optional `cdf_views` list and computes the add-only diff
- `cognite_toolkit/_cdf_tk/client/api/data_product_versions.py` — `update()` fetches the current version in replace mode and passes its views to `as_update`
- `tests/test_unit/test_cdf_tk/test_cruds/test_data_product_version.py` — updated tests covering: add net-new only, unchanged views omitted, empty CDF sends all, view dropped locally is silently ignored

## Bump

- [x] Patch
- [ ] Skip

### Fixed

- `DataProductVersion` redeploy no longer fails with a 400 duplicate error when views are already present in CDF. The update now diffs desired views against CDF and only sends `views.add` for net-new entries.

### DIFF
--- a/cognite_toolkit/_cdf_tk/client/api/data_product_versions.py
+++ b/cognite_toolkit/_cdf_tk/client/api/data_product_versions.py
@@ -103,14 +103,31 @@ class DataProductVersionsAPI(CDFResourceAPI[DataProductVersionResponse]):
         items: Sequence[DataProductVersionRequest],
         mode: Literal["patch", "replace"] = "replace",
     ) -> list[DataProductVersionResponse]:
+        """Apply updates; in ``replace`` mode we retrieve first so ``as_update`` can diff views.
+
+        The patch API only supports ``views.add`` (no full replace). We need the live view list
+        to compute add/remove keys and avoid resending existing refs (duplicate 400).
+        """
         results: list[DataProductVersionResponse] = []
         for dp_ext_id, group in self._group_by_parent(items).items():
             url = self._make_url(self._method_endpoint_map["update"].path.format(externalId=dp_ext_id))
             for item in group:
+                cdf_views = None
+                if mode == "replace":
+                    current = self.retrieve(
+                        [
+                            DataProductVersionId(
+                                data_product_external_id=dp_ext_id,
+                                version=item.version,
+                            )
+                        ],
+                        ignore_unknown_ids=True,
+                    )
+                    cdf_views = current[0].views if current else []
                 request = RequestMessage(
                     endpoint_url=url,
                     method="POST",
-                    body_content={"items": [item.as_update(mode=mode)]},
+                    body_content={"items": [item.as_update(mode=mode, cdf_views=cdf_views)]},
                     api_version=self._api_version,
                 )
                 response = self._http_client.request_single_retries(request)
--- a/cognite_toolkit/_cdf_tk/client/api/data_product_versions.py
+++ b/cognite_toolkit/_cdf_tk/client/api/data_product_versions.py
@@ -111,19 +111,12 @@ class DataProductVersionsAPI(CDFResourceAPI[DataProductVersionResponse]):
         results: list[DataProductVersionResponse] = []
         for dp_ext_id, group in self._group_by_parent(items).items():
             url = self._make_url(self._method_endpoint_map["update"].path.format(externalId=dp_ext_id))
+            cdf_views_by_version: dict[str, list] = {}
+            if mode == "replace":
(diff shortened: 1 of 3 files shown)


cognite_toolkit/_cdf_tk/client/api/data_product_versions.py
```diff
@@ -103,14 +103,31 @@ def update(
...
+                    current = self.retrieve(
+                        [
+                            DataProductVersionId(
+                                data_product_external_id=dp_ext_id,
+                                version=item.version,
+                            )
+                        ],
+                        ignore_unknown_ids=True,
+                    )
+                    cdf_views = current[0].views if current else []
```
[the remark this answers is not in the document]
ronpal: Good catch, fixed. We now collect all version IDs for the group, call `retrieve` once per data-product group, and build a `{version: views}` dict before the inner loop — so it's a single bulk fetch regardless of group size.

cognite_toolkit/_cdf_tk/client/api/data_product_versions.py
```diff
@@ -54,19 +54,19 @@ def _group_by_parent(
 
     def create(self, items: Sequence[DataProductVersionRequest]) -> list[DataProductVersionResponse]:
         results: list[DataProductVersionResponse] = []
-        for dp_ext_id, group in self._group_by_parent(items).items():
-            url = self._make_url(self._method_endpoint_map["create"].path.format(externalId=dp_ext_id))
-            for item in group:
+        for data_product_external_id, versions in self._group_by_parent(items).items():
```
ronpal, line 57: renamed for readability

cognite_toolkit/_cdf_tk/client/api/data_product_versions.py
```diff
@@ -103,29 +103,40 @@ def update(
...
         mode: Literal["patch", "replace"] = "replace",
     ) -> list[DataProductVersionResponse]:
+        """Apply updates; in ``replace`` mode we retrieve first so ``as_update`` can diff views. The patch API only supports ``views.add`` (no full replace). We need the live view list to compute add/remove keys and avoid resending existing refs (duplicate 400)."""
         results: list[DataProductVersionResponse] = []
-        for dp_ext_id, group in self._group_by_parent(items).items():
-            url = self._make_url(self._method_endpoint_map["update"].path.format(externalId=dp_ext_id))
-            for item in group:
+        for data_product_external_id, versions in self._group_by_parent(items).items():
+            url = self._make_url(self._method_endpoint_map["update"].path.format(externalId=data_product_external_id))
+            views_by_version: dict[str, list] = {}
```
Magssch, line 110: nitpick
```suggestion
            views_by_version: dict[str, list[DataProductVersionView]] = {}
```

Magssch:
One small comment, since DataProductVersion only supports adding views, it seems like a similar case to properties in containers. We print out warnings for that if the local YAML drifts from the CDF definition, while here it seems we swallow that silently. Optional, we can consider it for a later PR, but would it be an option to add something like this type of warning if there's a view missing locally:

_DataProductVersion <id> has differing views in your local YAML as opposed to CDF. This may occur if you have tried to remove previously deployed view refs from a DataProductVersion in your local YAML file. Please note that the DataProductVersion API does not support removing existing viewss (they are append-only). This warning will persist until the discrepancies are resolved. • Views in CDF but not in your local YAML: <space:externalId/version, ...>_

Plus the same HINT_LEAD_TEXT line pointing at `cdf modules pull`

Magssch (approved):

PR #2983, opened 2026-05-06
[CDF-27869] Fix confusing defaults in cdf auth init selection prompts

ronpal:
# Description

Fixes confusing defaults and UX issues in the `cdf auth init` interactive flow that caused users following naive defaults to end up with an invalid or incompatible configuration.

**Changes:**

- **Reorder login flow choices** — `device_code` is now first and the default for new users. It only requires a tenant ID and a browser sign-in, making it the most approachable starting point. `client_credentials` (the old default) requires a pre-existing service principal with client ID + secret, which most first-time users don't have.
- **Clearer login flow descriptions** — all four flows now have concise, action-oriented labels.
- **Validate CDF cluster and project inputs** — empty values are now rejected inline rather than silently accepted and failing at the end.
- **Cluster prompt shows an example** — the prompt now reads `Enter the CDF cluster (e.g. westeurope-1)`.
- **Example strings are hints, not defaults** — fields like `IDP_TENANT_ID` previously pre-filled the input with `"00000000-0000-0000-0000-000000000000 or mytenant.onmicrosoft.com"`, which users could accidentally submit verbatim. These are now shown as `(e.g. ...)` label text; only concrete resolved values are pre-filled.
- **Fix display name** — `"Tenant id for MS Entra"` → `"Tenant id for Microsoft Entra ID"`.
- **`--reset` flag** — hidden `cdf auth init --reset` flag to start the wizard from scratch, ignoring existing env vars. Any existing `.env` is automatically backed up without a confirmation prompt.
- **"Additional variables:" header** — printed before the derived optional variable list so users understand what they're looking at.

## Bump

- [x] Patch
- [ ] Skip

### Fixed

- `cdf auth init` now defaults to the `device_code` login flow (browser sign-in, no service principal required) instead of `client_credentials`, preventing first-time users from failing at the credential-verification step.
- Example/hint strings (e.g. tenant ID placeholder) are no longer pre-filled as default values in prompts — they are shown as label hints instead.
- CDF cluster and project prompts now reject empty input inline.

### DIFF
--- a/cognite_toolkit/_cdf_tk/utils/auth.py
+++ b/cognite_toolkit/_cdf_tk/utils/auth.py
@@ -510,13 +510,14 @@ def prompt_user_environment_variables(current: EnvironmentVariables | None = Non
             idp_tenant_id = user_value
 
     optional_values = env_vars.get_optional_with_value()
-    print("Additional variables:")
-    for field_, value in optional_values:
-        print(f"  {field_.name}={value}")
-    if questionary.confirm("Do you want to change any of these variables?", default=False).unsafe_ask():
+    if optional_values:
+        print("Additional variables:")
         for field_, value in optional_values:
-            user_value = get_user_value(field_, value, provider, cdf_cluster, cdf_project, idp_tenant_id)
-            setattr(env_vars, field_.name, user_value)
+            print(f"  {field_.name}={value}")
+        if questionary.confirm("Do you want to change any of these variables?", default=False).unsafe_ask():
+            for field_, value in optional_values:
+                user_value = get_user_value(field_, value, provider, cdf_cluster, cdf_project, idp_tenant_id)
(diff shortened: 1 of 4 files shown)


cognite_toolkit/_cdf_tk/utils/auth.py
```diff
@@ -502,6 +510,7 @@ def prompt_user_environment_variables(current: EnvironmentVariables | None = Non
             idp_tenant_id = user_value
 
     optional_values = env_vars.get_optional_with_value()
+    print("Additional variables:")
```
Magssch: Nitpicking: This would also print even if there are no additional variables

ronpal: Good catch, fixed — the header and confirm prompt are now guarded by `if optional_values`.

Magssch (approved):

PR #2984, opened 2026-05-06
feat(deploy): human-readable deploy drift (--diff human)

ronpal:
# Description

**What you get:** With `cdf deploy … --diff human`, each resource that differs from CDF is shown in one **Rich panel**: identifier and file path, a short diff summary, and **two columns of YAML** (**Local build** vs **CDF**, with the project name on the CDF side). Sensitive values are masked the same way as in `--verbose`; for those resources the long unified verbose diff is **not** printed, so output stays scannable.

**Also in this PR:** Deploy summary rows are ordered by impact (creates and updates before unchanged). Human diff YAML keeps build key order and pairs keys sensibly (including a fix when one side is a dict and the other a scalar). Hosted extractor source/destination warnings explain that CDF is compared by external id only. The processed build-directory block uses the same nesting as the plan section so bullets line up. `ToolkitPanelSection` gains optional `content_padding`, with padding tuple behavior documented on the panel helpers.

Demo with and without human diff:

https://github.com/user-attachments/assets/a29e724f-33d8-4c39-ae42-4e901aae6ade

## Bump

- [x] Patch
- [ ] Skip

### Added

- `cdf deploy --diff=human` and `deploy_v2/diff_display.py` for structured drift output.

### Changed

- Deployment summary table: sort rows by descending created, updated, deleted, then unchanged.

### Improved 

- Hosted extractor source/destination: clearer warning text for id-only CDF comparison.

### Fixed

- Human diff YAML alignment no longer replaces a scalar with `{}` when the other side is a dict for the same key.

### DIFF
--- a/cognite_toolkit/_cdf_tk/apps/_core_app.py
+++ b/cognite_toolkit/_cdf_tk/apps/_core_app.py
@@ -22,6 +22,7 @@ from cognite_toolkit._cdf_tk.commands import (
     BuildV2Command,
     CleanCommand,
     DeployCommand,
+    DeployDiffFormat,
     DeployOptions,
     DeployV2Command,
 )
@@ -520,6 +521,14 @@ class CoreApp(typer.Typer):
                 help="Turn on to get more verbose output when running the command",
             ),
         ] = False,
+        diff: Annotated[
+            DeployDiffFormat | None,
+            typer.Option(
+                "--diff",
+                help="When set to 'human', show a ToolkitPanel summary and a side-by-side line comparison of "
+                "CDF (API) dump vs build YAML for each resource that would change.",
+            ),
+        ] = None,
     ) -> None:
         """Deploys the configuration files in the build directory to the CDF project."""
         if drop:
@@ -548,6 +557,7 @@ class CoreApp(typer.Typer):
                     include=include,
                     force_update=force_update,
                     verbose=verbose,
+                    diff=diff,
                     environment_variables=env_vars.dump(),
                     deployment_dir=deploy_dir,
                 ),
--- a/cognite_toolkit/_cdf_tk/commands/deploy_v2/diff_display.py
+++ b/cognite_toolkit/_cdf_tk/commands/deploy_v2/diff_display.py
@@ -23,6 +23,35 @@ class DeployDiffFormat(str, Enum):
 _SECTION_BODY_PADDING = (0, 0, 1, 2)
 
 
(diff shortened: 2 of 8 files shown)


cognite_toolkit/_cdf_tk/apps/_core_app.py
```diff
@@ -520,6 +521,14 @@ def deploy_v2(
                 help="Turn on to get more verbose output when running the command",
             ),
         ] = False,
+        diff: Annotated[
+            DeployDiffFormat | None,
+            typer.Option(
+                "--diff",
+                help="When set to 'human', show a ToolkitPanel summary and a side-by-side line comparison of "
+                "CDF (API) dump vs build YAML for each resource that would change.",
+            ),
```
doctrino, line 530: New feature, alpha flag?

cognite_toolkit/_cdf_tk/commands/deploy_v2/diff_display.py
```diff
@@ -0,0 +1,195 @@
...
+            for idx in range(i1, i2):
+                yield (Text(cdf_lines[idx], style="red"), Text("-", style="dim"))
+        elif tag == "insert":
+            for idx in range(j1, j2):
+                yield (Text("-", style="dim"), Text(yaml_lines[idx], style="green"))
+        elif tag == "replace":
+            left = cdf_lines[i1:i2]
+            right = yaml_lines[j1:j2]
+            for k in range(max(len(left), len(right))):
+                l_txt = left[k] if k < len(left) else ""
```
Magssch, line 126:
```suggestion
                left_text = left[k] if k < len(left) else ""
```

cognite_toolkit/_cdf_tk/commands/deploy_v2/diff_display.py
```diff
@@ -0,0 +1,195 @@
...
+                r_txt = right[k] if k < len(right) else ""
```
Magssch, line 127:
```suggestion
                right_text = right[k] if k < len(right) else ""
```

cognite_toolkit/_cdf_tk/commands/deploy_v2/diff_display.py
```diff
@@ -0,0 +1,195 @@
...
+                yield (Text(l_txt, style="yellow"), Text(r_txt, style="cyan"))
```
Magssch, line 128:
```suggestion
                yield (Text(left_text, style="yellow"), Text(right_text, style="cyan"))
```

PR #3002, opened 2026-05-12
[CDF-27948] Make Infield config migration tolerate invalid root location configs

Magssch:
# Description

`cdf migrate infield-configs` now skips root location configurations that are missing required fields or whose root asset has not been migrated yet, instead of failing the entire command on the first bad entry. The purpose of this is to enable partial migrations to CDM and to avoid any potential single misconfigured root location to block an entire project migration. Each skipped location now emits a warning, and a summary panel is printed at the end of the run grouping skips by reason, which should somewhat mitigate the now increased risk of users not acknowledging that a location was skipped due to what could be a real data integrity issue that needs to be addressed.

## Bump

- [x] Patch
- [ ] Skip

### Improved

- [alpha] `cdf migrate infield-configs` now migrates the root location configurations it can and reports successes and failures as a summary, instead of aborting on the first invalid or non-migrated location.


ronpal (approved):
Good stuff!

PR #3006, opened 2026-05-18
[CDF-27549] Add App Hosting API client layer

Magssch:
# Description

Adds `AppsAPI` and `AppVersionsAPI` client classes for the CDF App Hosting API, along with
`AppRequest`/`AppResponse` and `AppVersionRequest`/`AppVersionResponse` resource classes (split
across `app.py` and `app_version.py` to mirror the API endpoint split). The HTTP client gains a
shared `_execute_raw_with_retries` method and a new `request_multipart_retries` helper for
multipart/form-data uploads, needed by the version upload endpoint. A hidden `CUSTOM_APPS` feature
flag is added to gate future deployment support.

The YAML layer is split into two separate kinds — `AppYAML` (`.App.yaml`) and `AppVersionYAML`
(`.AppVersion.yaml`) — mirroring the API's resource model and following the same pattern as
`DataProduct`/`DataProductVersion` and `Workflow`/`WorkflowVersion`.

Example of App YAML:
Screenshot 2026-05-19 at 12 03 33

Example of AppVersion YAML:
Screenshot 2026-05-19 at 12 03 49

`sourcePath` is a pseudo-field, only used by toolkit to refer to the relative path of the source files of the Custom App. This can _either_ refer to the root of the source files themselves _or_ the transpiled `/dist` subfolder (generated by `npm run build`)

## Bump

- [ ] Patch
- [x] Skip


ronpal (approved):
🐻 👍

PR #3015, opened 2026-05-19
[CDF-27986] Fix InfieldOnCDM deploy when legacy infield was not deployed previously

Magssch:
# Description

Deploying an InField CDM location config always called `apm_config.list()` to check for conflicts with legacy InField `appDataInstanceSpace`. This check requires the `APM_Config/1` view to exist AND for the user to have access to the `APM_Config` space both to read data models and data model instances. On DM-only projects (or hybrid projects without legacy Infield deployed already) this view is not installed, causing a 400 error and deployment failure. This PR changes the logic to catche this specific error and returns an empty set when this happens. To make the catch unambiguous, `create_acl` now also declares `dataModels:read` and `dataModelInstances:read` on the `APM_Config` space as required capabilities — ensuring the 400 can **only** mean the view is absent, not that access is missing.

## Bump

- [x] Patch
- [ ] Skip

### Fixed

- Fixes an issue when running `cdf deploy` with InField CDM location configs failing with `ResourceCreationError` when deploying to a project with no legacy Infield config installed.


cognite_toolkit/_cdf_tk/resource_ios/_resource_ios/fieldops.py
```diff
@@ -392,10 +394,19 @@ def get_minimum_scope(cls, items: Sequence[InFieldCDMLocationConfigRequest]) ->
     def create_acl(cls, actions: set[Literal["READ", "WRITE"]], scope: ScopeDefinition) -> Iterable[AclType]:
         if isinstance(scope, AllScope | SpaceIDScope):
             yield DataModelInstancesAcl(actions=as_instance_acl_actions(actions), scope=scope)
+        # Reading APM_Config is always required to check for legacy InField conflicts.
+        yield DataModelsAcl(actions=["READ"], scope=SpaceIDScope(space_ids=[APM_CONFIG_SPACE]))
```
Magssch, line 398: These are now required in order to verify whether the space for legacy infield config exists or not (since non-access and non-existence give the same 400 error, we cannot allow users without access to the space to do this operation)

cognite_toolkit/_cdf_tk/resource_ios/_resource_ios/fieldops.py
```diff
@@ -392,10 +394,19 @@ def get_minimum_scope(cls, items: Sequence[InFieldCDMLocationConfigRequest]) ->
...
+        yield DataModelInstancesAcl(actions=["READ"], scope=SpaceIDScope(space_ids=[APM_CONFIG_SPACE]))
 
     @cached_property
     def _legacy_instance_spaces(self) -> set[str]:
-        apm_configs = self.client.infield.apm_config.list(limit=None)
+        try:
+            apm_configs = self.client.infield.apm_config.list(limit=None)
```
Magssch, line 404: We now skip in case of a 400 error, but enforce the ACLs above to ensure that the user doesn't by chance _not_ have access to the space even if it exists (which would have caused this check to go through as well)

ronpal (approved):

PR #3044, opened 2026-06-01
[CDF-28064] Fix GroupYAML rejecting attributes field at build/deploy

ronpal:
# Description

`GroupYAML` was missing the `attributes` field that `GroupRequest` already supports via the Groups API. Because `GroupYAML` inherits `extra="forbid"`, any group YAML with `attributes:` failed build/deploy validation with `Unknown field: 'attributes'`.

This adds `attributes: GroupAttributes | None` to `GroupYAML`, reusing the existing resource class model so the build-time schema matches the API-facing model.

## Bump

- [x] Patch
- [ ] Skip

### Fixed

- Fixed group YAML validation rejecting the `attributes` field (CDF-28064)


dev.py
```diff
@@ -130,8 +130,22 @@ def create_changelog_entry() -> None:
...
+    for item in marko.parse(pr_description).children:
+        if not isinstance(item, marko.block.Heading) or item.level != 2:
+            continue
+        if not isinstance(item.children[0], marko.inline.RawText):
+            continue
+        if item.children[0].children.strip().casefold() == "test plan":
+            print("PR description must not contain a '## Test plan' section.")
+            raise SystemExit(1)
+
+
```
ronpal, line 145: Housekeeping. Test plans here are included in releases, which looks bad

cognite_toolkit/_cdf_tk/yaml_classes/groups.py
```diff
@@ -19,6 +20,7 @@ class GroupYAML(ToolkitResource):
     name: str
     capabilities: list[Capability] | None = None
     metadata: dict[str, str] | None = None
+    attributes: GroupAttributes | None = None
```
ronpal, line 23: The fix

Magssch (approved):

PR #3047, opened 2026-06-03
[CDF-28069] Fix cdf migrate infield-configs incorrectly populating dataExplorationConfig

Magssch:
# Description

Remove the migration of legacy APM `featureConfiguration` fields (`documents`, `notifications`,
`asset_page_configuration`, `observations`) into `dataExplorationConfig` on the InField CDM
location config. Confirmed with the InField team that `dataExplorationConfig` must be configured manually post-migration — none of the old fields map to it in a way that
is actually consumed by InField on CDM. As an additional housekeeping task; aligns the migrated location filter `externalId`
prefix to `loc_` per the naming convention enforced by Value Delivery.

## Bump

- [x] Patch
- [ ] Skip

### Fixed

- [alpha] Fix `cdf migrate infield-configs` incorrectly populating `dataExplorationConfig` on migrated
  InField CDM location configs with fields from the legacy APM config that are no longer used by
  InField on CDM.

### DIFF
--- a/cognite_toolkit/_cdf_tk/commands/_migrate/creators.py
+++ b/cognite_toolkit/_cdf_tk/commands/_migrate/creators.py
@@ -20,7 +20,6 @@ from cognite_toolkit._cdf_tk.client.resource_classes.apm_config_v1 import (
     APM_CONFIG_SPACE,
     APMConfigResponse,
     Discipline,
-    FeatureConfiguration,
     RootLocationConfiguration,
     RootLocationFeatureToggles,
 )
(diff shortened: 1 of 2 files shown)


cognite_toolkit/_cdf_tk/commands/_migrate/creators.py
```diff
@@ -406,7 +402,7 @@ def _display_summary(self, success_count: int, skipped_external_ids_by_label: di
 
     def _create_location_filter(self, config: RootLocationConfiguration) -> LocationFilterRequest:
         original_external_id = config.external_id or config.asset_external_id or str(uuid.uuid4())
-        external_id = f"location_filter_{original_external_id}"
+        external_id = f"loc_{original_external_id}"
```
Magssch, line 405: New naming convention for location configs per VD team

ronpal (approved):

PR #3005, opened 2026-05-17
EMT-391: Add Kuiper Aliasing Engine

szymon-blumczynski-cognite:
# Description

Add the Kuiper Aliasing Engine - a comprehensive system for entity matching through configurable aliasing rules. This includes:

- **Aliasing Engine Core**: Rule-based system for transforming and normalizing entity attributes
- **Rule Library**: 8 different rule types (case transformation, character substitution, leading zero normalization, prefix/suffix handling, regex substitution, value expansion, and composite rules)
- **Assembly & Composition**: Framework for building Kuiper expressions with configurable rule composition
- **Bootstrap System**: Initialization logic for setting up the aliasing engine
- **YAML Configuration**: Reader for loading aliasing rules from YAML with proper error handling
- **Registry & Discovery**: Dynamic rule discovery and registry system
- **Common Utilities**: JSON Path and macro support for flexible attribute access

## Bump

- [ ] Patch
- [x] Skip

### Added

- Kuiper Aliasing Engine with rule-based entity attribute transformation
- 8 configurable aliasing rule types for data normalization
- Aliasing engine assembly and expression composition framework
- Bootstrap system for engine initialization
- YAML-based rule configuration with validation and error reporting
- Rule registry and dynamic discovery system
- JSON Path and macro variable support

PR #3049, opened 2026-06-03
Feature/emt 391 improvements

olacognite:
# Description

We want to add the EM aliasing deployment pack #3005 
This is a giant PR with some styling shortcomings and some dead code also. A claude summary of possible actions was given as follows: 

## Overall verdict
The functionality is well-tested (98% patch coverage) and cleanly isolated, but the code reads as if it were written in another language's idiom — Java/C#-style enterprise patterns transplanted onto a Python codebase that has its own, much lighter conventions. The single biggest risk for a team unfamiliar with the codebase is that future maintainers will have two competing styles to reconcile. None of this blocks merge given the feature flag, but several items are worth fixing before this becomes the template others copy.

## Structural deviations from the codebase
This is the heart of it. The existing toolkit is pragmatic Python; this PR imports a different design culture. Concretely:

### A dependency-injection container written by hand. 
bootstrap/bootstrapper.py is ~10 provide_* factory functions wiring objects together with x or provide_default() fallbacks. This is a Spring/Guice idiom. Nothing else in the toolkit does this — collaborators construct objects directly. For the actual call path (one command, one facade), this entire file could be a couple of constructor calls.

### Abstraction layers with exactly one implementation. 
There are five ABCs in the new tree (RuleDefinitionRegistry, RulesDiscovery, AliasingKuiperBuilder, AliasingKuiperBuilderFactory, ExpressionComposer), each with a single Default*/Local* concrete class. A Builder, plus a BuilderFactory that does nothing but call the builder's constructor, plus a Facade over the factory, is three layers to do one thing. The toolkit's existing commands subclass ToolkitCommand and implement execute() — flat and direct. A reader coming from the rest of the repo will spend real time tracing facade → factory → builder → registry → composer to find where work actually happens.

### Filesystem-glob + importlib rule discovery. 
registry/rules_discovery.py globs rules/*.py and dynamically imports each module to find RuleDefinition subclasses. This is the only place in _cdf_tk that uses importlib.import_module for discovery. The codebase's established idiom for the same problem is __subclasses__() (used in resource_ios, builders, yaml_classes, etc.). The dynamic version is slower (cold-start is already a known toolkit pain point — see CDF-27851), harder to trace statically, and silently swallows import failures into log warnings, so a broken rule just vanishes from the registry instead of failing loudly. A plain explicit registry dict, or __subclasses__() after importing the package, would match convention and remove ~80 lines.

### Single-string value objects. 
RuleName and RuleDescription in rules/base.py wrap one str each with a __post_init__ non-empty check. Per-rule context *Builder classes (e.g. CaseTransformationContextBuilder) do the same for one field. This is ceremony Python usually expresses with a validated field or a guard clause.

### Custom exception hierarchy that bypasses the toolkit's. 
io/errors.py defines YamlReadError / InvalidRuleFormatError deriving from bare Exception, and the builder/registry raise BuilderConstraintError, RuleDefinitionNotFoundError, and plain ValueError/FileNotFoundError. The CLI's top-level handler (_cdf.py:178) catches ToolkitError to render clean error messages. None of these inherit from ToolkitError, so a user pointing the command at a malformed YAML gets a raw Python traceback instead of a formatted toolkit error. This is a user-facing behavioral deviation, not just stylistic — worth fixing. The repo already has ToolkitFileNotFoundError, ToolkitValueError, etc. for exactly this.

### Not using the toolkit's file I/O helpers. 
YamlRulesReader.read_file takes a str path (rest of the codebase passes pathlib.Path) and uses raw open() + yaml.safe_load; the command writes output with Path.write_text. The toolkit has read_yaml_file, safe_read, safe_write, and yaml_safe_dump (used throughout modules.py and elsewhere) that handle encoding and error wrapping consistently.

### Concrete bugs / dead code
- Dead method:YamlRulesReader._validate_rules_key_exists (lines ~122–140) is defined but never called — it was superseded by _validate_root_structure. Remove it.
- Dead computation: in aliasing_kuiper_builder.py._resolve_composite_rules, resolved_specs is built up (.append(...)) and then never used — only expanded_sub_rules feeds the result. Either it's leftover from a refactor or it signals an intended check that got dropped. Worth confirming with the author which.
- Non-exhaustive match:CaseTransformationRuleDefinition.create_kuiper_macro matches UPPERCASE/LOWERCASE with no fallback. If a third CaseStrategy is ever added, expression is unbound → UnboundLocalError at runtime rather than a clear error. Add an else/case _.
- Unused public surface: the Rule / RuleName / RuleDescription classes in rules/base.py aren't referenced anywhere in shipping code (the real flow uses the separate AliasingRule dataclass). Two parallel "rule" types in one feature is confusing; drop the unused one.
- O(n²) duplicate check:_validate_rules does rule_names.count(name) inside a set comprehension over the same list. Trivial at current scale, but collections.Counter is the idiomatic one-liner.

## Bump

- [ ] Patch
- [ ] Skip

### Added/Changed/Improved/Removed/Fixed

See commits, this does not go into the changelog anyway, since we are merging this into another PR branch.

### DIFF
--- a/cognite_toolkit/_cdf_tk/commands/entity_matching/aliasing/bootstrap/bootstrapper.py
+++ b/cognite_toolkit/_cdf_tk/commands/entity_matching/aliasing/bootstrap/bootstrapper.py
@@ -1,26 +1,17 @@
 from cognite_toolkit._cdf_tk.commands.entity_matching.aliasing.api.facade import AliasingFacade
 from cognite_toolkit._cdf_tk.commands.entity_matching.aliasing.assembly.aliasing_kuiper_builder import (
-    AliasingKuiperBuilder,
     AliasingKuiperBuilderFactory,
-    DefaultAliasingKuiperBuilder,
     DefaultAliasingKuiperBuilderFactory,
 )
 from cognite_toolkit._cdf_tk.commands.entity_matching.aliasing.assembly.expression_composer import (
     DefaultExpressionComposer,
-    ExpressionComposer,
 )
 from cognite_toolkit._cdf_tk.commands.entity_matching.aliasing.assembly.expression_composition_config import (
     AliasingCompositionConfig,
     OutputProjectionConfig,
 )
-from cognite_toolkit._cdf_tk.commands.entity_matching.aliasing.registry.registry import (
-    LocalRuleDefinitionRegistry,
-    RuleDefinitionRegistry,
-)
-from cognite_toolkit._cdf_tk.commands.entity_matching.aliasing.registry.rules_discovery import (
-    LocalRulesDiscovery,
-    RulesDiscovery,
-)
+from cognite_toolkit._cdf_tk.commands.entity_matching.aliasing.registry.registry import LocalRuleDefinitionRegistry
+from cognite_toolkit._cdf_tk.commands.entity_matching.aliasing.registry.rules_discovery import LocalRulesDiscovery
 from cognite_toolkit._cdf_tk.commands.entity_matching.common.json_path import JSONPath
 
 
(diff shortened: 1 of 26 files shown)


cognite_toolkit/_cdf_tk/commands/entity_matching/aliasing/bootstrap/bootstrapper.py
```diff
@@ -45,52 +38,13 @@ def provide_aliasing_composition_config(
...
 def provide_aliasing_facade(
-    factory: AliasingKuiperBuilderFactory | None = None,
+    builder_provider: Callable[[], AliasingKuiperBuilder] | None = None,
 ) -> AliasingFacade:
-    resolved_factory = factory or provide_aliasing_kuiper_builder_factory()
-    return AliasingFacade(resolved_factory)
+    if builder_provider is None:
+        registry = LocalRuleDefinitionRegistry.bootstrap(LocalRulesDiscovery.create())
+        composer = DefaultExpressionComposer(provide_aliasing_composition_config())
+        resolved_builder_provider = lambda: DefaultAliasingKuiperBuilder(registry=registry, composer=composer)
```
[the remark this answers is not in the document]
olacognite: @gemini-code-assist addressed now in [ce59665](https://github.com/cognitedata/toolkit/pull/3049/commits/ce59665e5521c36166fafcda4c46760754413cb3) 

What do you think about the other changes and how they match the issues in the description?

cognite_toolkit/_cdf_tk/commands/entity_matching/aliasing/io/yaml_rules_reader.py
```diff
@@ -40,24 +44,23 @@ def read_file(self, file_path: str) -> RulesFileContent:
...
             raise YamlReadError(
                 f"Invalid YAML syntax: {e!s}",
-                file_path=file_path,
+                file_path=str(file_path),
             ) from e
         except Exception as e:
             raise YamlReadError(
                 f"Error reading file: {e!s}",
-                file_path=file_path,
+                file_path=str(file_path),
```
ronpal: `file_path.as_posix()` ensures correct printing on Windows :)

olacognite: [acb7d28](https://github.com/cognitedata/toolkit/pull/3049/commits/acb7d28b4190f7418294a6c219d25b6908b47a21)

PR #3051, opened 2026-06-04
[CDF-28070] Add Atlas AI Skills resource and agent skills field

ronpal:
## Description

Added support for agent skills as a resource type: https://api-docs.cognite.com/20230101-beta/tag/Skills/operation/skill_create_ai_skills_post/

It is a distinct API, but closely tied to agents. Since Skills are typically markdown files, I decided to support a sidecar file, much like SQL for transformations: 

```
# my.Skill.yaml
externalId: mad_skill
```

Supported sidecar paths: 

- my.Skill.md: same prefix as yaml
- mad_skill.Skill.md: prefix is external id

Content

```
---
name: some name
description: longer description
---
## The thing 

```

The markdown must follow a particular format with a frontmatter between `---`separators. I chose to parse those and validate with Pydantic.

> Frontmatter: string (Content) non-empty `^(?:\ufeff)?---\s*\n([\s\S]*?)\n---\s*\n([\s\S]*\S[\s\S]*)$`
Skill content in SKILL.md form: YAML frontmatter (between --- lines) followed by a Markdown body. The frontmatter must be a YAML object that includes name and description keys whose values match the top-level name and description fields on this request.

That means that name and description would be redundant in the yaml. 

## Bump

- [x] Patch
- [ ] Skip

### Added

- Toolkit support for Atlas AI Skills (`agent/` folder, `Skill` kind) and agent `skills` references.


cognite_toolkit/_cdf_tk/client/resource_classes/agent.py
```diff
@@ -229,6 +229,7 @@ class Agent(AgentObject):
     instructions: str | None = None
     model: str | None = None
     tools: list[AgentTool] | None = None
+    skills: list[str] | None = None
```
ronpal, line 232: Technically this should be behind the alpha flag, but I don't think it is worth it the effort or complexity.

cognite_toolkit/_cdf_tk/commands/pull.py
```diff
@@ -808,7 +812,14 @@ def _update(
                         if placeholder in extra_content:
                             new_extra = new_extra.replace(variable.value, f"{{{{ {variable.key} }}}}")
                     extra_files[extra.path] = new_extra
-        return replacer.replace(loaded, loaded_with_placeholder, item_write)
+        split_resources = list(replacer._loader.split_resource(source_file, item_write))
+        base_to_write = item_write
+        for split_path, split_content in split_resources:
```
Magssch: Running this for all resource types will now force any `Transformation.yaml` with inline SQL query to convert to a YAML with .sql sidecar when using `cdf modules pull`. I don't mind the change really as well really, but since it's currently an alpha flag feature we should probably gate this behind being run only for the Skills resource

Magssch:
We should likely add some dump functionality for skills as well to make this even more useful. We could consider having skills dumped automatically as a side-car whenever users run `cdf dump agents` or/and we could have a separate command for `cdf dump skills`. For a follow-up PR, but just noting it since there's no way to dump skills created in the UI currently

cognite_toolkit/_cdf_tk/yaml_classes/skill.py
```diff
@@ -0,0 +1,66 @@
...
+            name = frontmatter.get("name")
+            description = frontmatter.get("description")
+            if not isinstance(name, str) or not isinstance(description, str):
+                raise ValueError("Invalid markdown frontmatter: name and description must be strings")
+            return cls(name=name, description=description, content=raw)
+        raise ValueError(f"Invalid markdown: {raw}")
+
+    @classmethod
+    def dump(cls, name: str, description: str, content: str) -> str:
+        return f"---\nname: {name}\ndescription: {description}\n---\n{content}"
```
Magssch, line 47: Not quite sure if it's possible in practice, but I think this potentially could become malformated frontmatter if `\n` is added to the description. We could consider using yaml.dump() here to avoid it, something like

```@classmethod
def dump(cls, name: str, description: str, content: str) -> str:
    frontmatter = yaml.dump({"name": name, "description": description}, default_flow_style=False).strip()
    return f"---\n{frontmatter}\n---\n{content}"
 ```

Magssch (approved):

PR #3054, opened 2026-06-08
Fix "cdf data download files" overwrites and CogniteFile/document filter issues

Magssch:
# Description

Fixes several issues in interactive file download (`cdf data download files --include-file-contents`):

- Files with the same **Name** but different IDs overwrote each other on disk. Downloaded files are now prefixed with a unique identifier (external ID / internal ID for classic files, `space_externalId` for CogniteFiles).
- A typo in the document filter checkbox (`validator` → `validate`) let empty selections through and caused a 400 from the documents API.
- The "Is CogniteFile" menu option only flipped an internal routing flag instead of actually filtering in the documents API; selecting it could leave classic files in the selection and crash the download with `NotImplementedError`. It now applies a real `exists(instanceId.space)` filter. Additionally, it reverts with a warning if it matches zero documents, and the menu no longer offers "Finished" or to do further filtering when on an empty result set.

## Bump

- [x] Patch
- [ ] Skip

### Fixed

- Fixes a bug in `cdf data download files` where files with the same name would overwrite each other when downloaded.
- Fixes a bug in `cdf data download files` that would lead to a crash when confirming a document filter without selecting any values.
- Fixes an issue with the "Is CogniteFile" option in `cdf data download files` where the selected files would not actually be CogniteFiles but classic files instead.

### DIFF
--- a/cognite_toolkit/_cdf_tk/dataio/_file_contentv2.py
+++ b/cognite_toolkit/_cdf_tk/dataio/_file_contentv2.py
@@ -56,6 +56,14 @@ MULTI_FILE_PART_MAX_SIZE_BYTES = 4_000 * 1024 * 1024  # Each part in a multi-par
 MULTI_FILE_MAX_PART_COUNT = 250  # Maximum number of parts
 
 
+def create_download_filepath(file_directory: Path, name: str, mime_type: str | None, external_id: str) -> Path:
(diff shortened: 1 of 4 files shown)


ronpal (dismissed):
Nice!

ronpal (approved):

PR #3056, opened 2026-06-11
[CDF-28094] Improve cdf data download instances CLI interface for multiple instance spaces

Magssch:
# Description

`cdf data download instances` in non-interactive mode has some unnecessary strict ergonomics. With this change, `--instance-space` / `-s` can be specified multiple times, matching the `--view` parameter which already supported this, and reflecting how multiple instance spaces can be selected in interactive mode. This also enables the command to be run with _only_ `--view` and `--schema-space` defined, which now means all instance spaces will be included in the query. The current user-facing behavior is preserved as this would not work in a CI-environment (it would previously fall back to interactive mode in this case), and the `--instance-space` property itself still supports just providing a single space to the command.

## Bump

- [x] Patch
- [ ] Skip

### Improved

- Allow specifying multiple `--instance-space` values in `cdf data download instances` to download instances from several spaces in one command.

### DIFF
--- a/cognite_toolkit/_cdf_tk/apps/_download_app.py
+++ b/cognite_toolkit/_cdf_tk/apps/_download_app.py
@@ -865,8 +865,8 @@ class DownloadApp(typer.Typer):
                 "--instance-space",
                 "-s",
                 help="Instance space(s) to download instances from. Can be specified multiple times. "
-                "When used without --schema-space and --view, it downloads all instances in the given space(s) without any view properties. "
-                "Required when --schema-space and --view are also provided.",
+                "When used without --schema-space and --view, downloads all instances in the given space(s) without any properties from views. "
+                "When used with --schema-space and --view, limits the download to those spaces.",
             ),
         ] = None,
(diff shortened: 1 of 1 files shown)


cognite_toolkit/_cdf_tk/apps/_download_app.py
```diff
@@ -1009,25 +1020,19 @@ def download_instances_cmd(
...
-                    instance_space=instance_space,
+                InstanceViewSelector(
                     view=SelectedView(
                         space=schema_space,
                         external_id=view_id_str.split("/", maxsplit=1)[0],
                         version=view_id_str.split("/", maxsplit=1)[1] if "/" in view_id_str else None,
                     ),
+                    instance_spaces=selected_instance_spaces,
                     instance_type=instance_type.value,
-                    download_dir_name=instance_space,
```
Magssch, line 1030: Don't use instance space here anymore, since we can be potentially be selecting multiple instance spaces now

ronpal (approved):
Nice!

PR #3058, opened 2026-06-15
[CDF-28109] Fix dml field leaking into GraphQlDmlVersionUpsert API payload

ronpal:
# Description

When a GraphQL schema YAML uses the split-file convention and includes a `dml` key
(pointing to the `.graphql` file), the field was stored as an extra Pydantic field
on `GraphQLDataModelRequest` and forwarded to the `GraphQlDmlVersionUpsert` mutation
variables. The CDF API rejected it as an unknown field.

**Changes:**
- Add `dml: str | None = Field(None, exclude=True)` to `GraphQLDataModelRequest`,
  excluding it from the API payload (mirroring the existing `graphql_file` pattern)
- Fix `_get_graphql_file`, `load_resource_file`, and `get_extra_files` in `GraphQLCRUD`
  to honour the `dml` path when present, completing the feature that was already
  documented in `GraphQLDataModelYAML`
- Add regression tests for both the API payload exclusion and custom dml path resolution

## Bump

- [x] Patch
- [ ] Skip

### Fixed

- GraphQL schema deployment no longer fails with `ResourceUpdateError` when the metadata
  YAML includes a `dml` key referencing the `.graphql` schema file


.pre-commit-config.yaml
```diff
@@ -34,7 +34,7 @@ repos:
   - hooks:
     - id: mypy
       name: mypy
-      entry: uv run dmypy run -- cognite_toolkit/ tests_smoke/ --config-file pyproject.toml
```
ronpal, line 37: This caused a major nuisance due to different behaviour locally and in CI, causing false positives from mypy

Magssch (approved):

PR #3068, opened 2026-06-18
[CDF-28126] Fix infield-data migration dropping solutionTags and producing noisy unmapped-property warnings

Magssch:
# Description

Fixes two issues in `cdf migrate infield-data`: `CogniteSolutionTag` instances used by infield locations were not being migrated over to the space being used by InfieldOnCDM, causing `solutionTags` direct relations on `Checklist`, `Template`, and other views to be silently dropped and existing solution tags (called "labels" in the UI) to not be visible for selection in migrated InfieldOnCDM locations. Additionally, several source properties (`sourceId`, `rootLocation`, `sourceCreatedTime`, `isArchived`, `solutionTags`) in the old `Observation` view that have no destination in the new default `FieldObservation` view were generating spurious unmapped-property warnings in the migration log, so these are now explicitly ignored.

## Bump

- [x] Patch
- [ ] Skip

### Fixed

- Fix `cdf migrate infield-data` dropping `solutionTags` references.
- Silences expected unmapped-property warnings for `sourceId`, `rootLocation`, `sourceCreatedTime`, `isArchived`, and `solutionTags` when running `cdf migrate infield-data`.

### DIFF
--- a/cognite_toolkit/_cdf_tk/commands/_migrate/conversion.py
+++ b/cognite_toolkit/_cdf_tk/commands/_migrate/conversion.py
@@ -27,14 +27,12 @@ from cognite_toolkit._cdf_tk.client.resource_classes.data_modeling import (
     EdgeId,
     EdgeProperty,
     EdgeRequest,
-    EdgeResponse,
     FileCDFExternalIdReference,
     InstanceResponse,
     InstanceSource,
     JSONProperty,
     NodeId,
     NodeRequest,
-    NodeResponse,
     SingleEdgeProperty,
     TimeseriesCDFExternalIdReference,
     ViewCorePropertyResponse,
@@ -773,12 +771,8 @@ class ConnectionCreator:
     def can_map_instance(self, space: str) -> bool:
         return self._instance_id_mapper.can_map(space)
 
-    def map_instance(self, node_id: NodeId | EdgeId | NodeResponse | EdgeResponse) -> NodeId:
+    def map_instance(self, instance_id: NodeId | EdgeId) -> NodeId:
         """Maps a source instance ID to the corresponding destination instance ID."""
-        if isinstance(node_id, NodeId | EdgeId):
-            instance_id = node_id
-        else:
-            instance_id = NodeId(space=node_id.space, external_id=node_id.external_id)
         return self._instance_id_mapper.map_instance_id(instance_id)
 
     def edges(self, view_id: ViewId) -> dict[str, EdgeProperty]:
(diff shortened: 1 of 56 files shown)


cognite_toolkit/_cdf_tk/commands/_migrate/infield_data_mappings.yaml
```diff
@@ -41,6 +53,10 @@
     troubleshooting: troubleshooting
     priority: priority
     type: type
+  ignoreSourceProperties:
+    - sourceId
+    - rootLocation
+    - sourceCreatedTime
```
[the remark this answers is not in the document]
Magssch: node.createdTime is the source property (the node metadata timestamp), and sourceCreatedTime is the destination property in FieldObservation.

The sourceCreatedTime in ignoreSourceProperties refers to a source property named sourceCreatedTime on the cdf_apm/Observation/v5 view itself. Not the same thing

cognite_toolkit/_cdf_tk/commands/_migrate/conversion.py
```diff
@@ -1072,7 +1072,6 @@ class InFieldAssetMapping(CustomConnectionMapping[NodeId]):
             (ViewId(space="cdf_apm", external_id="Checklist", version="v7"), "rootLocation"),
             (ViewId(space="cdf_apm", external_id="ChecklistItem", version="v7"), "asset"),
             (ViewId(space="cdf_apm", external_id="Observation", version="v5"), "asset"),
-            (ViewId(space="cdf_apm", external_id="Observation", version="v5"), "rootLocation"),
```
Magssch, line 1075: No longer passes through here at all since it is skipped

cognite_toolkit/_cdf_tk/commands/_migrate/infield_data_mappings.yaml
```diff
@@ -1,5 +1,18 @@
 # The mapping definitions are topological ordered based on the container constraints of the
 # destination views. This is to ensure that data can be uploaded without violating any constraints.
+- externalId: InFieldCogniteSolutionTagMapping
```
Magssch, line 3: CogniteSolutionTags wasn't previously mapped, they are both used in Canvas and Infield and deliberately not migrated for Canvases since they have some cross-usage. Per the source code for Infield on CDM though it can and will only read solution tags in the appInstanceSpace configured for each location, so "duplicating" the instances like this (into a new space) should be safe as we only do it for those instance spaces that are being used by infield locations.

cognite_toolkit/_cdf_tk/commands/_migrate/infield_data_mappings.yaml
```diff
@@ -41,6 +53,12 @@
     troubleshooting: troubleshooting
     priority: priority
     type: type
+  ignoreSourceProperties:
```
Magssch, line 56: None of these are present anymore on the new default FieldObservation view that comes with InfieldOnCDM

cognite_toolkit/_cdf_tk/commands/_migrate/infield_data_mappings.yaml
Magssch: We are not needing to remap any properties here, since despite it's name, CogniteSolutionTag is not a CDM type nor is it really "new". We bare basically creating a 1-1 copy of the existing instances into the new destination space. The only "new" requirement is for these to be in the same space as all other infield data to be visible and usable in the UI.

ronpal (approved):

PR #3072, opened 2026-06-23
fix: handle list[dict] variable values in BuildVariable hash

gaetan-h:
## Summary

`cdf modules list` (and any command that calls `compare_modules()`) crashes with:

```
TypeError: unhashable type: 'dict'
  File "_module_resources.py", line 163, in compare_modules
    if set(current_module_variables) != set(cached_module.build_variables):
```

when a config variable has a **list-of-dicts** value — for example:

```yaml
variables:
  modules:
    site_readiness:
      - site_name_lower: clov
        site_name_upper: CLOV
        site_prefix: CLV
      - site_name_lower: egina
        site_name_upper: EGINA
        site_prefix: EGI
```

## Root Cause

`BuildVariable` is a `frozen dataclass`, so Python auto-generates `__hash__` from all fields including `value`. Both `load_raw()` and `load()` converted list values with `tuple(value)`, but if the list contains dicts, those dicts end up inside the tuple and are still unhashable — causing `set()` to fail.

## Fix

Extract a `_to_hashable()` helper that **recursively** converts lists and dicts to nested tuples of sorted key-value pairs:

```python
def _to_hashable(v: Any) -> Any:
    if isinstance(v, list):
        return tuple(_to_hashable(i) for i in v)
    if isinstance(v, dict):
        return tuple(sorted((k, _to_hashable(val)) for k, val in v.items()))
    return v
```

Applied in both:
- `load_raw()` — live variables parsed from `config.*.yaml`
- `BuildVariable.load()` — variables read back from `build_info.*.yaml` cache

## Tests

Two new test cases in `test_build_variables.py`:
- `test_load_raw_list_of_dicts_is_hashable` — verifies `load_raw()` with a `list[dict]` variable produces a hashable `BuildVariable`
- `test_load_from_cache_list_of_dicts_is_hashable` — verifies `BuildVariable.load()` (cache read path) does the same

All 10 existing + new tests pass.

### DIFF
--- a/cognite_toolkit/_cdf_tk/commands/modules.py
+++ b/cognite_toolkit/_cdf_tk/commands/modules.py
@@ -785,12 +785,12 @@ class ModulesCommand(ToolkitCommand):
             return
 
         table = Table(title=f"{effective_build_env} {organization_dir.name} modules")
-        table.add_column("Module Name", style="bold")
-        table.add_column("Resource Folders", style="bold")
-        table.add_column("Resources", style="bold")
-        table.add_column("Build Warnings", style="bold")
-        table.add_column("Build Result", style="bold")
-        table.add_column("Location", style="bold")
+        table.add_column("Module Name", style="bold", min_width=20, no_wrap=True)
+        table.add_column("Resource\nFolders", style="bold", justify="right", min_width=8)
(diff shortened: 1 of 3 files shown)


doctrino:
Already resolved

PR #3075, opened 2026-06-24
[CDF-28146] Fix migration upload chunking and tracking ID propagation in FDM-to-CDM mapper

Magssch:
# Description

The migration command's `_upload` path passed entire pages to `upload_items` without splitting
them to each target's `CHUNK_SIZE`, causing a error to be thrown when a page contained more items than
the upload limit, which can happen during Infield migration when edges connected to nodes are being migrated as part of reading nodes through a view. This PR fixes the chunking at the upload layer.

It also fixes a latent bug where `FDMtoCDMMapper` (which fans out one source node into a node + edges) would silently drop or misassign tracking IDs via a `zip` with the shorter source list. The fix changes `DataMapper.map()` to take and return `DataItem`s directly, so each mapper owns its own tracking ID assignment rather than having them re-attached externally by the caller. For the fan-out mapper (currently only FDMtoCDMMapper), each output item can now get a tracking ID appropriately even if the destination list is larger than the source; for 1:1 mappers, the source tracking ID is propagated as before.

## Bump

- [x] Patch
- [ ] Skip

### Fixed

- [alpha] Fixes an issue where `cdf migrate infield-data` uploads could fail with a `ValueError` when migrating ChecklistItems

### DIFF
--- a/cognite_toolkit/_cdf_tk/commands/_migrate/command.py
+++ b/cognite_toolkit/_cdf_tk/commands/_migrate/command.py
@@ -293,7 +293,7 @@ class MigrationCommand(ToolkitCommand):
         mapper: DataMapper[T_Selector, T_DataResponse, T_DataRequest],
     ) -> Callable[[Page[T_DataResponse]], Page[T_DataRequest]]:
         def track_mapping(source: Page[T_DataResponse]) -> Page[T_DataRequest]:
-            return mapper.map_page(source)
+            return source.create_from(mapper.map(source.items))
 
         return track_mapping
 
--- a/cognite_toolkit/_cdf_tk/commands/_migrate/command.py
+++ b/cognite_toolkit/_cdf_tk/commands/_migrate/command.py
@@ -177,6 +177,10 @@ class MigrationCommand(ToolkitCommand):
 
             action = "Would migrate" if dry_run else "Migrating"
             target = "records" if isinstance(data, RecordsMigrationIO) else "instances"
+            # Here we use logger totals instead of the actual number of downladed items. For some selectors,
+            # download pages can include auxiliary edges that are, for example, converted to direct relations
+            # on a node after being migrated, alongside the nodes to migrate themselves. It would be confusing
+            # to a user if those edges were counted and displayed, since the initial estimate only counts the nodes to migrate.
             total = sum(result.count for result in items_results)
             console.print(f"{action} {total:,} {selected.display_name} to {target}.")
         return results_by_selector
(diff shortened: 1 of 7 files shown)


cognite_toolkit/_cdf_tk/commands/_migrate/data_mapper.py
```diff
@@ -1354,27 +1368,26 @@ def prepare(self, source_selector: InstanceSelector) -> None:
         views = self.client.tool.views.retrieve(list(view_ids))
         self._connection_creator.update_view_cache(views)
 
-    def map(self, source: Sequence[NodeOrEdgeResponse]) -> Sequence[NodeOrEdgeRequest | None]:
+    def map(self, source: Sequence[DataItem[NodeOrEdgeResponse]]) -> Sequence[DataItem[NodeOrEdgeRequest]]:
+        raw_items = [data_item.item for data_item in source]
```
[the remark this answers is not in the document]
Magssch: Created this in parallell so this was forgotten, will update!

cognite_toolkit/_cdf_tk/commands/_migrate/data_mapper.py
```diff
@@ -1399,8 +1412,11 @@ def map(self, source: Sequence[NodeOrEdgeResponse]) -> Sequence[NodeOrEdgeReques
                 for source_view_id in (node.properties or {}).keys()
                 if isinstance(source_view_id, ViewId)
             )
+            output.append(DataItem(tracking_id=str(node.as_id()), item=mapped_node))
             mapped_instances.append(mapped_node)
-            mapped_instances.extend(edges)
+            for edge in edges:
+                output.append(DataItem(tracking_id=str(node.as_id()), item=edge))
```
Magssch, line 1418: This might looks strange that I'm registering the edge using the nodes tracking id, but I do this because whenever the API returns an error it can't find back to the originating edge, which would produce _another_ error "Missing registration". This is because we skip registering edges in `emit_registered_page`, because otherwise they will get included in the statistics. The reasoning here is that it can look concerning to a user if there are more resultingly migrated instances compared to the number of nodes (to be migrated) that were initially read through the views. So instead the edges are also registered by the node. That should be fine since edges typically just look like relations on a view for a node anyways

cognite_toolkit/_cdf_tk/commands/_migrate/command.py
```diff
@@ -177,6 +177,7 @@ def _run_migration_steps(
 
             action = "Would migrate" if dry_run else "Migrating"
             target = "records" if isinstance(data, RecordsMigrationIO) else "instances"
+            total = sum(result.count for result in items_results)
```
Magssch, line 184: We can end up downloading many more edges as part of a node download, so this is to avoid the numbers looking weird (we kind of "hide" the fact that edges are part of a download)

ronpal: I would suggest adding this to a comment in the code since it is easy to get confused by :)

Magssch: Good call! Will do :)

ronpal (approved):

PR #3088, opened 2026-07-01
[CDF-28123] fix(graphql): surface real API error when upsertGraphQlDmlVersion returns null

ronpal:
# Description

When the CDF DML API rejects a GraphQL mutation (e.g. unknown fields, schema
validation failures, auth issues) it returns:

```json
{"data": {"upsertGraphQlDmlVersion": null}, "errors": [{"message": "...real reason..."}]}
```

The previous client called `GraphQLResponse.model_validate_json()` first — Pydantic
crashed on the `null` before the code ever reached the `errors` check. The customer saw:

```
Input should be an object (type=model_type, loc=["data","upsertGraphQlDmlVersion"])
```

instead of the actual API rejection reason, making the failure impossible to self-diagnose.

**Changes:**
- Check top-level GraphQL `errors` before Pydantic validation so the real message is always surfaced
- `UpsertResponseData.errors` corrected from `dict` → `list[DMLError]` (matches the API contract)
- `UpsertResponseData.result` made nullable (API returns `null` when DML validation fails)
- `GraphQLUpsertResponse.upsert_graph_ql_dml_version` made nullable (correct per API contract)
- Nested DML validation errors (kind/message/hint) are now surfaced as well
- Two regression tests added

## Bump

- [x] Patch
- [ ] Skip

### Fixed

- `cdf deploy` for GraphQL data models: the real API rejection reason is now surfaced
  instead of an opaque Pydantic `"Input should be an object"` error when
  `upsertGraphQlDmlVersion` returns null.

### DIFF
--- a/cognite_toolkit/_cdf_tk/apps/_core_app.py
+++ b/cognite_toolkit/_cdf_tk/apps/_core_app.py
@@ -89,8 +89,8 @@ class CoreApp(typer.Typer):
         ] = False,
     ) -> None:
         """
-        Docs: https://docs.cognite.com/cdf/deploy/cdf_toolkit/\n
-        Template reference documentation: https://developer.cognite.com/sdks/toolkit/references/configs
+        Docs: https://docs.cognite.com/cdf/deploy/cdf_toolkit/guides/usage\n
+        Resource reference: https://docs.cognite.com/cdf/deploy/cdf_toolkit/references/resource_library
         """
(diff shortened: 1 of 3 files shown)


ronpal:
Superseded by #3113, which includes all changes from this PR plus the root-cause fix (extra YAML fields leaking into the mutation payload).

PR #3090, opened 2026-07-02
feat(infield): update the observation validation to accept the new config structure

laylafy:
# Description

This updates validation for `InFieldCDMLocationConfig` to match the new observation view configuration format and tightens view reference validation across the config.

`viewMappings.observation` now expects a list of observation view configs (still limited to one item for now). Each entry can include a `view` reference and an optional `writeBack` section for notification and attachment endpoints. The toolkit also checks that the referenced observation view exists before deploy.

View references in `viewMappings` and `dataExplorationConfig` now use the same `ViewReference` validation used elsewhere in the toolkit, so invalid view IDs and versions are caught earlier.

closes: https://cognitedata.atlassian.net/browse/FO-3804

## Bump

- [x] Patch
- [ ] Skip

### Added

- `InFieldCDMLocationConfig`: `viewMappings.observation` entries now support an optional `writeBack` section for configuring the notifications and attachments endpoints used when writing observation data back to CDF.
- `InFieldCDMLocationConfig`: Toolkit now verifies that the view referenced under `viewMappings.observation` exists in the project before deploying.

### Changed

- `InFieldCDMLocationConfig`: `viewMappings.observation` entries must now be specified with the view reference nested under a `view` key (e.g. `observation: [{view: {space: ..., externalId: ..., version: ...}}]`) instead of specifying the view fields directly on the list item. Existing configs using the old flat format will need to be updated.
- `InFieldCDMLocationConfig`: view references under `viewMappings` and `dataExplorationConfig` are now validated more strictly (consistent with view references elsewhere in the toolkit), so invalid `space`, `externalId`, or `version` values are now caught at build time instead of failing later at deploy.

### DIFF
--- a/cognite_toolkit/_cdf_tk/resource_ios/_resource_ios/fieldops.py
+++ b/cognite_toolkit/_cdf_tk/resource_ios/_resource_ios/fieldops.py
@@ -403,40 +403,61 @@ class InFieldCDMLocationConfigIO(ResourceIO[NodeId, InFieldCDMLocationConfigRequ
 
     @classmethod
     def get_dependencies(cls, resource: InFieldCDMLocationConfigYAML) -> Iterable[tuple[type[ResourceIO], Identifier]]:
-        if resource.data_exploration_config is None:
-            return
-        for card_attr in INFIELD_CDM_CARD_VIEW_ATTR_TO_JSON_KEY:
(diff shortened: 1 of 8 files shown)


Magssch (approved):
LGTM! I slightly tweaked the changelog a bit to be more end user-facing :)

PR #3095, opened 2026-07-07
[CDF-28203] Improvements to migration logging

Magssch:
# Description

Fixes FDM-CDM migration direct-relation conversion errors being swallowed and replaced by generic messages (e.g. "expected exactly 1, got 0") when the real cause was already known, such as a missing migrated files/timeseries/asset reference. 

Additionally, add an early-abort circuit breaker to the migrate upload step — mirroring the existing `upload` command — such that a systemic write failure stops the run after the first all-failed chunk instead of repeating the expensive split-and-retry cycle for every subsequent page. This reflects user feedback that it's easy for some migrations like `cdf migrate infield-data` to run for long periods of time and then subsequently discover that a specific exception has occurred that makes it pointless to continue (e.g. like having CogniteEquipment populated instead of CogniteAsset).

## Bump

- [x] Patch
- [ ] Skip

### Fixed

- [alpha] When running `cdf migrate infield-data` the migration logs will not show more specific failure reasons when there are errors with direct relations.
- [alpha] When running `cdf migrate` commands, the migration will now stop early in case of repeated critical errors.

### DIFF
--- a/cognite_toolkit/_cdf_tk/commands/_migrate/conversion.py
+++ b/cognite_toolkit/_cdf_tk/commands/_migrate/conversion.py
@@ -851,16 +851,20 @@ class ConnectionCreator:
 
     def create_direct_relation(
         self, value: Any, dm_prop: DirectNodeRelation, source_prop_id: str, source_view_id: ViewId
-    ) -> tuple[NodeId | list[NodeId], list[str]]:
+    ) -> tuple[NodeId | list[NodeId] | None, list[str]]:
         targets, target_issues = self._create_targets(value, source_prop_id, source_view_id)
         relations, relation_issues = self._targets_to_direct_relation(
-            targets, dm_prop, f"{source_view_id!s}.{source_prop_id!s}"
+            targets, dm_prop, f"{source_view_id!s}.{source_prop_id!s}", known_issues=target_issues
         )
         return relations, target_issues + relation_issues
 
     def _targets_to_direct_relation(
-        self, targets: list[NodeId], dm_prop: DirectNodeRelation, source_display_name: str
-    ) -> tuple[NodeId | list[NodeId], list[str]]:
+        self,
+        targets: list[NodeId],
+        dm_prop: DirectNodeRelation,
+        source_display_name: str,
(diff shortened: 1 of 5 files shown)


tests/test_unit/test_cdf_tk/test_commands/test_migration_cmd/test_command.py
```diff
@@ -262,6 +263,22 @@ def upload_items(self, data_chunk: Page, http_client: object, selector: object |
         return []
 
 
+class _FailingUploadIO:
+    CHUNK_SIZE = 1000
+    KIND = "Instances"
```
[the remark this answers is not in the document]
Magssch: Irrelevant imo since it's a test

cognite_toolkit/_cdf_tk/commands/_migrate/conversion.py
```diff
@@ -748,14 +748,13 @@ def _create_targets(
...
-                issues.append(
-                    f"Failed to create direct relation for property {source_prop_id!r} with value {item!r}: {error}"
-                )
-            except KeyError:
-                issues.append(
-                    f"Failed to create direct relation for property {source_prop_id!r} with value {item!r}: "
-                    "no migrated instance found for reference"
-                )
+                issues.append(f"Failed to create direct relation for property {source_prop_id!r}: {error}")
+            except KeyError as error:
```
Magssch, line 752: KeyError should now only occur in case of a logic bug

cognite_toolkit/_cdf_tk/commands/_migrate/conversion.py
```diff
@@ -1206,7 +1217,8 @@ def convert_container_properties(
                 )
                 continue
             errors.extend(issues)
-            created_properties[dest_prop_id] = created_connection
+            if created_connection is not None:
```
Magssch, line 1220: Target node can now be returned as empty from `_targets_to_direct_relation` (which was done to avoid obscuring errors by throwing a new and generric ValueError)

doctrino (approved):

PR #3097, opened 2026-07-09
[CDF-28127] Add SAP writeback support when migrating InField Observations to custom CDM views

Magssch:
# Description

Adds support for migrating InField observations containing SAP writeback state over to custom CDM observation views, per the accepted ADR in the fusion repo: [SAP writeback on CDM custom observation views](https://github.com/cognitedata/fusion/blob/master/docs/decisions/infield/2026-06-19-sap-writeback-on-cdm-custom-observation-views.md).

- `resolve_observation_view_id()` scans deployed InField CDM location configs for the target space to auto-detect a custom observation view, raising an error if there are multiple location configs that sharing the same `appInstanceSpace` yet for some reason have a different Observation view configured. Such a conflict is unlikely to occur, and if so the user will need to run a fully manual migration.
- `InFieldObservationSapStatusMapping` splits the APM model's conflated `status` field into the new business logic `status` field and the predetermined and Infield-enforced SAP-specific `sapStatus`/`notificationIdInSap` fields. This is only done when the destination view supports those two properties.
- Wired into `cdf migrate infield-data`: the destination view for Observations is now always swapped to the resolved custom view (unless `--skip-observations` is set, in which case the observations won't be migrated at all), and the new mapping is added to `custom_properties_mappings`. 
- A practical side-effect of these changes is that we now also do a "best effort" attempt to map legacy observations to custom observations view. This will however only work if the user in practice has made a very basic extension / modification of FieldObservation, if any property rename or more advanced mapping logic is done, the solution will still be to use `--skip-observations` and make a custom script. The method to enable best-effort custom observation migration is to have a custom view defined in the Infield location.

## Bump

- [x] Patch
- [ ] Skip

### Added

- [alpha] When running `cdf migrate infield-data`, toolkit now supports migrating Observations with SAP writeback state onto a custom CDM observation view

### DIFF
--- a/cognite_toolkit/_cdf_tk/apps/_migrate_app.py
+++ b/cognite_toolkit/_cdf_tk/apps/_migrate_app.py
@@ -21,6 +21,7 @@ from cognite_toolkit._cdf_tk.commands._migrate.conversion import (
     ConnectionCreator,
     InFieldAssetMapping,
     InFieldConditionMapping,
+    InFieldObservationSapStatusMapping,
     InFieldUserMapping,
     SpaceMappingInstanceIdMapper,
     SuffixInstanceIdMapper,
@@ -52,6 +53,7 @@ from cognite_toolkit._cdf_tk.commands._migrate.infield_data_mappings import (
     DIRECT_RELATION_EDGE_TIEBREAKERS,
     create_infield_data_mappings,
     create_infield_schedule_selector,
+    resolve_observation_view_id,
 )
 from cognite_toolkit._cdf_tk.commands._migrate.migration_io import (
     AnnotationMigrationIO,
@@ -1578,8 +1580,10 @@ class MigrateApp(typer.Typer):
             bool,
             typer.Option(
                 "--skip-observations",
-                help="Skip migrating Observation data to the default cdf_infield/FieldObservation view. "
-                "Only use this when you intend to migrate observations to a custom observation view manually.",
+                help="Skip migrating Observation data. If not specified, toolkit will automatically migrate data "
+                "to the default cdf_infield/FieldObservation view. If a custom Observation view is configured, "
+                "toolkit will attempt a best-effort migration to this custom view. Only use this flag "
+                "when you intend to migrate observations manually yourself, for example if you need custom logic.",
             ),
         ] = False,
     ) -> None:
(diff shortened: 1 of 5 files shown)


cognite_toolkit/_cdf_tk/apps/_migrate_app.py
```diff
@@ -1672,6 +1676,17 @@ def infield_data(
             # If this skip is not done, users will end up with observations both in the custom observation view and the default FieldObservation view,
             # which can lead to the wrong view being rendered for migrated observations in Infield since it relies on the instances/inspect endpoint.
             infield_mappings = [m for m in infield_mappings if m.destination_view.external_id != "FieldObservation"]
+        else:
+            # If a custom observation view is configured for the target space (e.g. to support SAP writeback),
+            # migrate Observations onto it instead of the default FieldObservation view.
+            custom_observation_view = resolve_observation_view_id(infield_cdm_configs, target_space)
```
Magssch, line 1682: This is a bit of a hacky solution, but it ensures there's only one single place to configure observation views (e.g. on the infield location itself) instead of having to separately provide it as part of the `infield-data` command. If a custom view is defined, the user has previously not been supposed to use this command without `--skip-observations` so this is just enabling wider usage of the command than before. The behavior should remain unchanged for those on the default FieldObservation view.

cognite_toolkit/_cdf_tk/apps/_migrate_app.py
```diff
@@ -1578,8 +1580,10 @@ def infield_data(
             bool,
             typer.Option(
                 "--skip-observations",
-                help="Skip migrating Observation data to the default cdf_infield/FieldObservation view. "
-                "Only use this when you intend to migrate observations to a custom observation view manually.",
+                help="Skip migrating Observation data. If not specified, toolkit will automatically migrate data "
+                "to the default cdf_infield/FieldObservation view. If a custom Observation view is configured, "
+                "toolkit will attempt a best-effort migration to this custom view. Only use this flag "
+                "when you intend to migrate observations using custom logic.",
```
doctrino:
```suggestion
                help="Skip migrating Observation data. If not specified, Toolkit will automatically migrate data "
                "to the default cdf_infield/FieldObservation view. If a custom Observation view is configured, "
                "Toolkit will attempt a best-effort migration to this custom view. Only use this flag "
                "when you intend to migrate observations using custom logic.",
```

doctrino: Further questions.

1. What is best-effort? Can you describe it in a sentence?
2. `Only use when you intend to migrate observations using custom logic.`. Not sure I understand this, the flag here is to NOT migrate observations, by default they will be migrated.

tests/test_unit/test_cdf_tk/test_commands/test_migration_cmd/test_data_mapper.py
```diff
@@ -1070,6 +1073,167 @@ def test_dry_run_treats_mapped_nodes_as_existing(self, dry_run: bool, expected_l
...
+                },
+                False,
+                id="file_not_sent_with_writeback_view",
+            ),
+            pytest.param(
+                "File not sent", False, {"status": "completed"}, True, id="file_not_sent_without_writeback_view"
+            ),
+        ],
+    )
+    def test_infield_observation_sap_status_migrates_to_custom_observation_view(
```
doctrino: I think this tests is fine., but I think I would have a tests only for the logic of `InFieldObservationSapStatusMapping`, i.e., testing the logic of `convert` directly. Currently, this is a large tests with a lot of set up to maintain.

doctrino (dismissed):

cognite_toolkit/_cdf_tk/apps/_migrate_app.py
Magssch: Agreed, this text is confusing, I will make some tweaks

Magssch: I modified the text to not address the "best-effort" aspect, I will document that thoroughly instead

tests/test_unit/test_cdf_tk/test_commands/test_migration_cmd/test_data_mapper.py
Magssch: I agree actually, I did have some of those but ended up deleting as I needed to run a more comprehensive test which surfaced a bug. I'll restore those and delete this one since it's less maintenance

doctrino (approved):

PR #3101, opened 2026-07-10
[CDF-28256] Use /sync endpoint for downloading instances and migrating Infield data

Magssch:
# Description

Switch `cdf download instances` and `cdf migrate infield-data` to use the `/models/instances/sync` endpoint, now with `twoPhase` mode and with a matching _backfill sort_, instead of `/query`. This gives a longer-lived, more robust resumable cursors and allows tailing, which in turn should mitigate a few issues we've seen recently and asks from users. 

This follows doing hands-on testing in multiple projects where queries currently fail with 408s outright or runs slowly with the current setup using /query for both downloads and infield migrations.
Additionally, it tunes the default sort to _skip_ the `(space, externalId)` sort when a query is pinned to only a single space for better performance, since this was discovered after some testing to, on average, provide a better compromise on performance across multiple types of projects.

This PR also fixes a bug in `InstancesAPI.query()` where a merged multi-batch response kept the first batch's cursor instead of the last one, which caused callers to re-fetch and duplicate already-returned items and gave a "duplicate registrations" error. 

This PR aims to (partially) address / mitigate **CDF-28256**, **CDF-28191** and **CDF-28185**.

## Bump

- [x] Patch
- [ ] Skip

### Fixed

- Fix a bug where downloading or migrating instances could produce duplicate items when a page was internally split into multiple batches.

### Improved

- When running `cdf download instances` and `cdf migrate infield-data`, toolkit will now use longer-lived, resumable cursors, making large and/or interrupted downloads/migrations more reliable, easy to recover from, and performant.
- [alpha] When running `cdf migrate infield-data`, it is now possible to rerun a migration without having to re-migrate every instance. You can now continue off from a previous migration run by supplying the `--log-dir` argument and selecting the relevant log folder from the last run. Toolkit will now pick up any added or changed instances since the last migration run given by the `--log-dir` argument, and only migrate those, significantly reducing the processing time from rerunning a migration in case of new data needing to be synced over. PLEASE NOTE: this will currently not propagate _deleted_ instances, meaning that if any users deletes checklists and similar in Infield then this operation will _not_ be reflected in InfieldOnCDM automatically.

### DIFF
--- a/cognite_toolkit/_cdf_tk/client/api/instances.py
+++ b/cognite_toolkit/_cdf_tk/client/api/instances.py
@@ -337,6 +337,11 @@ class InstancesAPI(CDFResourceAPI[InstanceResponse]):
                         first.items[key].extend(items)  # type: ignore[arg-type]
                     else:
                         first.items[key] = items  # type: ignore[assignment]
+            # Advance next_cursor to reflect the last batch. Otherwise callers
+            # that use this merged response as a paginated result will re-issue
+            # the first batch's cursor and re-fetch items 2..N, causing duplicate
+            # registrations downstream.
+            first.next_cursor = results[-1].next_cursor
         return first
 
     @overload
--- a/cognite_toolkit/_cdf_tk/client/api/instances.py
+++ b/cognite_toolkit/_cdf_tk/client/api/instances.py
@@ -165,8 +165,9 @@ class InstancesAPI(CDFResourceAPI[InstanceResponse]):
         # compromise across different projects compared to the (space, externalId) sort.
         filter_spaces = filter.space if filter is not None else None
         space_ext_id_sort: list[QuerySortSpec] | None = None
-        if filter is not None and (filter_spaces is None or len(filter_spaces) > 1):
-            instance_type = filter.instance_type or "node"
+        if filter is None or filter_spaces is None or len(filter_spaces) > 1:
+            instance_type = filter.instance_type if filter is not None else None
+            instance_type = instance_type or "node"
             space_ext_id_sort = [
                 QuerySortSpec(property=[instance_type, "space"], direction="ascending"),
                 QuerySortSpec(property=[instance_type, "externalId"], direction="ascending"),
--- a/cognite_toolkit/_cdf_tk/client/api/instances.py
+++ b/cognite_toolkit/_cdf_tk/client/api/instances.py
@@ -156,8 +156,10 @@ class InstancesAPI(CDFResourceAPI[InstanceResponse]):
     ) -> QueryRequest:
         """Create a query from the instance filter"""
 
-        # Sort to ensure performance. If you do not sort, you get the internal index,
-        # which includes all deleted instances as well.
+        # We sort by space and externalId to get a stable sort order.
+        # This is also more performant than sorting by using the default sort, which will sort on
+        # internal CDF IDs. This will be slow if you have deleted a lot of instances, as they will be counted.
+        # By sorting on space and externalId, we avoid this issue.
         # Exception: when pinned to exactly one space, omit the sort so the server uses
         # its internal-node-id order. This was observed to be provide a better performance
         # compromise across different projects compared to the (space, externalId) sort.
--- a/cognite_toolkit/_cdf_tk/client/api/instances.py
+++ b/cognite_toolkit/_cdf_tk/client/api/instances.py
@@ -28,6 +28,7 @@ from cognite_toolkit._cdf_tk.client.resource_classes.data_modeling import (
 )
 from cognite_toolkit._cdf_tk.client.resource_classes.data_modeling._instance import InstanceSlimDefinition
 from cognite_toolkit._cdf_tk.client.resource_classes.data_modeling._query import (
+    QueryDebugParameters,
     QueryEdgeExpression,
     QueryEdgeTableExpression,
     QueryNodeExpression,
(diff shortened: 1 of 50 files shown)


cognite_toolkit/_cdf_tk/client/api/instances.py
```diff
@@ -143,30 +143,54 @@ def paginate(
...
-        # By sorting on space and externalId, we avoid this issue.
+        # Exception: when the filter isn't pinned to exactly one space (multiple
+        # spaces, or none at all), node_id order no longer correlates with the
+        # container's (space, node_id) btree, so the planner may have to scan and
+        # heapsort the whole container to find the top-N. Requesting a
+        # (space, externalId) sort avoids that by streaming the
+        # (project_id, space, externalId) btree in order instead.
+        filter_spaces = filter.space if filter is not None else None
+        space_ext_id_sort: list[QuerySortSpec] | None = None
+        if filter is not None and (filter_spaces is None or len(filter_spaces) > 1):
```
Magssch: This was something I observed to improve performance specifically when a single space is being filtered on in multiple projects. Not sure how consistent the gain is, but it's just disabling a non-standard sort anyways in that specific case, so I would think it's not a big issue (and can be reverted if we see degraded performance). This seems to provide a good compromise in performance.

cognite_toolkit/_cdf_tk/client/resource_classes/data_modeling/_query.py
```diff
@@ -153,6 +165,15 @@ def dump(
                 exclude_unset=True,
                 by_alias=camel_case,
             )
+        # Sync cursors otherwise expire after 3 days (soft-deleted instances are cleaned up after that grace
+        # period). We accept the risk of missing deletes past that window so long-paused migrations/downloads
+        # can still resume from an old cursor instead of being forced to restart.
+        allow_expired_key = (
+            "allowExpiredCursorsAndAcceptMissedDeletes"
```
Magssch: This is to allow infield data to sync over longer periods of time, with the caveat that we do not sync deletes (we wouldn't have done so anyways without adding specific logic for it, so this is a "free" method of extending the syncing windows beyond 3 days currently).

cognite_toolkit/_cdf_tk/client/api/instances.py
```diff
@@ -300,6 +328,11 @@ def query(
                         first.items[key].extend(items)  # type: ignore[arg-type]
                     else:
                         first.items[key] = items  # type: ignore[assignment]
+            # Advance next_cursor to reflect the last batch. Otherwise callers
+            # that use this merged response as a paginated result will re-issue
+            # the first batch's cursor and re-fetch items 2..N, causing duplicate
+            # registrations downstream.
+            first.next_cursor = results[-1].next_cursor
```
Magssch, line 327: This should fix the "missing registration" errors during 408s as reported here: https://cognitedata.slack.com/archives/C05RW13C6TX/p1782903125146979?thread_ts=1782830091.542489&cid=C05RW13C6TX

cognite_toolkit/_cdf_tk/commands/_migrate/command.py
```diff
@@ -220,7 +220,7 @@ def _create_plan(
                         f"Found progress file for {selector.display_name}. But total items "
                         f"does not match the expected total. Starting from beginning..."
                     )
-                elif progress.status == "completed" and (not is_sync or completed_count == total_items):
```
Magssch, line 223: Removing this allows deltas only to be consumed when using sync after a full migration has been run, making it possible to run incremental migrations (given that the logs with the Progress files have not been deleted)

cognite_toolkit/_cdf_tk/client/api/instances.py
```diff
@@ -143,30 +143,54 @@ def paginate(
...
                     "root": QueryNodeExpression(
                         limit=limit,
                         nodes=QueryNodeTableExpression(),
-                        sort=[
-                            QuerySortSpec(property=["node", "space"]),
-                            QuerySortSpec(property=["node", "externalId"]),
-                        ],
+                        sort=node_sort,
+                        mode=sync_mode,
+                        backfill_sort=node_backfill_sort,
```
Magssch: Using `mode="twoPhase"` and `backFillsort` seems to be the key for performant /sync queries as of today

cognite_toolkit/_cdf_tk/commands/_migrate/command.py
```diff
@@ -186,7 +186,7 @@ def _run_migration_steps(
             else:
                 executor.raise_on_error()
 
-            action = "Would migrate" if dry_run else "Migrating"
+            action = "Would migrate" if dry_run else "Migrated"
```
Magssch, line 189: Somewhat unrelated, but a bit of a confusing typo. This is printed at the end of a migration run for a view, so it makes more sense in past tense.

Magssch: Before:

Image

cognite_toolkit/_cdf_tk/client/api/instances.py
```diff
@@ -143,30 +143,46 @@ def paginate(
...
         """Create a query from the instance filter"""
 
         # We sort by space and externalId to get a stable sort order.
-        #
         # This is also more performant than sorting by using the default sort, which will sort on
         # internal CDF IDs. This will be slow if you have deleted a lot of instances, as they will be counted.
         # By sorting on space and externalId, we avoid this issue.
+        # Exception: when pinned to exactly one space, omit the sort so the server uses
+        # its internal-node-id order. This was observed to be provide a better performance
+        # compromise across different projects compared to the (space, externalId) sort.
```
doctrino, line 165: The amount of ad-hoc knowledge needed to query DMS continuous to increase 😝

doctrino (approved):
I attempted do use sync initially when running InField, but I probably failed to set a few arguments correctly. Anyways that is to say, I think this is the best way to do InField migration. Assume you have tested this on real cases.

PR #3108, opened 2026-07-16
[CDF-28267] Fix 360-image migration pagination issue and dry-run/dangling-model issues

Magssch:
# Description

In `cdf migrate 360-images`, the root ("image360") group of the `InstanceQuerySelector` used for 360ImageStations was only ever meant to drive query-endpoint pagination but lacked a cursor, causing downloads to silently stop after the first page. Once fixed, the root group's property-less stub items were also being flattened into the downloaded dataset alongside the real `image360station` subselection items, causing "Toolkit bug - multiple registrations" warnings when a station identity collided with a root stub. Added an explicit `include_root` flag on `InstanceQuerySelector` so callers can opt the root group out of the downloaded data when it is only there for pagination.

Also extracted 3D model creation out of `Image360CollectionMapper` into a new `Image360CollectionInstanceIO` class (replacing the plain `InstanceIO` for the 360-image-nodes migration step). The mapper now emits revision nodes without a `model3D` reference; `Image360CollectionInstanceIO.upload_items` first uploads the node, and only if that succeeds does it create the underlying 3D model and re-upload the node with the `model3D` reference patched in. This ordering ensures a failed node upsert can no longer leave behind a dangling, orphaned 3D model.

This by effect fixed a latent bug in `Image360CollectionMapper` not honoring `--dry-run`: it never received the `dry_run` flag from the outer mapper because these typically do not perform writes, however with this mapper specifically it has a side-effect of creating a 3D model, meaning a dry run could still perform real 3D model/revision-node writes. 

Finally, `Image360FDMtoCDMMapper` now overrides `map()` to collect all cubemap face file external IDs from the page and batch-warm the `migration.lookup.files` cache before per-node mapping begins, following the same `_populate_cache` pattern used by `ThreeDMapper`, `ThreeDAssetMapper`, and `ChartMapper`. The old approach was way to slow to run at scale.

## Bump

- [x] Patch
- [ ] Skip

### Fixed

- [alpha] Fix `cdf migrate 360-images` failing to migrate more than 1000 360ImageStations
- [alpha] Fixes an issue with `cdf migrate 360-images --dry-run` performing actual 3D model writes instead of simulating them.

### Improved

- [alpha] Improve download speed for `cdf migrate 360-images`

### DIFF
--- a/cognite_toolkit/_cdf_tk/commands/_migrate/data_mapper.py
+++ b/cognite_toolkit/_cdf_tk/commands/_migrate/data_mapper.py
@@ -2155,7 +2155,8 @@ class Image360CollectionMapper(DataMapper[InstanceSelector, NodeOrEdgeResponse, 
             instance_space = node.space
             collection_ext_id = sanitize_instance_external_id(node.external_id, "_cdm")
             migrated_id = NodeId(space=instance_space, external_id=collection_ext_id)
-            name = ((node.properties or {}).get(LEGACY_IMAGE360_COLLECTION_SOURCE_VIEW) or {}).get("label")
+            label_value = ((node.properties or {}).get(LEGACY_IMAGE360_COLLECTION_SOURCE_VIEW) or {}).get("label")
+            name = label_value if isinstance(label_value, str) else None
 
             model_external_id: str | None = None
             migrated_node = existing_migrated_by_id.get(migrated_id)
@@ -2186,7 +2187,9 @@ class Image360CollectionMapper(DataMapper[InstanceSelector, NodeOrEdgeResponse, 
         return results
 
     @staticmethod
-    def _revision_node_request(space: str, external_id: str, name: Any, model_external_id: str | None) -> NodeRequest:
+    def _revision_node_request(
+        space: str, external_id: str, name: str | None, model_external_id: str | None
+    ) -> NodeRequest:
         revision_properties: dict[str, Any] = {
             "status": "Done",
             "published": True,
--- a/cognite_toolkit/_cdf_tk/commands/_migrate/data_mapper.py
+++ b/cognite_toolkit/_cdf_tk/commands/_migrate/data_mapper.py
@@ -10,6 +10,7 @@ from cognite.client.exceptions import CogniteException
 from pydantic import JsonValue
 
 from cognite_toolkit._cdf_tk.client import ToolkitClient
+from cognite_toolkit._cdf_tk.client.http_client import ToolkitAPIError
 from cognite_toolkit._cdf_tk.client.identifiers import ContainerId, EdgeTypeId, ExternalId, InstanceId, InternalId
 from cognite_toolkit._cdf_tk.client.resource_classes.annotation import (
     AnnotationResponse,
@@ -2130,13 +2131,22 @@ class Image360FDMtoCDMMapper(FDMtoCDMMapper):
 class Image360CollectionMapper(DataMapper[InstanceSelector, NodeOrEdgeResponse, NodeOrEdgeRequest]):
     """Maps each legacy Image360Collection node to a Cognite360ImageCollection CDM node.
 
-    First, as a side-effect, registers an Image360 3D model in the 3D backend per collection
-    (or reuses an existing model3D reference when the collection was already migrated previously)
-    and sets model3D on the collection revision.
+    As a side-effect, registers an Image360 3D model in the 3D backend per collection (or reuses
+    an existing model3D reference when the collection was already migrated previously) and sets
+    model3D on the collection revision.
+
+    To avoid ever leaving a dangling/orphaned 3D model behind, the revision node is first
+    created/upserted *without* a model3D reference. Only once that write has been confirmed to
+    succeed do we create the 3D model; the mapped node returned from `map` (which includes the
+    model3D reference) is then written by the regular upload step, effectively patching the
+    revision node with the reference to the newly created model.
 
     Registered via FDMtoCDMMapper.custom_instance_mappings so it runs instead of the default
     ViewToViewMapping path for Image360Collection source nodes.
     """
+
+    SOURCE_LABEL: ClassVar[str] = "Image360Collection"
+    DESTINATION_LABEL: ClassVar[str] = "Cognite360ImageCollection"
 
     def map(self, source: Sequence[DataItem[NodeOrEdgeResponse]]) -> Sequence[DataItem[NodeOrEdgeRequest]]:
         raw_items = [data_item.item for data_item in source]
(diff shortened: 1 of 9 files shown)


cognite_toolkit/_cdf_tk/commands/_migrate/data_mapper.py
```diff
@@ -2156,10 +2168,10 @@ def map(self, source: Sequence[DataItem[NodeOrEdgeResponse]]) -> Sequence[DataIt
                 if self.dry_run:
                     model_external_id = "cog_3d_model_<dry-run>"
                 else:
-                    label = str(
-                        ((node.properties or {}).get(LEGACY_IMAGE360_COLLECTION_SOURCE_VIEW) or {}).get("label")
-                        or node.external_id
+                    self.client.tool.instances.create(
```
Magssch: First, create the node without model3D DR set

cognite_toolkit/_cdf_tk/commands/_migrate/data_mapper.py
```diff
@@ -2168,33 +2180,37 @@ def map(self, source: Sequence[DataItem[NodeOrEdgeResponse]]) -> Sequence[DataIt
...
-                                source=ContainerId(space="cdf_cdm", external_id="CogniteDescribable"),
-                                properties={
-                                    "name": (
-                                        (node.properties or {}).get(LEGACY_IMAGE360_COLLECTION_SOURCE_VIEW) or {}
-                                    ).get("label")
-                                },
-                            ),
-                        ],
-                    ),
+                    item=self._revision_node_request(instance_space, collection_ext_id, name, model_external_id),
```
Magssch: Then only after creating the initial revision node and the 3D model, create the link to the 3D model

cognite_toolkit/_cdf_tk/commands/_migrate/data_mapper.py
```diff
@@ -2113,14 +2114,24 @@ def _map_single_node(
 class Image360CollectionMapper(DataMapper[InstanceSelector, NodeOrEdgeResponse, NodeOrEdgeRequest]):
     """Maps each legacy Image360Collection node to a Cognite360ImageCollection CDM node.
 
-    First, as a side-effect, registers an Image360 3D model in the 3D backend per collection
-    (or reuses an existing model3D reference when the collection was already migrated previously)
-    and sets model3D on the collection revision.
+    As a side-effect, registers an Image360 3D model in the 3D backend per collection (or reuses
```
doctrino: What do you mean with register?

cognite_toolkit/_cdf_tk/commands/_migrate/data_mapper.py
```diff
@@ -2168,33 +2181,37 @@ def map(self, source: Sequence[DataItem[NodeOrEdgeResponse]]) -> Sequence[DataIt
...
+            "published": True,
+            "type": "Image360",
+        }
+        if model_external_id is not None:
+            revision_properties["model3D"] = {"space": space, "externalId": model_external_id}
+        return NodeRequest(
+            space=space,
+            external_id=external_id,
+            sources=[
+                InstanceSource(
```
doctrino: (for my understanding) Why are you writing directly to a container and not a view?

cognite_toolkit/_cdf_tk/commands/_migrate/data_mapper.py
```diff
@@ -2156,10 +2169,10 @@ def map(self, source: Sequence[DataItem[NodeOrEdgeResponse]]) -> Sequence[DataIt
                 if self.dry_run:
                     model_external_id = "cog_3d_model_<dry-run>"
                 else:
-                    label = str(
-                        ((node.properties or {}).get(LEGACY_IMAGE360_COLLECTION_SOURCE_VIEW) or {}).get("label")
-                        or node.external_id
+                    self.client.tool.instances.create(
```
doctrino: It is confusing to me that you are creating nodes in a mapper. In my mind, the mapper should only map, i.e., create migrated nodes from existing nodes. I would expect this to be done in the upload portion of the migration. Making migration = download -> map -> upload. And skip upload if dry run is set.

doctrino (commented):
I am confused by the mapper that also do writing. Can you explain that to me?

cognite_toolkit/_cdf_tk/commands/_migrate/data_mapper.py
Magssch: As in making a call to `post /3d/models`, will make this clearer

cognite_toolkit/_cdf_tk/commands/_migrate/data_mapper.py
Magssch: Had to go back a bitand remember why I did this, so I will add a comment on it. The reason for this is that the Cognite360ImageCollection (which implemens Cognite3DRevision) view is broken currently, so I didn't want to write directly through it. There are two version of that view, v1 and v2, and only the v1 is included in the data model, which doesn't show any valid instances in the DM UI due to a broken hasData filter. Looking at it again, for some reason when this view is queried in the UI to show 3D models, this does seems to work anyways though, since they use an Equals filter in the with which probably overrides that hasData, so it might have been fine to write directly through it anyways. 

Cognite3DRevision on its part does not extend CogniteDescribable and has only its own container, but "name" is required from describable as well in order to show name appropriately, so I had to write to both.

cognite_toolkit/_cdf_tk/commands/_migrate/data_mapper.py
Magssch: Hmm that is true. The thinking is I had to create the 3D models themselves in the mapper since the resultingly created externalId is non-deterministic and that ID to be added as a direct relation in the Revision node before that one is uploaded.

Magssch: This makes less sense now when we are making the nodes first anyways so I will try to refactor this a bit and see if we maybe make a new IO class for this action specifically instead

cognite_toolkit/_cdf_tk/commands/_migrate/migration_io.py
```diff
@@ -560,6 +564,119 @@ def data_to_json_chunk(
...
+    collections that do not already have one, and patches the model3D reference onto the revision
+    node once the model has been created.
+
+    To avoid ever leaving a dangling/orphaned 3D model behind, the revision node is first
+    uploaded *without* a model3D reference (as mapped by Image360CollectionMapper). Only once that
+    write has succeeded do we create the 3D model; the revision node is then re-uploaded with the
+    model3D reference patched in.
+    """
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
```
doctrino, line 580: This maybe a personal preference, but I don't like args kwargs. They have a benefit in what they do not need to repeat the parent if the parent is extended, but it makes it hard to use the class directly. Up to you whether you keep it or not.

doctrino (approved):
This was easier to follow and think less error-prone with respect to dry-run behavior.

PR #3109, opened 2026-07-16
feat(infield): support requiredProperties on observation view mapping

DiemasMichiels:
# Description

InField now lets customers mark arbitrary properties (e.g. direct-relation fields like `assets`/`files`) as required on a custom observation view, via `viewMappings.observation[].requiredProperties`. This adds `requiredProperties` (a plain list of property names, not restricted to any fixed set) as a sibling of `view`/`writeBack` on the observation view mapping, and extends the existing `InFieldCDMViewPropertiesRuleSet` (already used for card views) to validate at build time that the referenced observation view actually has each listed property.

Follows on from #3090, which introduced the `view`/`writeBack` structure for `viewMappings.observation`.

## Bump

- [x] Patch
- [ ] Skip

### Added

- When deploying `InFieldCDMLocationConfig` resources: `viewMappings.observation` entries now support an optional `requiredProperties` list. Toolkit will now validate at build time that the referenced observation view has each listed property, surfacing a `INFIELD-CDM-VIEW-MISSING-PROPERTIES` error if not.


Magssch:
LGTM!

Magssch (approved):

PR #3110, opened 2026-07-16
Update InField CDM location config space defaults during migration and emit required spaces

Magssch:
# Description

The understanding of what should be defaults when running `cdf migrate infield-configs` has changed based on recent trial runs:

- Previously the `InFieldCDMLocationConfig` node's own `space` defaulted to the legacy `SourceDataInstanceSpace`, which is wrong per InField's access-control model: this space should be used for source data only. The `space` of the `InFieldCDMLocationConfig` itself is what governs access to Infield locations for users. The node's `space` now uses a dedicated `<appInstanceSpaceBase>_cfg` space so that write access to InField location configs are independently access-controlled.
- When `featureToggles` was not configured in the old location, all toggles now default to `true` instead of being left null, ensuring features are available out of the box after migration.
- `dataFilters` for `timeseries` and `files` default to the asset instance space (same as `assets`/`rootLocation`), with an inline YAML comment instructing users to update `instanceSpaces` if their time series or files live elsewhere. This is the closest we can get to an automatically derived space for these.
- The `assets` data filter now defaults to the same space as the `rootLocation` asset, since divergence between the two can lead to the asset hierarchy not being visible in InField.
- The space derived from `SourceDataInstanceSpace` is now also suffixed with `_cdm`, and a soon-to-follow-up PR will introduce separate tooling to migrate those instances to new spaces as well since the data in the APM_SourceData model(s) currently have not straightforward migration path.
- The command now emits the `_cdm` and `_cdf`-suffixed spaces it depends on, so the generated config is deployable out of the box.

## Bump

- [x] Patch
- [ ] Skip

### Improved

- [alpha] Improved `cdf migrate infield-configs` to generate location configs with better defaults that reflect the access control and data setup required for InField on CDM to work.

### DIFF
--- a/cognite_toolkit/_cdf_tk/commands/_migrate/creators.py
+++ b/cognite_toolkit/_cdf_tk/commands/_migrate/creators.py
@@ -62,6 +62,7 @@ class CreatedResource(Generic[T_RequestResource]):
     resource: T_RequestResource
     config_data: dict[str, Any] | None = None
     filestem: str | None = None
+    field_comments: dict[str, str] | None = None
 
 
 @dataclass
@@ -284,6 +285,13 @@ class SourceSystemCreator(MigrationCreator):
             raise ValueError("This should not happen.")
 
 
+LOCATION_CONFIG_SPACE_COMMENT = (
+    "This space controls who can see this location in InField (dataModelInstancesAcl READ). "
+    "Defaulted to the appInstanceSpace; set a dedicated space here if you need different access "
+    "control for location visibility than for write access."
+)
+
+
 class InfieldV2ConfigCreator(MigrationCreator):
     TARGET_SPACE = "APM_Config"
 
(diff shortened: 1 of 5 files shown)


cognite_toolkit/_cdf_tk/commands/_migrate/creators.py
```diff
@@ -463,15 +513,23 @@ def _create_location_config(
         if config.checklist_admins:
             access_management["checklistAdmins"] = config.checklist_admins  # type: ignore[assignment]
 
-        data_filters: dict[str, JsonValue] = {}
-        for key in ["assets", "files", "timeseries", "maintenanceOrders", "operations", "notifications"]:
-            # Todo Assets does not support multiple instanceSpaces.
-            #    add to Validation in cdf build.
-            data_filters[key] = {"instanceSpaces": [config.source_data_instance_space]}
+        app_instance_space = f"{config.app_data_instance_space}_cdm"
+        source_instance_space = f"{config.source_data_instance_space}_cdm"
```
Magssch, line 545: The source data migration (not yet implemented) migrating data in APM_SourceData->IDM model will also suffix its destination space with "_cdm", so we anticipate that space here rather than the raw (classic) source_data_instance_space. Although we likely would not have needed to actually change the instance space of the migrated nodes, we do it to reduce the risk of data loss in case of accidental deletions propagating from one view to another.

ronpal (approved):

PR #3111, opened 2026-07-16
[CDF-28270] Infield source data migration

Magssch:
# Description

Adds a new `cdf migrate infield-source-data` CLI command that migrates APM source data (Activities, Operations, Notifications) from the `APM_SourceData` FDM space to the IDM views (`CogniteMaintenanceOrder`, `CogniteOperation`, `CogniteNotification`). Reusing as much as possible of the existing logic and behavior from `cdf migrate infield-data` to keep things consistent, but some modifications are necessary since both the source and destination views can differ, as opposed to for the infield system model.

The command resolves source APM view IDs and instance spaces at runtime based on the `APM_Config` node since this will vary for each project, whole destination views are determined by looking up `InfieldCDMLocationConfigs` similar to how custom `Observations` are now handled. Field mappings and unmappable properties are defined in a new YAML config, and the connection creator is extended to support runtime-resolved view IDs for asset and maintenance order relations.

## Bump

- [x] Patch
- [ ] Skip

### Added

- [alpha] Add `migrate infield-source-data` command to migrate APM source data (Activities, Operations, Notifications) to the IDM data model.


cognite_toolkit/_cdf_tk/commands/_migrate/apm_source_data_mappings.yaml
```diff
@@ -0,0 +1,77 @@
...
+    id: sourceId
+    title: name
+    startTime: scheduledStartTime
+    endTime: scheduledEndTime
+    actualStartTime: startTime
+    actualEndTime: endTime
+    assetExternalId: mainAsset
+    node.createdTime: sourceCreatedTime
+    node.lastUpdatedTime: sourceUpdatedTime
+  ignoreSourceProperties:
```
Magssch, line 29: Ignoring these by default since they do not exist for CogniteMaintenaceOrder, though if a user has created a custom view these will still be mapped. Ignore only applies if no mapping can be found

Magssch: Same logic applies to the other types as well

ronpal (dismissed):

ronpal (approved):

PR #3118, opened 2026-07-24
feat(infield): update config to add sorting for infield

laylafy:
# Description

Replaces legacy InField CDM location config fields with structured field mappings and validates them at schema and build time.

**Observation config (`viewMappings.observation[]`):**
- Removes `formView` and `requiredProperties`
- Adds `fieldsConfig` — a map of view property names to per-field settings (`isRequired`, `isEditable`, `orderNumber`)

Asset properties card (`dataExplorationConfig`):
- Removes `assetPropertiesCardView`
- Adds `assetPropertiesCardConfig` — a map of asset view property names to display settings (`displayName`, `orderNumber`)

**Validation:**
- Schema: property key format, forbidden identifiers, `orderNumber >= 0`, unknown nested fields rejected
- Build: `fieldsConfig` keys checked against the observation view; `assetPropertiesCardConfig` keys checked against `viewMappings.asset` (defaults to `CogniteAsset` if asset view is missing)

**Breaking change:** configs using `formView`, `requiredProperties`, or `assetPropertiesCardView` must migrate to the new fields.

## Bump

- [x] Patch
- [ ] Skip

### Removed

- `formView` and `requiredProperties` from `viewMappings.observation`
- `assetPropertiesCardView` from `dataExplorationConfig`

### Added
- `fieldsConfig` on observation view config for per-field `isRequired`, `isEditable`, and `orderNumber`
- `assetPropertiesCardConfig` on `dataExplorationConfig` for per-field `displayName` and `orderNumber`
- Schema validation for field-config property key format and reserved identifiers
- Build-time validation (`INFIELD-CDM-UNKNOWN-VIEW-PROPERTY`, `INFIELD-CDM-MISSING-ASSET-VIEW`) that field-config keys exist on the referenced CDF views

### DIFF
--- a/cognite_toolkit/_cdf_tk/yaml_classes/infield_cdm_location_config.py
+++ b/cognite_toolkit/_cdf_tk/yaml_classes/infield_cdm_location_config.py
@@ -1,10 +1,32 @@
-from pydantic import Field
+import re
+from collections.abc import Mapping
+
+from pydantic import Field, field_validator
 
 from cognite_toolkit._cdf_tk.client.identifiers import NodeId
-from cognite_toolkit._cdf_tk.constants import SPACE_FORMAT_PATTERN
+from cognite_toolkit._cdf_tk.constants import (
+    CONTAINER_AND_VIEW_PROPERTIES_IDENTIFIER_PATTERN,
+    FORBIDDEN_CONTAINER_AND_VIEW_PROPERTIES_IDENTIFIER,
+    SPACE_FORMAT_PATTERN,
+)
+from cognite_toolkit._cdf_tk.utils.collection import humanize_collection
 
 from .base import BaseModelResource, ToolkitResource
 from .view_field_definitions import ViewReference
+
+_PROPERTY_KEY_PATTERN = re.compile(CONTAINER_AND_VIEW_PROPERTIES_IDENTIFIER_PATTERN)
+
+
+def _validate_view_property_keys(val: Mapping[str, object]) -> Mapping[str, object]:
+    for key in val:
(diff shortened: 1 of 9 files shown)


cognite_toolkit/_cdf_tk/yaml_classes/infield_cdm_location_config.py
```diff
@@ -99,10 +145,20 @@ class ViewMappings(BaseModelResource):
...
 
+    @field_validator("asset_properties_card_config")
+    @classmethod
+    def validate_asset_properties_card_config_keys(
+        cls, val: dict[str, AssetPropertiesCardFieldConfig] | None
+    ) -> dict[str, AssetPropertiesCardFieldConfig] | None:
+        if val is None:
+            return val
+        _validate_view_property_keys(val)
+        return val
```
doctrino: This is correct, but might be overkill as it is already validated when the view was deployed. It could be confusing for the user to get complains on an InField ViewMapping.

doctrino (dismissed):
One suggestion. Your decision whether to take it into account or not. Will be fairly quick to re-approve if you do.

cognite_toolkit/_cdf_tk/yaml_classes/infield_cdm_location_config.py
laylafy: that's a good point. I removed this check: [554c558](https://github.com/cognitedata/toolkit/pull/3118/commits/554c5588e28a610d01c0f38e645b53ff81789ad5)

doctrino (approved):

PR #3122, opened 2026-08-03
[CDF-28399, CDF-28157] Add source file to build insights

ronpal:
## Description

Adds a `source_file` to console, as well as `insights.csv` and `insights.json` so users can see which YAML file triggered each build insight. Populates the field for syntax warnings, file read errors, validation rules, and dependency checks. Also updates the create-jira-ticket skill to use the H2 maintenance epic (CDF-28398).

Screenshot 2026-08-03 at 14 27 11

## Bump

- [x] Patch
- [ ] Skip

### Fixed

- Build insights export now includes `source_file` paths so users can locate the specific YAML(s) that triggered each insight.


cognite_toolkit/_cdf_tk/rules/_infield.py
Magssch, line 179: I think this one slipped through, might want to add source_file here as well

Magssch (approved):
LGTM!

PR #3123, opened 2026-08-03
Add automatic debug output for 408s and surface X-Request-ID in errors

Magssch:
# Description

When a sync query fails without any way to recover (which is, after we have attempted to reduce the chunk size all the way down to 1 and it still fails), automatically issue a diagnostic query with plan/profile debug parameters and write the result to disk, then reference that file path in the raised error as to aid the debugging process.

Additionally, we now surface the `X-Request-ID` header on API errors so it's available for support investigations, and deduplicate the log-filestem-generation logic (`_download.py`, `_upload.py`, `_migrate/command.py`) into a shared `create_logfile_stem` util.

## Bump

- [x] Patch
- [ ] Skip

### Added

- When downloading instances, toolkit will now automatically capture a query plan/profile debug file and reference it in the error message when a sync query request repeatedly times out (408).

### Fixed

- Include the `X-Request-ID` in API error messages so it can be shared with support when troubleshooting failed requests.

### DIFF
--- a/cognite_toolkit/_cdf_tk/client/api/instances.py
+++ b/cognite_toolkit/_cdf_tk/client/api/instances.py
@@ -182,7 +182,7 @@ class InstancesAPI(CDFResourceAPI[InstanceResponse]):
                 QuerySortSpec(property=[instance_type, "space"], direction="ascending"),
                 QuerySortSpec(property=[instance_type, "externalId"], direction="ascending"),
             ]
-        sync_mode: Literal["onePhase", "twoPhase", "noBackfill"] | None = "onePhase" if endpoint == "sync" else None
+        sync_mode: Literal["onePhase", "twoPhase", "noBackfill"] | None = "twoPhase" if endpoint == "sync" else None
 
         if filter is None:
             query = QueryRequest(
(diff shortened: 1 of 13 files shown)


cognite_toolkit/_cdf_tk/client/api/instances.py
```diff
@@ -490,6 +524,61 @@ def _make_query(
...
+            retry_status_codes=set(),
+            api_version="alpha",
+            client_timeout=65,
+        )
+        try:
+            response = self._http_client.request_single_retries(request)
+            success_response = response.get_success_or_raise(request)
+            debug_response = QueryResponseUntyped.model_validate_json(success_response.body)
+            if debug_response.debug:
+                return self._write_debug_info(endpoint_name, debug_response.debug, debug_writer.output_dir)
```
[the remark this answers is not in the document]
Magssch: That's completely wrong

cognite_toolkit/_cdf_tk/client/http_client/_client.py
```diff
@@ -206,7 +209,7 @@ def _handle_response_single(self, response: httpx.Response, request: RequestMess
         if retry_request := self._retry_request(response, request, error_details):
             return retry_request
         else:
-            # Permanent failure
+            # Permenant failure
```
ronpal: "Permanent" is correct spelling, no?

ronpal (dismissed):
Nice! Just a small typo introduced ;)

cognite_toolkit/_cdf_tk/client/http_client/_client.py
Magssch: Haha, whoops, fixed now!

ronpal (approved):

PR #3127, opened 2026-08-04
Fix misleading union validation errors and duplicate unresolved-variable warnings during build

Magssch:
# Description

Four small, unrelated bug fixes split out of a larger build-insights overhaul:
- Pydantic reported misleading errors for reverse direct relations due to a missing union discriminator. If you had a RDR using `through` that targets a view property instead of a container property, it would still get parsed as a ContainerReference, and you would get validation errors for the `type="view"` and `version` fields, which are valid for ViewReference but not ContainerReference. This seemed to only apply when Neat DM validation was disabled.
- The same unresolved template variable could be reported multiple times in the same file
- Neat validation could flood the terminal with cryptic errors for data models with unresolved variables, we now skip Neat validation if there are any unresolved variables.

Additionally, since we can't always control the output of Neat validation and similar "external" validators, we now have a hard truncation of 1000 characters in the terminal for these types of unhandled exceptions. The full error is still written to the log file.

## Bump

- [x] Patch
- [ ] Skip

### Fixed

- Fix misleading validation errors when a view/container reference under a reverse direct relation property has an invalid or missing field.
- Fix duplicate reporting of the same unresolved template variable (e.g. `{{space}}`) in build warnings.

### Changed
- Toolkit will now skip Neat data model validation for files with unresolved template variables since these need to be fixed before further validation can be performed.


cognite_toolkit/_cdf_tk/commands/build_v2/build_v2.py
```diff
@@ -904,6 +907,16 @@ def _display_insights(self, insights: InsightList, insight_path: Path, console:
...
 
+    _MAX_TERMINAL_MESSAGE_LENGTH: ClassVar[int] = 1000
+
+    @classmethod
+    def _truncate_for_terminal(cls, message: str) -> str:
+        """Truncates a message for terminal display only; the full message is always written to the insights file."""
+        if len(message) <= cls._MAX_TERMINAL_MESSAGE_LENGTH:
+            return message
+        truncated_notice = "[dim italic](truncated, see full message in the insights file)[/]"
+        return f"{message[: cls._MAX_TERMINAL_MESSAGE_LENGTH]}... {truncated_notice}"
```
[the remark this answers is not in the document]
Magssch: I think this is somewhat of a pre-existing issue so I'd prefer to not address it for now

cognite_toolkit/_cdf_tk/rules/_neat.py
```diff
@@ -47,6 +47,11 @@ def validate(self) -> Iterable[Insight | FailedValidation]:
         for module in self.modules:
             for resource in module.resources:
                 if resource.type == data_model_type:
+                    if resource.source_path in module.unresolved_variables_by_source:
```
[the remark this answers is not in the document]
Magssch: False premise, unresolved_variables_by_source is required to be non-null

cognite_toolkit/_cdf_tk/commands/build_v2/build_v2.py
```diff
@@ -676,11 +676,13 @@ def _read_resource_file(
 
     @classmethod
     def _find_unresolved_variables(cls, content: str) -> list[str]:
-        return [
-            # Removing the '{{' and '}}'
-            variable[2:-2].strip()
-            for variable in re.findall(pattern=r"\{\{.*?\}\}", string=content)
-        ]
+        return list(
```
Magssch, line 679: - Before: 
Screenshot 2026-08-04 at 16 31 08
  - After: 
Screenshot 2026-08-04 at 16 32 10

cognite_toolkit/_cdf_tk/rules/_neat.py
Magssch, line 50: This avoids Neat triggering multiple parsing errors (since it can't handle {{}} variables): Image

cognite_toolkit/_cdf_tk/commands/build_v2/build_v2.py
```diff
@@ -904,6 +907,16 @@ def _display_insights(self, insights: InsightList, insight_path: Path, console:
             )
         )
 
+    _MAX_TERMINAL_MESSAGE_LENGTH: ClassVar[int] = 1000
+
+    @classmethod
+    def _truncate_for_terminal(cls, message: str) -> str:
```
Magssch, line 913: Also avoids flooding if a similar issue like shown below occurs

ronpal (approved):
👌

jakobvaa:
LGTM 🔥

PR #3129, opened 2026-08-05
[CDF-28401] Include source filename in upload tracking IDs and propagate split error messages

Magssch:
# Description

Tracking IDs for uploaded items now include the source filename (e.g. `part-0000.ndjson:line-1`) instead of a generic `row N`/`line N`, so failures can be traced back to the exact file when reading spans multiple files. As an additional benefit, since one single upload log is used for multiple ndjson files (e.g. "part-0000", "part-0001" and so on), this mitigates potential `Toolkit bug - multiple registrations` warnings that could occur if e.g. the item on "line 1" had an error in both files (since both individual errors would previously get the same tracking ID). 

Additionally, when a batch is aborted after exceeding the failed-split limit, the error message now propagates the actual error from the parent request instead of a generic `Aborting further splitting of requests after 50 failed attempts` message. This makes the originating error / the reason for the item being skipped more obvious. 

## Bump

- [x] Patch
- [ ] Skip

### Improved

- Upload failure logs now reference the source filename and line number (e.g. `part-0000.ndjson:line-1`), making it easier to locate the offending item when uploading data from multiple files.
- Error messages for aborted upload batches now reflect the underlying cause instead of the generic `Aborting further splitting of requests after 50 failed attempts` message.

### DIFF
--- a/cognite_toolkit/_cdf_tk/client/http_client/_client.py
+++ b/cognite_toolkit/_cdf_tk/client/http_client/_client.py
@@ -386,7 +386,7 @@ class HTTPClient:
             return [
                 ItemsFailedRequest(
                     ids=[str(item) for item in message.items],
-                    error_message=f"Aborting further splitting of requests after {message.tracker.failed_split_count} failed attempts.",
+                    error_message=message.parent_error_message or "Unknown error",
                 )
             ]
         try:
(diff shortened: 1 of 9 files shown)


ronpal (approved):
Super!

PR #3132, opened 2026-08-06
[CDF-28431] Emit distinct CDM instance spaces when Infield locations share a space.

Magssch:
# Description

When multiple classic Infield root locations share the same `appDataInstanceSpace` / `sourceDataInstanceSpace`, `cdf migrate infield-configs` previously naively migrated them onto the same single `{space}_cdm` / `{space}_cfg` target. THis will however not be sufficient for InfieldOnCDM as the `rootLocation` concept that is typically used for filtering in the Infield UI is getting deprecated. This must be solved during migrations and not later as to avoid complicated workarounds with rekeying instances to new spaces. This PR makes source Infield locations using shared legacy spaces emit InfieldOnCDM config iwth per-location names using the root asset external ID (`{legacy}_{rootAsset}_cdm`), sanitized to the DMS 43-character limit. Unshared locations keep the existing `{legacy}_cdm` / `_cfg` naming.

This is **part 1** of a stacked series and a prerequisite for part 2 which will resolve and fan out instance data at migrate time into these distinct spaces.

## Bump

- [x] Patch
- [ ] Skip

### Changed

- [alpha] `cdf migrate infield-configs` now creates distinct per-location CDM instance spaces when multiple root locations share a legacy app/source data space. Sharing the same `appInstanceSpace` or APM_SourceData / IDM data across multiple Infield locations is no longer recommended and will be deprecated in the future.

### DIFF
--- a/cognite_toolkit/_cdf_tk/apps/_migrate_app.py
+++ b/cognite_toolkit/_cdf_tk/apps/_migrate_app.py
@@ -1492,7 +1492,7 @@ class MigrateApp(typer.Typer):
         external_id: Annotated[
             list[str] | None,
             typer.Argument(
-                help="The external IDs of the APM Configurations to create Infield V2 configurations for. "
+                help="The external IDs of the APM Configurations to create CDM configurations for Infield for. "
                 "If not provided, an interactive selection will be performed to select the APM Configurations."
             ),
         ] = None,
@@ -1501,7 +1501,7 @@ class MigrateApp(typer.Typer):
             typer.Option(
                 "--output-dir",
                 "-o",
-                help="Path to the directory where the Infield V2 configuration definitions will be dumped. It is recommended "
+                help="Path to the directory where the Infield CDM configuration definitions will be dumped. It is recommended "
                 "to govern these configurations in a git repository.",
             ),
         ] = Path("tmp"),
@@ -1514,7 +1514,7 @@ class MigrateApp(typer.Typer):
             ),
         ] = False,
     ) -> None:
-        """Creates Infield V2 configurations from existing APM Configurations in CDF."""
+        """Creates Infield CDM configurations from existing APM Configurations in CDF."""
         client = EnvironmentVariables.create_from_environment().get_client()
 
         cmd = MigrationCommand(client=client)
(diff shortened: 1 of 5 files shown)


cognite_toolkit/_cdf_tk/apps/_migrate_app.py
```diff
@@ -1492,7 +1492,7 @@ def infield_configs(
         external_id: Annotated[
             list[str] | None,
             typer.Argument(
-                help="The external IDs of the APM Configurations to create Infield V2 configurations for. "
```
Magssch, line 1495: The naming for InfieldOnCDM isn't fully determined, but v2 is confusing since there already technically are a lot of users on what we have called Infield 2.0

cognite_toolkit/_cdf_tk/commands/_migrate/creators.py
```diff
@@ -462,7 +484,7 @@ def _display_summary(self, success_count: int, skipped_external_ids_by_label: di
                 for label, identifiers in skipped_external_ids_by_label.items()
             ]
             items.append(
-                ItemsResult(status="failure", count=skipped_total, severity=Severity.failure.value, labels=labels)
+                ItemsResult(status="skipped", count=skipped_total, severity=Severity.skipped.value, labels=labels)
```
Magssch, line 486: Unrelated fix to confusing terminology reported earlier

cognite_toolkit/_cdf_tk/commands/_migrate/location_split.py
```diff
@@ -0,0 +1,67 @@
...
+                locations_by_source_space[root.source_data_instance_space].add(root.asset_external_id)
+
+    shared: set[str] = set()
+    for space, locations in locations_by_app_space.items():
+        if len(locations) > 1:
+            shared.add(space)
+    for space, locations in locations_by_source_space.items():
+        if len(locations) > 1:
+            shared.add(space)
+    return shared
```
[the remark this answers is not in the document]
Magssch: That is fine, if someone has used appDataInstanceSpace and sourceDataInstanceSpace as the same space then this is a misconfiguration in the first place and not something we can (or should) address with this PR.

ronpal (approved):

PR #3136, opened 2026-08-07
[prep CDF-28431] Extract shared Infield migration space-resolution logic

Magssch:
# Description

Deduplicates the nearly-identical source/target instance-space selection and validation logic in `infield-data` and `infield-source-data` into a single `MigrateApp._resolve_infield_migration_spaces` static method. Pulled out ahead of an upcoming location-split feature so that new branching logic only needs to be added in one place instead of two. No user-facing changes in this PR.

## Bump

- [ ] Patch
- [x] Skip


ronpal (approved):
🧹

PR #3138, opened 2026-08-07
Write query debug info as a single JSON file instead of NDJSON

Magssch:
# Description

The `QueryDebugResponse` written during `cdf data download instances` (query plan/profile diagnostics) was written using the NDJSON writer to stay consistent with other logging, despite always being a single object, producing a `.ndjson` file. I realized after the fact that deploy logs already produces `.json` files however, so this PR changes such that it generates as a plain, pretty-printed `.json` file instead, as this is much more convenient for this type of debugging file, matching the pattern already used by `DeployV2Command` for writing validation error details.

Keeping this as a skip as there is negligible user-facing difference

## Bump

- [ ] Patch
- [x] Skip


ronpal (approved):
👍

PR #3140, opened 2026-08-12
feat(transformations): add OneLake external data API client

Khyat-Cognite:
# Description

OneLake external data API client (PR 2/4). API client, ACL capability, and approval-client registration.

Stack: main -> models (#3098) -> api (this PR) -> deploy (#3142) -> dump (#3141)

## Bump

- [ ] Patch
- [x] Skip

### Added

- Transformation externaldata API client and TransformationExternalDataSourcesAcl.


cognite_toolkit/_cdf_tk/client/api/transformation_externaldata.py
```diff
@@ -0,0 +1,65 @@
...
+class TransformationExternalDataSourcesAPI(CDFResourceAPI[ExternalDataSourceResponse]):
+    def __init__(self, http_client: HTTPClient) -> None:
+        super().__init__(
+            http_client=http_client,
+            method_endpoint_map={
+                "upsert": Endpoint(method="POST", path="/transformations/externaldata", item_limit=1000),
+                "delete": Endpoint(method="POST", path="/transformations/externaldata/delete", item_limit=1000),
+                "list": Endpoint(method="GET", path="/transformations/externaldata", item_limit=1000),
+            },
+        )
```
ronpal, line 24: The [API doc](https://docs.cognite.com/api-reference/concepts/20230101-beta/transformation-external-data) requires `Cdf-Version: beta` on every request. `CDFResourceAPI` stores the `api_version` arg and passes it to each `RequestMessage`, which the HTTP client puts in the `cdf-version` header — falling back to `config.api_subversion` (defaults to `None`) when not set. Without this, every upsert/delete/list call hits the wrong API version.

Compare with the already-shipped `SignalSubscriptionsAPI`, which correctly uses `api_version="alpha"` for the same reason.

Note: `verify_usability` constructs its own `RequestMessage` directly, so it also needs `api_version=self._api_version` added there.

```suggestion
        super().__init__(
            http_client=http_client,
            method_endpoint_map={
                "upsert": Endpoint(method="POST", path="/transformations/externaldata", item_limit=1000),
                "delete": Endpoint(method="POST", path="/transformations/externaldata/delete", item_limit=1000),
                "list": Endpoint(method="GET", path="/transformations/externaldata", item_limit=1000),
            },
            api_version="beta",
        )
```

cognite_toolkit/_cdf_tk/client/api/transformation_externaldata.py
```diff
@@ -0,0 +1,65 @@
+from __future__ import annotations
+
```
ronpal:
```suggestion
```

Don't need futures

cognite_toolkit/_cdf_tk/client/api/transformation_externaldata.py
```diff
@@ -0,0 +1,65 @@
...
+from cognite_toolkit._cdf_tk.client.cdf_client.api import Endpoint
+from cognite_toolkit._cdf_tk.client.http_client import HTTPClient, ItemsSuccessResponse, RequestMessage, SuccessResponse
+from cognite_toolkit._cdf_tk.client.identifiers import ExternalId
+from cognite_toolkit._cdf_tk.client.resource_classes.externaldata import (
+    ExternalDataSourceRequest,
+    ExternalDataSourceResponse,
+)
+
+
+class ExternalDataSourceUsability(BaseModel):
```
ronpal:
```suggestion
class ExternalDataSourceUsability(BaseModelObject):
```

Response classes should use  BaseModelObject, which already has alias_generator=to_camel, populate_by_name=True, and extra="allow" (for forward compat).  

This class should live in resource_classes/externaldata.py as ExternalDataSourceUsabilityResponse(BaseModelObject), not inline in the API file. The two manual alias= annotations would then be dropped for free.

tests/test_unit/test_cdf_tk/test_client/test_transformation_externaldata.py
```diff
@@ -0,0 +1,96 @@
...
+        assert len(listed) == 1
+        batches = list(api.iterate(limit=10))
+        assert batches[0][0].external_id == "fabric-lakehouse-prod"
+
+        delete_url = toolkit_config.create_api_url("/transformations/externaldata/delete")
+        respx_mock.post(delete_url).mock(return_value=httpx.Response(status_code=200, json={}))
+        api.delete([ExternalId(external_id="fabric-lakehouse-prod")])
+        assert respx_mock.calls[-1].request.url.path.endswith("/transformations/externaldata/delete")
+
+    def test_verify_usability(self, toolkit_config: ToolkitClientConfig, respx_mock: respx.MockRouter) -> None:
```
ronpal, line 82: The test_verify_usability test (and the upsert test) mock the URL and assert the response, but never assert that the cdf-version: beta header was actually sent. Adding a assert respx_mock.calls[-1].request.headers["cdf-version"] == "beta" would be wise.

Khyat-Cognite: Done -- both tests now assert `cdf-version == "beta"`, and verify_usability passes `api_version=self._api_version` on its RequestMessage.

ronpal (approved):
Good to go!

PR #3141, opened 2026-08-12
feat(transformations): wire OneLake deps and dump

Khyat-Cognite:
# Description

Completes Toolkit support for transformation external data sources (Fabric OneLake): YAML deploy, dump, and `ext_onelake` dependencies.

## Bump

- [x] Patch
- [ ] Skip

### Added

- [alpha] Toolkit support for transformation external sources (Fabric).

### DIFF
--- a/cognite_toolkit/_cdf_tk/resource_ios/_resource_ios/externaldata.py
+++ b/cognite_toolkit/_cdf_tk/resource_ios/_resource_ios/externaldata.py
@@ -17,6 +17,7 @@ from cognite_toolkit._cdf_tk.client.resource_classes.group import (
 from cognite_toolkit._cdf_tk.exceptions import ToolkitRequiredValueError
 from cognite_toolkit._cdf_tk.resource_ios._base_ios import ResourceIO
 from cognite_toolkit._cdf_tk.resource_ios._resource_ios.data_organization import DataSetsIO
+from cognite_toolkit._cdf_tk.tk_warnings import HighSeverityWarning
 from cognite_toolkit._cdf_tk.utils import sanitize_filename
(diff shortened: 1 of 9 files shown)


cognite_toolkit/_cdf_tk/resource_ios/_resource_ios/externaldata.py
```diff
@@ -84,21 +85,23 @@ def load_resource(self, resource: dict[str, Any], is_dry_run: bool = False) -> E
...
     ) -> dict[str, Any]:
-        dumped = resource.as_request_resource().dump()
-        local = local or {}
+        HighSeverityWarning(
+            "External data sources will always be considered different, and thus will always be redeployed."
+        ).print_warning(console=self.client.console)
+        # Secrets cannot be compared (API never returns clientSecret). Dumping only the identifier
+        # when a local YAML exists makes deploy always upsert, matching hosted extractor sources.
+        if local:
+            return self.dump_id(self.get_id(resource))
```
ronpal:
```suggestion 
        if local:
            HighSeverityWarning(
            "External data sources will always be considered different, and thus will always be redeployed."
        ).print_warning(console=self.client.console)
        # Secrets cannot be compared (API never returns clientSecret). Dumping only the identifier
        # when a local YAML exists makes deploy always upsert, matching hosted extractor sources.
            return self.dump_id(self.get_id(resource))
```

Khyat-Cognite: Fixed it.

ronpal (approved):
Nice!

PR #3142, opened 2026-08-12
feat(transformations): add OneLake external data deploy IO

Khyat-Cognite:
# Description

OneLake external data deploy IO (PR 3/4). ExternalDataSourceIO deploy/pull, ext_onelake parser, and auth capability skip.

Stack: main -> models (#3098) -> api (#3140) -> deploy (this PR) -> dump (#3141)

## Bump

- [ ] Patch
- [x] Skip

### Added

- ExternalDataSourceIO deploy/pull and ext_onelake query parser, gated by `alpha_flags.external-data-sources`.


ronpal:
For APIs still in  `beta`, we use the Flags gate feature in the Toolkit. Please add a new flag in `_cdf_tk/feature_flags.py` and see how it is used to gate functionality in`cognite_toolkit/_cdf_tk/resource_ios/__init__.py `

Khyat-Cognite:
> For APIs still in `beta`, we use the Flags gate feature in the Toolkit. Please add a new flag in `_cdf_tk/feature_flags.py` and see how it is used to gate functionality in`cognite_toolkit/_cdf_tk/resource_ios/__init__.py `

Added the beta gate, Thanks ^^

cognite_toolkit/_cdf_tk/resource_ios/_resource_ios/externaldata.py
```diff
@@ -0,0 +1,142 @@
...
+            return
+        if data_set_external_id is None:
+            yield from self.client.tool.transformations.external_data_sources.list(limit=None)
+            return
+        data_set_id = self.client.lookup.data_sets.id(data_set_external_id)
+        if data_set_id is None:
+            raise ToolkitRequiredValueError(f"DataSet {data_set_external_id!r} does not exist")
+        for source in self.client.tool.transformations.external_data_sources.list(limit=None):
+            if source.data_set_id == data_set_id:
+                yield source
```
ronpal, line 142: There are cases where a missing dataset is legit, for example if the user is running dry-run and the dataset doesn't exist _yet_. So raising the exception in line 138/139 is wrong. It should just return empty instead. 

The validation that a dataset must exist is already enforced by load_resource (the local→CDF direction), which correctly passes is_dry_run and resolves the ID before constructing the request.

tests/test_unit/utils.py
```diff
@@ -347,6 +347,8 @@ def create_pydantic_instance(self, model_cls: type[BaseModel], skip_defaulted_ar
             else:
                 value = self.create_value(field.annotation, var_name=field_id)
             keyword_arguments[name] = value
+        if model_cls.__name__ in {"ExternalDataSourceRequest", "ExternalDataSourceResponse"}:
+            keyword_arguments["format"] = "one_lake"
```
ronpal: This will silently break if either class is renamed. The better approach is to either make format required on ExternalDataSourceResponse (typed as Literal["one_lake"] there too, or at minimum non-optional) so the factory generates it naturally.

tests/test_unit/test_cdf_tk/test_cruds/test_base.py
```diff
@@ -241,6 +244,18 @@ def test_resource_types_is_up_to_date() -> None:
     assert not extra, f"Extra {extra=}"
 
 
+def test_external_data_source_io_respects_feature_flag() -> None:
```
ronpal: I don't think we need this. We have other backstops/safeties :)

cognite_toolkit/_cdf_tk/resource_ios/_resource_ios/externaldata.py
Khyat-Cognite: Agreed -- missing dataset is valid in dry-run. _iterate now returns empty; existence stays in load_resource with is_dry_run.

tests/test_unit/utils.py
Khyat-Cognite: Fixed -- Typed ExternalDataSourceResponse.format as Literal["one_lake"] with the same default as the request class, so the factory generates it without a class-name special case.

ronpal (approved):

PR #3147, opened 2026-08-13
[CDF-28191] Remove (space,externalId) sort from all instance download queries 🧹 

Magssch:
# Description

Stop sending `sort`/`backfillSort` on all instance query / sync requests (nodes, edges, view selectors, and InField/360 image migration queries). Queries now use the server default order instead of `(space, externalId)`.

Many internal sources claim that sorting by space+externalId helps mitigate 408s and the original intent of applying this filter was to avoid timeouts in cases of high amounts of soft-deletes. As demonstrated in practice when testing on certain large-scale customer projects however, this seems to in contrast be detrimental to performance in some cases: https://cognitedata.slack.com/archives/C05RW13C6TX/p1786636988060219?thread_ts=1782830091.542489&cid=C05RW13C6TX

Therefore we are removing the sort and going back to the default for now, until we have any indication that we can find any better defaults / catch-all sorts than what is given out of the box by DMS.

## Bump

- [x] Patch
- [ ] Skip

### Improved

- Improved performance in certain cases when downloading data modeling instances through toolkit.

### DIFF
--- a/cognite_toolkit/_cdf_tk/client/api/instances.py
+++ b/cognite_toolkit/_cdf_tk/client/api/instances.py
@@ -40,7 +40,6 @@ from cognite_toolkit._cdf_tk.client.resource_classes.data_modeling._query import
     QueryResponseUntyped,
     QuerySelect,
     QuerySelectSource,
-    QuerySortSpec,
 )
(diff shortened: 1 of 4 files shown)


ronpal (approved):
🧹 👍

PR #3155, opened 2026-08-17
(fix): cdf modules add command -d flag multiple package error fix 🛠️

Aashutosh-cognite:
# Description

`cdf modules add -d <module>` now supports a `<package>:<module>` syntax to select a specific module when its name exists in more than one package.

- `cdf modules add -d <module>` — works as before when the module exists in exactly one package.
- `cdf modules add -d <package>:<module>` — installs that module from the specified package only.
- `cdf modules add -d <module>` (unprefixed, module exists in multiple packages) — fails with an error naming the conflicting packages and showing the exact `<package>:<module>` syntax to resolve it.
- `cdf modules add -d <package>:<module>` where the package doesn't exist, or exists but doesn't contain that module — fails with a specific error identifying which of the two is wrong.
- Whole-package selection (`cdf modules add -d <package>`) is unchanged.

This makes non-interactive module installation fully deterministic and scriptable, which is especially useful for AI agents driving the Toolkit CLI — they can now resolve and install any module in a single, unambiguous command without needing a human to interactively pick between conflicting packages.

Updated the `--help` text for `-d/--deployment-pack` to document the new syntax.

## Bump

- [x] Patch
- [ ] Skip

### Added

- `cdf modules add -d <package>:<module>` syntax to disambiguate and install a module that exists in more than one deployment package, enabling fully non-interactive installs (e.g. for AI agents) even in ambiguous cases.

### DIFF
--- a/cognite_toolkit/_cdf_tk/commands/modules.py
+++ b/cognite_toolkit/_cdf_tk/commands/modules.py
@@ -620,6 +620,14 @@ class ModulesCommand(ToolkitCommand):
         """
         Find a module in a specific package using the '<package>:<module>' syntax and
         return it as a Packages selection.
+
+        Args:
+            packages (Packages): Available packages
+            package_and_module (str): Package and module name
+            existing (set[str]): Installed module names
+
+        Returns:
+            Packages: Selected package and module
         """
         package_part, _, module_part = package_and_module.partition(":")
         if not package_part or not module_part:
--- a/cognite_toolkit/_cdf_tk/commands/modules.py
+++ b/cognite_toolkit/_cdf_tk/commands/modules.py
@@ -536,8 +536,15 @@ class ModulesCommand(ToolkitCommand):
         """
         Find a package or module by name and return it as a Packages selection.
         Raises ToolkitError if nothing is found or every matched module is already installed.
+
+        The module_name may use the '<package>:<module>' syntax to disambiguate a module
+        that exists in more than one package.
         """
         existing = set(existing_module_names)
+
+        if ":" in module_name:
+            return self._find_and_select_module_in_package(packages, module_name, existing)
+
         name_lower = module_name.casefold()
 
         by_package = {name.casefold(): pkg for name, pkg in packages.items()}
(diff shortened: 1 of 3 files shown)


cognite_toolkit/_cdf_tk/commands/modules.py
```diff
@@ -595,6 +611,53 @@ def _find_and_select_module(
...
+        Args:
+            packages (Packages): Available packages
+            package_and_module (str): Package and module name
+            existing (set[str]): Installed module names
+
+        Returns:
+            Packages: Selected package and module
+        """
+        package_part, _, module_part = package_and_module.partition(":")
+        if not package_part or not module_part:
```
ronpal, line 638: How does this fare if you get a string with multiple colons? Like `dp:contextualization:cdf_entity_matching`? Will it still work?

cognite_toolkit/_cdf_tk/commands/modules.py
```diff
@@ -595,6 +611,53 @@ def _find_and_select_module(
...
+            raise ToolkitError(f"Invalid syntax '{package_and_module}'. Expected format '<package>:<module>'.")
+
+        by_package = {name.casefold(): pkg for name, pkg in packages.items()}
+        package = by_package.get(package_part.casefold())
+        if package is None:
+            raise ToolkitError(f"Package '{package_part}' not found.")
```
ronpal: Suggest you use difflib.get_close_matches here and below to get "Did you mean..."

cognite_toolkit/_cdf_tk/commands/modules.py
Aashutosh-cognite: good catch - will raise a syntax error incase of more than one colon present.

cognite_toolkit/_cdf_tk/commands/modules.py
Aashutosh-cognite: Will implement this.

cognite_toolkit/_cdf_tk/commands/modules.py
Aashutosh-cognite: Addressed in [d633db8](https://github.com/cognitedata/toolkit/pull/3155/commits/d633db8a85dd7c83235ce681c825604c417fe1ff)

cognite_toolkit/_cdf_tk/commands/modules.py
Aashutosh-cognite: Addressed in [d633db8](https://github.com/cognitedata/toolkit/pull/3155/commits/d633db8a85dd7c83235ce681c825604c417fe1ff)

ronpal (approved):

PR #3160, opened 2026-08-25
[CDF-28496] Validate agent models and runtime capabilities against the /ai/services/availability endpoint

Magssch:
# Description

Replaces the hardcoded list of supported agent language models (and the hardcoded runtime-version-supports-subagents set) with a live check against the `/ai/services/availability` endpoint. `model` is no longer a strict enum in the local schema, so new models no longer trigger a syntax error; instead, a new `AgentRules` global rule validates model, tools-per-agent limit, and runtime capabilities (subagents, skills) against the CDF project's actual availability, falling back to allowing any value when the endpoint can't be reached. This mirrors the existing limits validation functionality used for the Functions service.

## Bump

- [x] Patch
- [ ] Skip

### Improved

- Agent `model` values are now validated against your CDF project's actual available models, tools-per-agent limit, and runtime version capabilities (e.g. subagents, skills) when a client is available, and accepts any value if that check can't be performed.

### Fixed

- Agent `dataModels.dataModels[].viewExternalIds` is no longer incorrectly required, and its max length is now 20 instead of 10.

### DIFF
--- a/cognite_toolkit/_cdf_tk/yaml_classes/agent.py
+++ b/cognite_toolkit/_cdf_tk/yaml_classes/agent.py
@@ -17,7 +17,6 @@ else:
 MAX_SUB_AGENTS_PER_AGENT = 20
 EXAMPLE_QUESTIONS_MAX_LENGTH = 5
 EXAMPLE_QUESTIONS_MAX_SERIALIZED_SIZE = 40960
-RUNTIME_VERSIONS_SUPPORTING_SUBAGENTS = frozenset({"1.3.0", "1.4.0"})
 
 
 class AgentToolModelResource(BaseModelResource, extra="allow"): ...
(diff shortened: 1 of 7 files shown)


cognite_toolkit/_cdf_tk/yaml_classes/agent.py
```diff
@@ -104,7 +103,9 @@ class AgentDataModel(AgentToolModelResource):
         max_length=43,
         pattern=DM_VERSION_PATTERN,
     )
-    view_external_ids: list[str] = Field(description="The views of the data model.", min_length=1, max_length=10)
```
Magssch, line 107: Pre-existing bug, verified and changed according to docs: https://api-docs.cognite.com/20230101-alpha/tag/Agents/operation/get_agents_by_ids_ai_agents_byids_post#!c=200&path=items/tools/3/configuration/dataModels/0/dataModels/viewExternalIds&t=response

ronpal (approved):
Great improvement!

PR #3162, opened 2026-08-25
[CDF-28497] Validate Cognite Function runtimes dynamically against project limits

Magssch:
# Description

Cognite Functions supports new runtimes over time (e.g. `py313`, `py314`), which previously required hardcoding an updated `Literal` list in the toolkit for every release. This PR replaces that static list with dynamic validation against the `/functions/limits` endpoint's `runtimes` field, and removes the toolkit's hardcoded `py311` default for the `runtime` field so that functions without an explicit runtime always defer to CDF's own (evolving) default instead of being silently pinned to a stale version. We already used `/functions/limits` for other types of validations, so this is simply closing a gap introduced after #3121 where we temporarily allowed any string to pass through if unknown.

## Bump

- [x] Patch
- [ ] Skip

### Changed

- Cognite Function `runtime` is now validated dynamically against the CDF project's available runtimes

### DIFF
--- a/cognite_toolkit/_cdf_tk/resource_ios/_resource_ios/function.py
+++ b/cognite_toolkit/_cdf_tk/resource_ios/_resource_ios/function.py
@@ -405,10 +405,6 @@ class FunctionIO(ResourceIO[ExternalId, FunctionRequest, FunctionResponse]):
             if key in local and local[key] in {None, ""} and key not in dumped:
                 dumped[key] = local[key]
 
-        if dumped.get("runtime") == "py312" and "runtime" not in local:
-            # Remove the default value of the runtime for Python 3.12, as it is the default runtime and does not need to be specified.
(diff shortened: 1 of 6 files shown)


cognite_toolkit/_cdf_tk/resource_ios/_resource_ios/function.py
```diff
@@ -405,10 +405,6 @@ def dump_resource(self, resource: FunctionResponse, local: dict[str, Any] | None
             if key in local and local[key] in {None, ""} and key not in dumped:
                 dumped[key] = local[key]
 
-        if dumped.get("runtime") == "py312" and "runtime" not in local:
```
Magssch, line 408: This was just dead code, runtime is already popped unconditionally further up if "runtime" not in local

ronpal (approved):

PR #3163, opened 2026-08-25
[CDF-28431] Improve migration progress messages and fix custom mapper dry-run/logging propagation

Magssch:
# Description

Cleans up migration UX: progress table now shows the selector's `display_name` instead of `str()`, and the FDM-to-CDM conversion description clarifies instance type/space/view info. Also fixes custom instance mappers not inheriting `logger` and `dry_run` from the parent mapper, causing them to log incorrectly and (hypothetically, but not in practice currently) always apply writes even during dry runs.

## Bump

- [x] Patch
- [ ] Skip

### Improved

- [alpha] Show clearer progress descriptions during instance migration, including instance type, space, and destination view.

### Fixed

- [alpha] Fix Infield Schedules migrations not logging migration results correctly.

### DIFF
--- a/cognite_toolkit/_cdf_tk/commands/_migrate/data_mapper.py
+++ b/cognite_toolkit/_cdf_tk/commands/_migrate/data_mapper.py
@@ -1392,14 +1392,17 @@ class FDMtoCDMMapper(DataMapper[InstanceSelector, NodeOrEdgeResponse, NodeOrEdge
 
     def process_description(self, source_selector: InstanceSelector) -> str:
         destination_view = self._get_destination_view(source_selector)
-        if destination_view is None:
+        if destination_view is None or not isinstance(source_selector, InstanceViewSelector):
             return "Converting"
-        message = f"Converting into view {destination_view!s}"
+        message = f"Converting into {source_selector.instance_type}s"
         source_spaces = source_selector.get_instance_spaces()
         if source_spaces:
             destination_spaces = self._connection_creator.get_destination_spaces(source_spaces)
-            if destination_spaces:
-                message += f" with {humanize_collection(destination_spaces)} instance spaces"
+            if len(destination_spaces) == 1:
+                message += f" in instance space {destination_spaces[0]}"
+            elif len(destination_spaces) > 1:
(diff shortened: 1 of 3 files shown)


cognite_toolkit/_cdf_tk/commands/_migrate/data_mapper.py
```diff
@@ -1437,7 +1440,10 @@ def map(self, source: Sequence[DataItem[NodeOrEdgeResponse]]) -> Sequence[DataIt
         ):
             if len(intersecting_view_ids) == 1:
                 intersection_view_id = next(iter(intersecting_view_ids))
-                custom_mapped = self._custom_instance_mappings[intersection_view_id].map(source)
+                custom_mapper = self._custom_instance_mappings[intersection_view_id]
+                custom_mapper.logger = self.logger
+                custom_mapper.dry_run = self.dry_run
```
Magssch, line 1445: This is a bit dirty, but it stops a potential footgun that I've had the displeasure of tripping before, and fixes the logging for schedules so they actually show up (currently all schedule-related errors are swallowed / sent to a no-op logger)

cognite_toolkit/_cdf_tk/commands/_migrate/data_mapper.py
```diff
@@ -1974,7 +1997,11 @@ def _find_schedule_edges(
                 template_item_edges.append(template_item_edge)
                 for template_edge in template_edges_by_item_id.get(template_item_edge.other_side, []):
                     template_edges.append(template_edge)
-        return template_edges, template_item_edges
+        # Deduplicate by target: multiple edges to the same template/template item is expected
```
Magssch, line 1983: Surfacing logger warnings will now also surface every duplicated schedule as a warning that it will be truncated/deduplicated. That would lead to a lot of unnecessary noise (the fact that we need to deduplicate is the whole point of this mapper after all), so this "resupresses" this expected case while keeping other truly unexpected errors surfaced. As an additional benefit, line 2003 will ensure the `Schedule.templateItems` list of DR does not contain duplicated TemplateItem nodeIds, which could happen before this change.

ronpal (approved):

PR #3167, opened 2026-08-26
[CDF-28431] Add location-split-aware Infield data mappers

Magssch:
# Description

Add `LocationSplitFDMtoCDMMapper` and `LocationSplitSolutionTagMapper` that resolve each instance's per-location CDM space (via `rootLocation`, parent edge/property, or a referenced classic asset) and tag it with `InstanceSpaceRelocationSource`. Edges inherit the owning node's space; schedules inherit from their template and are grouped per template so identical configs on different templates are not collapsed. Mapping YAML is reordered parent-before-child so target spaces can be inherited by dependents. Not yet wired into the migrate command.

## Bump

- [ ] Patch
- [x] Skip


cognite_toolkit/_cdf_tk/commands/_migrate/issues.py
```diff
@@ -187,16 +187,6 @@ def label(self) -> str:
         return "Instance conversion"
 
 
-class TargetSpaceResolutionIssue(InstanceConversionIssue):
```
Magssch, line 190: Changed my mind about adding this

cognite_toolkit/_cdf_tk/commands/_migrate/data_mapper.py
```diff
@@ -1775,6 +1792,189 @@ def _iterate_constrained_direct_relation_properties(
...
+        super().__init__(
+            client,
+            mappings,
+            connection_creator,
+            custom_properties_mappings=custom_properties_mappings,
+            custom_instance_mappings=custom_instance_mappings,
+        )
+        self._instance_id_mapper = instance_id_mapper
+        self._target_by_root_asset = target_by_root_asset
+        self._source_views = dict(source_views) if source_views is not None else None
```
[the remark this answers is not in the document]
Magssch: I don't think this is worth it, it's only called in one place

cognite_toolkit/_cdf_tk/commands/_migrate/location_split.py
```diff
@@ -17,6 +20,59 @@
...
+    """Direct relations are ``{space, externalId}`` dicts; some APM properties are plain strings."""
+    if isinstance(value, str):
+        return value
+    if isinstance(value, dict):
+        external_id = value.get("externalId")
+        return external_id if isinstance(external_id, str) else None
+    return None
+
 
 def _iter_root_locations(
```
doctrino: I would add this function as a method on the APMConfigResponse. It is a utility method for that class

```python
class APMConfigResponse;
    ....
    @classmethod
    def iterate_root_locations(cls, configs: Sequence[Self]) -> ...
```

doctrino (dismissed):
Complex one 😅

I have gone through the tests, and it is far as I can see solid. Hard to follow all of the logic and edge cases, but as this is behind a feature flag, lets release and do a full scale test of it.

doctrino (approved):

PR #3179, opened 2026-09-01
[CDF-28526] 🧼 Cleanup tracking

doctrino:
# Description

These are not used. So either we remove them, or we need to add them to the build/deploy v2.

## Bump

- [ ] Patch
- [x] Skip


cognite_toolkit/_cdf_tk/data_classes/_tracking_info.py
```diff
@@ -37,26 +37,19 @@ class CommandTracking(TrackingEvent):
...
         downloaded_module_ids: List of module IDs that were downloaded.
     """
 
-    module_ids: set[str] = Field(default_factory=set)
-    package_ids: set[str] = Field(default_factory=set)
     installed_module_ids: set[str] = Field(default_factory=set)
     installed_package_ids: set[str] = Field(default_factory=set)
     downloaded_library_ids: set[str] = Field(default_factory=set)
     downloaded_package_ids: set[str] = Field(default_factory=set)
     downloaded_module_ids: set[str] = Field(default_factory=set)
```
doctrino, line 51: Note these are currently in use, I think they are safer to keep. But can be considered whether to remove or not. Alternative, create a new specific tracking event for the modules commands, same as we do for build/deploy/data.

ronpal (approved):

PR #3182, opened 2026-09-01
[CDF-28532]🪟 Support record views

doctrino:
# Description

Currently we get the following warnings of `streamId` is set
image

## Bump

- [x] Patch
- [ ] Skip

### Improved

- Toolkit no longer gives a syntax warning when deploying a record view. In addition, Toolkit now checks that record view only have container properties and gives a syntax warning if it does not.

### DIFF
--- a/tests/test_unit/test_cdf_tk/test_resource_classes/test_view.py
+++ b/tests/test_unit/test_cdf_tk/test_resource_classes/test_view.py
@@ -155,6 +155,87 @@ def view_with_filters_cases() -> Iterable:
     yield from (pytest.param(next(iter(f.values())), id=next(iter(f.keys()))) for f in view_filters)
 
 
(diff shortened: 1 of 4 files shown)


tests/test_unit/test_cdf_tk/test_resource_classes/test_view.py
```diff
@@ -155,103 +155,119 @@ def view_with_filters_cases() -> Iterable:
     yield from (pytest.param(next(iter(f.values())), id=next(iter(f.keys()))) for f in view_filters)
 
 
+def invalid_view_cases() -> Iterable:
```
doctrino, line 158: Standardizing these tests, same as we have for all other yaml classes

doctrino:
Included in #3171

PR #3187, opened 2026-09-02
[prep CDF-28524]🏞 Extend build v2 to cover v1 functionality

doctrino:
# Description

Preparation for replacing build v1 in `cdf modules list/pull` and `cdf dev run`

Introducing a method `tmp_build` that allows to build to get an overview over all resources and modules that are available. It has caching to avoid rebuilding everything if thhere has been no changes to the source file. 

In addition, a few renaming for clarity. 

## Bump

- [ ] Patch
- [x] Skip


Magssch:
Quick question, should we perhaps also invalidate the cache if `config.<env>.yaml` which injects variables is changed?

Magssch (dismissed):

doctrino:
> Quick question, should we perhaps also invalidate the cache if `config.<env>.yaml` which injects variables is changed?

Of course, you are right.

Magssch (approved):

PR #3189, opened 2026-09-02
[CDF-28521] Skip deleted timeseries when migrating charts

Magssch:
# Description

Charts failed to migrate if they referenced a timeseries missing from both Classic and CDM (deleted). Those references are now left as classic `timeSeriesCollection` entries with a warning, and the rest of the chart is migrated. The reference is retained here in order to make the migration idempotent, in case the missing timeseries is due to some kind of error or possibly wrongly deleted timeseries, such that the references to them can be migrated again later if needed. Charts that still have **unmigrated** Classic timeseries continue to fail as per before.

Also fixes an unrelated issue where we stop counting failed/skipped items as migrated in the summary, and drop the item-count prefix from chart/canvas selector display names to avoid showing the count twice as it did before: `Migrated 1 1 charts to instances.`

## Bump

- [x] Patch
- [ ] Skip

### Fixed

- [alpha] Fix `cdf migrate charts` failing for a Chart when a referenced timeseries in it has been deleted. Toolkit will now successfully migrate these Charts despite referencing some non-existing timeseries.


cognite_toolkit/_cdf_tk/commands/_migrate/data_mapper.py
```diff
@@ -505,7 +510,7 @@ def map(self, source: Sequence[DataItem[ChartResponse]]) -> Sequence[DataItem[Ch
                 self.logger.log(
                     MigrationEntryV2(
                         id=identifier,
-                        label="No asset-centric timeseries in Chart.",
```
Magssch, line 508: To standardize the messages

cognite_toolkit/_cdf_tk/commands/_migrate/command.py
```diff
@@ -201,8 +201,11 @@ def _run_migration_steps(
             # download pages can include auxiliary edges that are, for example, converted to direct relations
             # on a node after being migrated, alongside the nodes to migrate themselves. It would be confusing
             # to a user if those edges were counted and displayed, since the initial estimate only counts the nodes to migrate.
-            total = sum(result.count for result in items_results)
-            console.print(f"{action} {total:,} {selected.display_name} to {target}.")
+            migrated_count = sum(
```
Magssch, line 204: To avoid counting non-migrated items as migrated

cognite_toolkit/_cdf_tk/dataio/selectors/_canvas.py
```diff
@@ -16,7 +16,3 @@ class CanvasExternalIdSelector(CanvasSelector):
     def __str__(self) -> str:
         hash_ = hashlib.md5(",".join(sorted(self.external_ids)).encode()).hexdigest()[:8]
         return f"canvas_count_{len(self.external_ids)}_hash_{hash_}"
-
-    @property
-    def display_name(self) -> str:
-        return f"{len(self.external_ids)} {self.kind}"
```
doctrino, line 22: Are these unused? Why are they removed?

cognite_toolkit/_cdf_tk/dataio/selectors/_charts.py
```diff
@@ -35,7 +35,3 @@ class ChartExternalIdSelector(ChartSelector):
     def __str__(self) -> str:
         hash_ = hashlib.md5(",".join(sorted(self.external_ids)).encode()).hexdigest()[:8]
         return f"chart_count_{len(self.external_ids)}_hash_{hash_}"
-
-    @property
-    def display_name(self) -> str:
```
doctrino, line 40: Same question here, why are these removed? 

As far as I remember, I added this to have a consistent display name everywhere when referencing charts/canvases.

doctrino (commented):
Just an open question

cognite_toolkit/_cdf_tk/dataio/selectors/_canvas.py
Magssch: The way it was setup it caused the text to show `Migrated 1 1 charts to instances` at the end of the migration, and the number of migrated items is already printed in MigrationCommand._run_migration_steps. So the count was duplicated due to the `len(self.external_ids)`. Removing that means it's just `self.kind` left, and the default implementation in DataSelector is just returning self.kind so there shouldn't be a need to override it anymore.

cognite_toolkit/_cdf_tk/dataio/selectors/_charts.py
Magssch: Answered above

doctrino (approved):

PR #3191, opened 2026-09-03
[CDF-28536, CDF-28542] 🤗Stabilize InField migrate snapshot and register external data source smoke tests

doctrino:
## Description

Fix two failing greenfield smoke tests:

- Ignore `FieldObservation.startTime` in the InField migrate snapshot comparison. It is set to the current time when missing on the source instance, so it changes on every run.
- Register `TransformationExternalDataSourcesAPI` in the generic `test_crudl` smoke tests by adding `ExternalDataSourceResponse` example data, following the existing CDFResourceAPI pattern.

## Bump

- [ ] Patch
- [x] Skip

### DIFF
--- a/tests_smoke/test_commands/test_migrate_infield.py
+++ b/tests_smoke/test_commands/test_migrate_infield.py
@@ -327,10 +327,14 @@ class TestMigrateInfield:
                     exclude={"last_updated_time", "version", "created_time"},
                 )
                 for view_properties in dumped.get("properties", {}).values():
-                    for properties in view_properties.values():
+                    for view_name, properties in view_properties.items():
                         # These also change for each run.
                         properties.pop("sourceCreatedTime", None)
                         properties.pop("sourceUpdatedTime", None)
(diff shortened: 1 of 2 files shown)


tests_smoke/test_commands/test_migrate_infield.py
```diff
@@ -327,10 +327,14 @@ def test_migrate_data(
                     exclude={"last_updated_time", "version", "created_time"},
                 )
                 for view_properties in dumped.get("properties", {}).values():
-                    for properties in view_properties.values():
+                    for view_name, properties in view_properties.items():
                         # These also change for each run.
                         properties.pop("sourceCreatedTime", None)
                         properties.pop("sourceUpdatedTime", None)
+                        # FieldObservation.startTime is populated with the current time when
+                        # missing on the source instance.
```
doctrino:
```suggestionnce.
```

Magssch (approved):

PR #3192, opened 2026-09-03
[CDF-28543] 🤨Add transformationsExternalDataSourcesAcl

doctrino:
# Description
Adds support for the `transformationsExternalDataSourcesAcl` group capability so Toolkit can parse, validate, and deploy this ACL in both YAML group configs and API group resources. Actions are READ, WRITE, and USE; scopes are all and datasetScope.

## Bump
- [x] Patch
- [ ] Skip

### Added
- Toolkit no longer gives a syntax warning when you build a group with a `transformationsExternalDataSourcesAcl` capability in a group.


tests/test_unit/test_cdf_tk/test_client/test_group_api_classes.py
```diff
@@ -226,9 +233,7 @@ def test_serialize_deserialize_known_capability_with_unknown_action(self) -> Non
     def test_capability_in_sync(self) -> None:
         """Checks that the request/response capabilities are in sync with the YAML spec."""
         request_capabilities = {acl.model_fields["acl_name"].default for acl in get_all_subclasses(Acl)} - {
-            "unknownAcl",
-            # Until cognite-sdk publishes this capability class.
```
doctrino, line 230: We do not depend on cognite-sdk, so no need to have this exception.

Magssch (approved):

PR #3197, opened 2026-09-03
[CDF-28524] 📋 Improve terminal output for cdf modules list

doctrino:
# Description

`cdf modules list` now uses Build v2 `tmp_build` and `BuildLineage` instead of the v1 `ModuleResources` listing path. Terminal output follows the Build v2 panel and table style, and each module shows insight types plus syntax warnings instead of a single build-warning count.

## Before

image

## After

image

## Bump

- [x] Patch
- [ ] Skip

### Improved

- Improved terminal output for `cdf modules list`


ronpal (approved):
Nice!

PR #3204, opened 2026-09-04
[CDF-28550] Topologically sort agents by subagent references on deploy and delete

Magssch:
# Description

The agents service validates that `subagents` references point to existing agents on upsert, and rejects deleting an agent still referenced as a subagent by another agent. `AgentIO` previously deployed/deleted agents in build/file order, so an agent depending on another agent as a subagent could be sent to the API before the subagents it references, causing deployment failures. Similarly, a delete will fail if the agents referencing it as a subagent isn't deleted first. Added `AgentIO.topological_sort`, used on `create`/`update` to deploy subagents before referencing agents, and reversed on `delete` to remove referencing agents before their subagents.

## Bump

- [x] Patch
- [ ] Skip

### Fixed

- [alpha] Fixes sporadic `400` and `409` HTTP errors in `cdf deploy` and `cdf clean` when an agent references a subagent


doctrino (approved):
Great stuff.

PR #3207, opened 2026-09-07
[CDF-28525] 🗿 Remove build v1 from run commands

doctrino:
# Description

Replace ModuleResources which depends on Build v1 with BuildV2 and the `build_tmp` function.

## Bump

- [ ] Patch
- [x] Skip

### DIFF
--- a/cognite_toolkit/_cdf_tk/commands/run.py
+++ b/cognite_toolkit/_cdf_tk/commands/run.py
@@ -722,7 +722,7 @@ class RunWorkflowCommand(ToolkitCommand):
         credentials: ClientCredentials | None = None
         input_: dict | None = None
         for trigger in triggers:
-            trigger_dict = trigger.get_resource_dict(env_vars.dump(), validate=False)
+            trigger_dict = trigger.load_resource_dict(env_vars.dump(), validate=False)
             if (
                 trigger_dict["workflowExternalId"] == id_.workflow_external_id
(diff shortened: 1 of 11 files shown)


cognite_toolkit/_cdf_tk/commands/run.py
```diff
@@ -284,33 +298,38 @@ def _clean_environ_var(value: str) -> str:
 
     @staticmethod
     def _get_call_args_interactive(
-        function_external_id: str, resources: ModuleResources
+        function_external_id: str, build_folder: BuildLineage, environment_variables: dict[str, str | None]
```
doctrino: Note that there was a bug here that environment variables were not passed when loading. Note that it is unlikely that this had any impact as we only load the function definitions to get their IDs, and environment variables are typically never used in IDs.

cognite_toolkit/_cdf_tk/commands/run.py
```diff
@@ -341,17 +360,29 @@ def _get_call_args_interactive(
 
     @staticmethod
     def _geta_call_args_from_data_source(
-        data_source: str | WorkflowVersionId, function_external_id: str, resources: ModuleResources
+        data_source: str | WorkflowVersionId,
+        function_external_id: str,
+        build_folder: BuildLineage,
+        environment_variables: dict[str, str | None],
```
doctrino: Same as above.

ronpal (approved):

PR #3208, opened 2026-09-07
[CDF-28553] Add lineageAnnotation support to workflow version tasks

Magssch:
# Description

Adds `lineageAnnotation` (sources/targets as CDF resource URIs) to workflow version tasks, in both the YAML class and the request/response resource classes. The field is gated behind the existing `data_products` alpha flag since even though it requires its own feature flag in CDF, it's main purpose is to be used with data products. The YAML validator rejects it when the flag is off, and `dump_resource` strips it from CDF responses so it does not show up in diffs/dumps. Also adds identifiable list diffing on `uri` for lineage sources/targets, and finally makes the `data_products` flag visible in `cdf about` since the API itself is now publicly documented.

## Bump

- [x] Patch
- [ ] Skip

### Added

- [alpha] Toolkit will now recognize `lineageAnnotation` on workflow version tasks, declaring the data sources and targets of each task. Requires the `data_products` alpha flag.


cognite_toolkit/_cdf_tk/feature_flags.py
```diff
@@ -67,8 +67,8 @@ class Flags(Enum):
         description="Enables validation of function requirements.txt during build using pip dry-run",
     )
     DATA_PRODUCTS = FlagMetadata(
-        visible=False,
-        description="Enables support for data product resources",
+        visible=True,
```
Magssch, line 70: Making this visible since data products is now publicly documented

doctrino (approved):

PR #3212, opened 2026-09-07
[CDF-27466] 😄 Rewrite asset centric mapping to pydantic

doctrino:
# Description

Drop private dependency on Python-SDK

## Bump

- [ ] Patch
- [x] Skip

### DIFF
--- a/cognite_toolkit/_cdf_tk/commands/_migrate/data_classes.py
+++ b/cognite_toolkit/_cdf_tk/commands/_migrate/data_classes.py
@@ -1,13 +1,7 @@
-import sys
 from collections.abc import Sequence
-from dataclasses import dataclass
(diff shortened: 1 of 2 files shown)


cognite_toolkit/_cdf_tk/commands/_migrate/data_classes.py
```diff
@@ -1,13 +1,7 @@
-import sys
 from collections.abc import Sequence
-from dataclasses import dataclass
 from pathlib import Path
 from typing import Any, Generic, Literal
 
-from cognite.client.data_classes._base import (
```
doctrino, line 7: One of two remaining private dependencies.

ronpal (approved):

PR #3218, opened 2026-09-08
Fix skills create to upload SKILL.md via /ai/skills/upload

ronpal:
# Description

Skill create/update posted JSON to `/ai/skills`, which creates a CDF file record without uploading the markdown bytes. Greenfield smoke tests then failed with `Files not uploaded`. Create and update now use multipart `POST /ai/skills/upload`.

Also fixes wrong URL to doc for transformation external source. 

## Bump

- [x] Patch
- [ ] Skip

### Fixed

- Create and update Atlas AI skills via `/ai/skills/upload` so skill content is actually uploaded
- Wrong URL to doc for transformation external source


doctrino (approved):
Good stuff

PR #3224, opened 2026-09-09
[CDF-28568] 🧽 Remove unused utils

doctrino:
# Description

Added test `test_utils_module_independent` to ensure that we do not have dependencies inside the utils on other part of the code base (with some reasonable exceptions), then ensured that that test passed along with deleting unused code inside utils.

Importantly, this removes the duplicated logic with functions like `module_iterate` that is now covered by the `ModuleParser` in build v2, thus ensuring that how we define a module is now only done one place in the code base. 

## Bump

- [ ] Patch
- [x] Skip


tests/test_unit/test_toolkit_package.py
doctrino, line 1: FYI. I let my friend Claude refactor this code such that the logic for finding imports could be reused between the three tests.

tests/test_unit/test_cli/test_command_sequences.py
```diff
@@ -40,13 +41,14 @@
 
 
 def find_all_modules() -> Iterator[Path]:
-    for module, _ in iterate_modules(BUILDABLE_PACKAGE):
-        if module.name == "references":  # this particular module should never be built or deployed
+    result, _ = ModuleParser.find_modules(BUILDABLE_PACKAGE)
+    for module_relative_path in result.keys():
+        if module_relative_path.name == "references":  # this particular module should never be built or deployed
```
ronpal, line 46: Is this the old built-in module? I think we safely can drop this code (and maybe the next elif too

ronpal (dismissed):

tests/test_unit/test_cli/test_command_sequences.py
doctrino: Actually, I am not sure what this is. It is partly the old module, but it committed to the repo in `tests/data/buildable_pacages`. I think it is worth keeping as it helped find issues with the refactoring I just did.

ronpal (approved):

PR #3229, opened 2026-09-09
[CDF-28572, CDF-28573] 🚄  Speed up tests

doctrino:
# Description

Speeding up tests round 1. 

- Unit test were runing around 5-6 minutes. Now 2.5 minutes. 
- Coverage tests were running at 7-8 minutes. Now 4.5 minutes. 

In total, shaving off 3 - 3.5 minutes on current runtimes, almost 50% improvement.

Will do a second round on it.

FYI: Locally I am able to run all unit test in 45 sec.

## Bump

- [ ] Patch
- [x] Skip

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -45,7 +45,7 @@ jobs:
       - name: Run pytest
         env:
           IS_GITHUB_ACTIONS: "true"
-        run: uv run pytest tests/test_unit -n8
+        run: uv run pytest tests/test_unit -n auto --durations=20
 
   test:
     name: Run tests
@@ -76,7 +76,7 @@ jobs:
       - name: Run pytest
         env:
           IS_GITHUB_ACTIONS: "true"
-        run: uv run pytest tests/test_unit -n8
+        run: uv run pytest tests/test_unit -n auto --durations=20
   coverage:
     name: Create coverage report
     runs-on: ubuntu-latest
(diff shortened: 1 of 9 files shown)


.github/workflows/build.yml
```diff
@@ -4,7 +4,7 @@ on:
   pull_request:
 
 env:
-  PYTHON_VERSION: '3.10'
+  PYTHON_VERSION: '3.14'
```
doctrino, line 7: We should test on 3.14 and running on the latest python is faster :)

.github/workflows/build.yml
```diff
@@ -45,9 +45,10 @@ jobs:
       - name: Run pytest
         env:
           IS_GITHUB_ACTIONS: "true"
-        run: uv run pytest tests/test_unit -n8
+        run: uv run pytest tests/test_unit -n auto --durations=20
```
doctrino, line 48: Should not for 8 processes when only 4 is available  - that is slower.

tests/test_unit/test_cdf_tk/test_commands/test_run.py
```diff
@@ -54,6 +55,12 @@ def function_build_folder() -> BuildLineage:
     )
 
 
+@pytest.fixture
+def mock_function_venv() -> Iterator[MagicMock]:
```
doctrino, line 59: Mocking the virtual env and running the proper test in smoke.

tests_smoke/test_commands/test_run.py
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+
+def mask_secrets(secrets: dict) -> dict:
+    return {k: "***" for k in secrets}
+
+"""
+    REQUIREMENT_TXT = "cognite-sdk>=7.85.0"
+
+    def test_run_local_function(self, toolkit_client: ToolkitClient, tmp_path: Path) -> None:
```
doctrino, line 29: Moved to smoke test as this is a slow test.

tests/test_unit/conftest.py
```diff
@@ -66,6 +68,14 @@ def create_session(*args: Any, **kwargs: Any) -> CreatedSession:
         yield approval_client
 
 
+@pytest.fixture(scope="session")
+def toolkit_client_cheap() -> ToolkitClient:
```
doctrino, line 72: Creating an approval client is expensive, this fixture is much much cheaper and saves us a lot when we do not need the approval client.

ronpal (approved):

PR #3233, opened 2026-09-10
[CDF-28585] 🌌 API list instances

doctrino:
# Description

image

## Bump

- [x] Patch
- [ ] Skip

### Added

- [alpha] New command `cdf api instances list` that enables listing out selected instances.

### DIFF
--- a/cognite_toolkit/_cdf_tk/apps/_api_app.py
+++ b/cognite_toolkit/_cdf_tk/apps/_api_app.py
@@ -1,3 +1,4 @@
+from enum import Enum
 from typing import Any
 
 import typer
@@ -8,6 +9,11 @@ from cognite_toolkit._cdf_tk.utils.auth import EnvironmentVariables
 from ._helpers import print_help_if_no_subcommand
 
 DEFAULT_LIST_LIMIT = 25
+
+
+class InstanceTypeEnum(str, Enum):
+    node = "node"
+    edge = "edge"
 
 
 class APIApp(typer.Typer):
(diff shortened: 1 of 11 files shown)


cognite_toolkit/_cdf_tk/apps/_api_app.py
```diff
@@ -0,0 +1,73 @@
...
+
+    @staticmethod
+    def main(ctx: typer.Context) -> None:
+        """Commands to work with data modeling instances"""
+        print_help_if_no_subcommand(ctx)
+
+    @staticmethod
+    def list(
+        view: str = typer.Option(
+            "cdf_cdm:CogniteDescribable/v1",
```
doctrino, line 44: I chose to include a default here as that is required for the search endpoint. My thinking is that calling `cdf api instances list` should just work. Two other options 1. require it, 2. Interactive prompting. However, I think a philosophy of this set of endpoints is that they should always just end in one request. They need to be simple, thus interactive is out. And require it make it harder to use, having the most basic view as a default makes it clearer how to use.

cognite_toolkit/_cdf_tk/feature_flags.py
```diff
@@ -98,6 +98,10 @@ class Flags(Enum):
         visible=True,
         description="Enables location-split InField data migration in order to untangle shared legacy instance spaces",
     )
+    API = FlagMetadata(
```
ronpal, line 101: Could we use the V09 flag instead?

cognite_toolkit/_cdf_tk/apps/_api_app.py
ronpal: I couldn't agree more, the power is in the simplicity. On that note, could we take it further and support 

`cdf api instances list -l 1001` 

It clearly has consequences, but...

cognite_toolkit/_cdf_tk/feature_flags.py
doctrino: Hmm, that would means we intend to release this very soon. Are we sure about that? I am not even sure we should do this.

cognite_toolkit/_cdf_tk/apps/_api_app.py
doctrino: We can, but then we have to use the slower query endpoint. my pushback would be that we wait until someone requests it - as it enables the user/agent to shot themselves in the foot by requesting 1 000 000 instances and thus freeze the terminal for minutes. Which leads us to do some progress bar style stuff, and thus increasing the complexity and violating the philosophy of a simple implementation for the API commands.

cognite_toolkit/_cdf_tk/feature_flags.py
ronpal: Fair point, stick with "api"

ronpal (dismissed):

ronpal (approved):

PR #3235, opened 2026-09-10
Warn earlier and more precisely in cdf build about unsupported DMS operations

Magssch:
# Description

CDF has a number of no-ops that, to a user, would _seem_ to work due to the inherent nature of the YAML format. Some of these operations include removing properties from containers and views, and removing views from data models without bumping the version. Additionally, a number of operations will be outright rejected by DMS, which we previously have shown no warnings for, e.g. changing the type of a property, changing a view's implements without a version bump, or changing a data model's view version without a version bump. Previously, `cdf deploy` would report on no-ops for containers, though only *after* attempting the deploy, when the message is easy to miss. This PR moves this drift detection earlier, into `cdf build`'s `DependencyRuleSet`, so users see a `ConsistencyError` before anything is deployed, and removes the now-redundant post-deploy warning in `ContainerCRUD`. The detection is also expanded to cover a number of invalid operations for views and data models, in addition to containers. Both operations that lead to errors and those that don't (but are silently ignored) are for now grouped with the same urgency (ConsistencyError).

The checks are based on the following reference material:
https://docs.cognite.com/cdf/dm/dm_concepts/dm_containers_views_datamodels#impact-of-changes-to-views-and-data-models

## Bump

- [x] Patch
- [ ] Skip

### Added

- `cdf build` now reports a clear error when local changes to containers, views, or data models would result in a no-op or error due to unsupported changes in CDF. Examples include removing properties from a deployed container, changing a container property's type or nullability, or changing a view's properties/implements or a data model's view versions without bumping the version.

### Removed

- Removed the post-deploy warning that was previously printed for container no-ops; an equivalent (and earlier) warning is now printed as part of `cdf build`.


cognite_toolkit/_cdf_tk/rules/_dependencies.py
```diff
@@ -72,3 +91,202 @@ def _create_reference_string(self, expected_resources: list[BuiltResource]) -> s
...
+                code=self.INVALID_OPERATION_CODE,
+                message=(
+                    f"Local config for container {container_id} has drifted from the state of the deployed container in CDF. Containers only "
+                    f"support a limited set of changes once deployed, so CDF will not reflect all the changes currently made to the local YAML config."
+                ),
+                fix=(
+                    f"Use 'cdf modules pull' to sync your local container config with the deployed version. See {URL.container_changes_docs}."
+                ),
+                source_file=source_file,
+            )
```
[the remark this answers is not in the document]
Magssch: Good point, made the detection quite a bit more granular now

cognite_toolkit/_cdf_tk/resource_ios/_resource_ios/datamodel.py
```diff
@@ -464,63 +460,13 @@ def update(self, items: Sequence[ContainerRequest]) -> list[ContainerResponse]:
...
                 raise ToolkitAPIError(
                     f"The container {item_id} was not updated due to an unknown error.",
                     code=500,
                 )
-            cdf_dict = self.dump_resource(updated_by_id[item_id], local_dict)
-            if cdf_dict != local_dict:
-                self._print_container_diff_warning(item_id, local_dict, cdf_dict)
         return updated
 
-    def _print_container_diff_warning(
```
doctrino, line 478: As discussed, lest keep this for now.

cognite_toolkit/_cdf_tk/rules/_dependencies.py
```diff
@@ -72,3 +91,310 @@ def _create_reference_string(self, expected_resources: list[BuiltResource]) -> s
...
         )
+
+    def _validate_data_modeling_changes(self) -> Iterable[ConsistencyError | InternalValidatorException]:
+        """Reports local container, view and data model changes that CDF will silently drop on deploy.
+
+        Containers cannot have properties removed, and views and data models are immutable per version.
+        Today this is only discovered during (or after) ``cdf deploy``. This surfaces the same findings
+        during ``cdf build``.
+        """
+        assert self.client is not None
```
doctrino: Typically we do not use assert in production. I guess this is because of MyPy, suggest you just pass the client in the `_validate_data_modeling_changes` as you have already checked it when you call this.

```suggestion
    def _validate_data_modeling_changes(self, client: ToolkitClient) -> Iterable[ConsistencyError | InternalValidatorException]:
        """Reports local container, view and data model changes that CDF will silently drop on deploy.

        Containers cannot have properties removed, and views and data models are immutable per version.
        Today this is only discovered during (or after) ``cdf deploy``. This surfaces the same findings
        during ``cdf build``.
        """
```

cognite_toolkit/_cdf_tk/rules/_dependencies.py
```diff
@@ -72,3 +91,310 @@ def _create_reference_string(self, expected_resources: list[BuiltResource]) -> s
...
+                item_id = crud.get_id(cdf_item)
+                if item_id not in local_by_id:
+                    continue
+                resource, local_dict = local_by_id[item_id]
+                cdf_dict = crud.dump_resource(cdf_item, local_dict)
+                if cdf_dict == local_dict:
+                    continue
+                yield from self._as_data_modeling_insights(item_id, resource, local_dict, cdf_dict)
+
+    def _load_local_resources(self, crud: ResourceIO[Any, Any, Any]) -> LocalResources:
```
doctrino: This logic will be used by any rule that needs to load resources. Suggest making it a method on `BuiltModule`.

cognite_toolkit/_cdf_tk/rules/_dependencies.py
```diff
@@ -72,3 +91,310 @@ def _create_reference_string(self, expected_resources: list[BuiltResource]) -> s
...
+                    resources_by_build_path.setdefault(resource.build_path, []).append(resource)
+
+        local_by_id: LocalResources = {}
+        for build_path, resources in resources_by_build_path.items():
+            resource_by_id = {resource.identifier: resource for resource in resources}
+            for raw in crud.load_resource_file(build_path):
+                request = crud.load_resource(raw)
+                item_id = crud.get_id(request)
+                if item_id in resource_by_id:
+                    local_by_id[item_id] = (resource_by_id[item_id], request.dump())
```
doctrino: I think this makes the code more complex and hard to follow. I know you need to dump the request to do a simple comparison later, but I suggest postponing that, by passing around dicts here, you end up having to do a lot of unsafe dict lookups later when comparing the containres, views and data models. This results in less readability. 
```suggestion
                    local_by_id[item_id] = (resource_by_id[item_id], request)
```

cognite_toolkit/_cdf_tk/rules/_dependencies.py
```diff
@@ -72,3 +91,310 @@ def _create_reference_string(self, expected_resources: list[BuiltResource]) -> s
...
+    def _container_insights(
+        self,
+        container_id: ContainerId,
+        resource: BuiltResource,
+        local_dict: dict[str, Any],
+        cdf_dict: dict[str, Any],
+    ) -> Iterable[ConsistencyError]:
+        source_file = format_insight_source_file(resource.source_path)
+        local_properties = local_dict.get("properties") or {}
+        cdf_properties = cdf_dict.get("properties") or {}
```
doctrino: These are examples of consequences of the pre-mature dump above. You know that local and cdf are both containers, so I don't think having an unstructured dict passed around is necessary.

cognite_toolkit/_cdf_tk/rules/_dependencies.py
```diff
@@ -72,3 +91,310 @@ def _create_reference_string(self, expected_resources: list[BuiltResource]) -> s
...
+
+        changed = sorted(
+            name
+            for name in set(cdf_properties) & set(local_properties)
+            if self._is_disallowed_container_property_change(local_properties[name], cdf_properties[name])
+        )
+
+        if changed:
+            removed = sorted(set(cdf_properties) - set(local_properties))
```
doctrino, line 239: ~If a containers has removed properties, but not changed anything this won't be reached and the user will not be warnend.~

I see you do it below, it is a bit hard to follow though.

cognite_toolkit/_cdf_tk/rules/_dependencies.py
```diff
@@ -72,3 +91,310 @@ def _create_reference_string(self, expected_resources: list[BuiltResource]) -> s
...
+                ),
+                source_file=source_file,
+            )
+
+        for field_name in ("properties", "constraints", "indexes"):
+            if field_name == "properties" and changed:
+                continue  # Already reported above as a "changed" insight.
+            missing = sorted(set(cdf_dict.get(field_name) or {}) - set(local_dict.get(field_name) or {}))
+            if missing:
+                yield self._missing_field_insight(container_id, source_file, field_name, missing)
```
doctrino: I though you could remove constraints and indexes. It is just that we cannot do it in Toolkit. Then, I rather add support for that than a warning indicating that is not possible.

cognite_toolkit/_cdf_tk/rules/_dependencies.py
```diff
@@ -72,3 +91,310 @@ def _create_reference_string(self, expected_resources: list[BuiltResource]) -> s
...
+                    f"Local config for view {view_id} has some properties {affected} that have been modified in a way CDF "
+                    f"does not support without a version bump. Deploying the current local YAML config will not apply "
+                    f"these changes to the view in CDF unless you update the view version."
+                ),
+                fix=(
+                    f"Update the view version to apply the change, or use 'cdf modules pull' to sync your local view config with the deployed version. See {URL.dm_changes_docs}."
+                ),
+                source_file=source_file,
+            )
+        elif removed:
```
doctrino, line 318: This is simpler to read, suggest the same for containers.

cognite_toolkit/_cdf_tk/rules/_dependencies.py
```diff
@@ -72,3 +91,310 @@ def _create_reference_string(self, expected_resources: list[BuiltResource]) -> s
...
+                    f"Local config for view {view_id} is missing properties "
+                    f"{humanize_collection([f'{name!r}' for name in removed])} that have previously been deployed to CDF for the "
+                    f"same view version. Deploying the current local YAML config will not remove them from the view in CDF unless you update the view version."
+                ),
+                fix=(
+                    f"Update the view version to apply the change, or use 'cdf modules pull' to sync your local view config with the deployed version. See {URL.dm_changes_docs}."
+                ),
+                source_file=source_file,
+            )
+        if local_dict.get("implements") != cdf_dict.get("implements"):
```
doctrino: Shouldn't this be a set comparison. Or does the order matter?

doctrino (commented):
See comments, I think there are both some logic/correctness issues as well as some potential for improved readability/reusability.

cognite_toolkit/_cdf_tk/rules/_dependencies.py
Magssch: From what I've understood, changing the ordering of implements views is a breaking change

cognite_toolkit/_cdf_tk/rules/_dependencies.py
Magssch: Removing indexes and constraints altogether from the container messages and reworking the code to be similarly structured, we should support updates/removal of those later down the line anyways so there isn't really a need to report on them as long as we keep the post-deploy message as well

cognite_toolkit/_cdf_tk/rules/_dependencies.py
Magssch: Yes, we need to add support for it, so since we now keep the post-deploy message, that one already explains that it's just toolkit that doesn't support it yet. Removing the reporting on indexes/constraints from the cdf build step so we don't need to remove it later when we implement support for updating indexes/constraints

cognite_toolkit/_cdf_tk/rules/_dependencies.py
Magssch: Should be a bit easier to follow now, the idea is to be able to print the message "properties has changed" as a catch-all when a user has both changed and deleted properties, while if they have _only_ deleted properties, we show "properties has been deleted" instread

cognite_toolkit/_cdf_tk/rules/_dependencies.py
Magssch: Refactored to use Toolkit response/request objects to pass around instead

cognite_toolkit/_cdf_tk/rules/_dependencies.py
Magssch: Refactored to use Toolkit response/request objects to pass around instead

doctrino (approved):

PR #3242, opened 2026-09-14
[CDF-28599] Fix node deploys failing on container-constrained direct relations

Magssch:
# Description

`NodeCRUD.create` sent nodes to the instances API in filesystem order, but DMS requires the target of a container-constrained direct relation to already exist or be included in the same request. Nodes are now split into dependency-respecting (topologically sorted) batches, mirroring the Tarjan-SCC batching `ContainerCRUD`/`ViewIO` already use, so a from-scratch deploy of interdependent nodes (including self-referential relations, e.g. parent/child within the same container) no longer fails.

## Bump

- [x] Patch
- [ ] Skip

### Fixed

- Nodes with direct relations to container-constrained properties can now be deployed successfully when using `cdf deploy,` and should no longer fail spuriously with a `Cannot auto-create a direct relation target in node` error.


cognite_toolkit/_cdf_tk/resource_ios/_resource_ios/datamodel.py
```diff
@@ -1253,7 +1263,105 @@ def dump_resource(self, resource: NodeResponse, local: dict[str, Any] | None = N
...
+        target container, either by existing in CDF already or by being included in the same request.
+        Computes the strongly connected components in topological order, then packs consecutive SCCs
+        into batches up to the instances API's limit, such that a node is always sent in the same or an
+        earlier batch than nodes referring to it through a constrained direct relation.
+        """
+        all_sources = {source.source for item in items for source in item.sources or []}
+        self._lookup_constrained_properties(all_sources)
+        if not any(self._constrained_properties_by_source.get(source) for source in all_sources):
+            # Without constrained properties, we don't need to compute dependencies.
+            return list(chunker(items, INSTANCE_UPSERT_ENDPOINT.item_limit))
```
doctrino: A bit overkill do chunk it here, the `.create` call already does that chunking. 

```suggestion
            return [[items]]
```

tests/test_unit/test_cdf_tk/test_cruds/test_node.py
```diff
@@ -0,0 +1,276 @@
...
+    ) -> None:
+        referrer_container_id = ContainerId(space="sp", external_id="Referrer")
+        target_container_id = ContainerId(space="sp", external_id="Target")
+        referrer = _node("sp", "referrer_node", referrer_container_id, {"ref": ref_value})
+        targets = [_node("sp", target_id, target_container_id, {}) for target_id in target_ids]
+
+        with monkeypatch_toolkit_client() as client:
+            loader = NodeCRUD(client, Path("build_dir"), None)
+            loader._constrained_properties_by_source = {referrer_container_id: {"ref"}, target_container_id: set()}
+            batches = loader._compute_deploy_batches([referrer, *targets])
```
doctrino: Learning from speeding up tests, the `monkeypatch_toolkit_client` is very expensive and using it in tests that do not need it causes minutes of test run time when it is done across 100s of tests. 

You are not using the client at all, you are testing the `_compute_deploy_batches` method. Thus, use the new `toolkit_client_cheap` fixture.  
```suggestion
    def test_container_constrained_direct_relation_orders_targets_before_referrer(
        self, ref_value: Any, target_ids: list[str], toolkit_client_cheap: ToolkitClient
    ) -> None:
        referrer_container_id = ContainerId(space="sp", external_id="Referrer")
        target_container_id = ContainerId(space="sp", external_id="Target")
        referrer = _node("sp", "referrer_node", referrer_container_id, {"ref": ref_value})
        targets = [_node("sp", target_id, target_container_id, {}) for target_id in target_ids]

        loader = NodeCRUD(toolkit_client_cheap, Path("build_dir"), None)
        loader._constrained_properties_by_source = {referrer_container_id: {"ref"}, target_container_id: set()}
        batches = loader._compute_deploy_batches([referrer, *targets])

```

doctrino: Note the same holds for all tests below.

tests/test_unit/test_cdf_tk/test_cruds/test_node.py
```diff
@@ -0,0 +1,276 @@
...
+
+        flat_ids = [node.external_id for batch in batches for node in batch]
+        for target_id in target_ids:
+            assert flat_ids.index(target_id) < flat_ids.index("referrer_node")
```
doctrino: FYI: This is O(n2). It does not matter since you have only a one element list.

In addition, I prefer avoiding loops in tests, especially in asserts. 

Wouldn't this be doing the same?
```suggestion
        assert flat_ids[-1] == "referrer_node", "Referrer must be sent after all its targets"
```

doctrino (commented):
Needs to remove all usage of `monkeypatch_toolkit_client` and replace it with fixture `toolkit_client_cheap` as currently these tests are slow.

tests/test_unit/test_cdf_tk/test_cruds/test_node.py
doctrino: Just to demo the difference. 

## Tests with monkeypatch_toolkit_client
image

## Test with toolkit_client_cheap (and only mocking what you need)
image

cognite_toolkit/_cdf_tk/resource_ios/_resource_ios/datamodel.py
Magssch: Good point, no need to manually chunk when we have no constraints

doctrino (approved):

PR #3260, opened 2026-09-16
[CDF-28172] Catch missing auth group space references at build time

ronpal:
### Problem

When a `Group.yaml` contained an unrecognised capability name, `GroupYAML.model_validate()` threw a `ValidationError` and `validated` was `None`. Space IDs in otherwise-valid capabilities (e.g. `dataModelsAcl` + `spaceIdScope`) were never checked against CDF — the missing space only surfaced as a `400` at real deploy time.

### Solution

Added `UnknownCapability(Capability)` in `capabilities.py`, wrapping unrecognised names in it instead of raising. The scope is parsed eagerly with the known `Scope` types so scope-based dependencies (spaces, datasets, …) are still extracted normally. `IDScope` is skipped since the resource type cannot be inferred without the capability name.

A `syntax_warning(source_file)` hook on `ToolkitResource` (overridden in `GroupYAML`) feeds a `MODEL-SYNTAX-WARNING` into the existing build output for each unknown capability name.

## Bump

- [x] Patch
- [ ] Skip

### Fixed

- Missing space references in auth group `spaceIdScope` capabilities are now caught at build time even when the same group contains an unknown capability name


cognite_toolkit/_cdf_tk/resource_ios/_auth.py
```diff
@@ -192,6 +193,37 @@ def get_dependencies(cls, resource: GroupYAML) -> Iterable[tuple[type[ResourceIO
...
+
+        if not valid_capabilities:
+            return
+
+        # model_construct bypasses validators — safe here because we only need the capabilities list
+        partial_group = GroupYAML.model_construct(
+            name=raw.get("name", ""),
+            capabilities=valid_capabilities,
+        )
+        yield from cls.get_dependencies(partial_group)
```
doctrino: I am afraid this will introduce new bugs worse than the one it fixes. The `.get_dependencies` method expects a validated YAML and with `model_construct` you have no such guarantee. I think we need to solve it in the GroupYAML class by introducing an unknown capability there and then maybe a special rule to deal with it.

cognite_toolkit/_cdf_tk/yaml_classes/capabilities.py
```diff
@@ -171,7 +171,26 @@ def find_capability_cls(cls, data: Any, handler: ModelWrapValidatorHandler[Self]
...
+            content = next(iter(data.values()), {})
+            scope_raw = content.get("scope") if isinstance(content, dict) else None
+            parsed_scope: Scope
+            if scope_raw and isinstance(scope_raw, dict):
+                try:
+                    parsed_scope = Scope.model_validate(scope_raw)
+                except Exception:
+                    parsed_scope = AllScope()
+            else:
+                parsed_scope = AllScope()
```
doctrino: Ideally use an unknown scope as well.

cognite_toolkit/_cdf_tk/commands/build_v2/build_v2.py
```diff
@@ -947,6 +947,7 @@ def _read_resource_file(
             try:
                 toolkit_resource = crud_class.yaml_cls.model_validate(parsed_yaml, extra="forbid")
                 identifier = toolkit_resource.as_id()
+                syntax_warning = toolkit_resource.syntax_warning(resource_file)
```
doctrino: If you have multiple groups in a file, then you must update the below as well. 

  Problem now is that there is an assumption that you can only have one syntax warning per file, but now you can have multiple. So you will need to update `SuccessfulReadYAMLFile` to support multiple syntax warnings.

cognite_toolkit/_cdf_tk/yaml_classes/base.py
```diff
@@ -16,5 +20,14 @@ def as_id(self) -> Identifier:
         """Return an identifier for this resource."""
         raise NotImplementedError()
 
+    def syntax_warning(self, source_file: Path) -> "ModelSyntaxWarning | None":
```
doctrino: I think this needs to be a list of syntax warnings.

doctrino (commented):
Better solution, I think you still need deal with multiple groups in a file as well.

cognite_toolkit/_cdf_tk/yaml_classes/capabilities.py
```diff
@@ -30,9 +30,10 @@ def find_scope_cls(cls, data: Any, handler: ModelWrapValidatorHandler[Self]) ->
             return handler(data)
         name, content = next(iter(data.items()))
         if name not in _SCOPE_CLASS_BY_NAME:
-            raise ValueError(
-                f"invalid scope name '{name}'. Expected one of {humanize_collection(_SCOPE_CLASS_BY_NAME.keys(), bind_word='or')}"
-            )
+            scope = UnknownScope.model_construct()
+            scope._unknown_name = name
+            scope._raw_data = data
+            return cast(Self, scope)
```
ronpal, line 36: @doctrino 

This "softens" unknown scope to warning regardless of capability. Is that OK, or should we only allow that for UnknownCapability? In other words, should we continue to raise if the Capability is known but the Scope isn't recognized?

doctrino (approved):

PR #3273, opened 2026-09-22
[CDF-28558]💂 Workflow v2

doctrino:
# Description

image

## Bump

- [x] Patch
- [ ] Skip

### Changed

- [alpha] The `cdf dev run workflows` is now based on workflows in CDF and not local configurations. In addition, the terminal output has been improved when running with the `--wait` flag, it now accounts for retries of tasks.


ronpal:
Screenshot 2026-09-24 at 14 41 32
Bug or feature, I can't decide 😄 

If I only have one WF, it shouldn't and doesn't provide an interactive selector. But should we print the WFs name to remove uncertainty? We print it on the next line anyway though.

ronpal (approved):

